PR #404, opened 2015-01-14
"Fixes GetterSetterToPropertyAdvancedPass by visiting the bases and conve...

golddranks:
...rting their properties first. This is neccessary for overrided properties of derived classes to work."

Hi, thanks for the quick fix for my previous crash report. I've got one more.

In the GetterSetterToPropertyAdvancedPass, QtGui generation crashes, when trying to access the base property of overrided property of a derived type. This happens with QWindow.surfaceType, which overrides QSurface.surfaceType.

In AstVisitor, it always visits the base classes of visited class declarations, but the overrided VisitClassDecl of GetterSetterToPropertyAdvancedPass doesn't do that - that's why the properties of the base classes aren't set.

Strange that it has worked this far, though. I wonder what changed?

### DIFF
--- a/src/Generator/Passes/GetterSetterToPropertyAdvancedPass.cs
+++ b/src/Generator/Passes/GetterSetterToPropertyAdvancedPass.cs
@@ -278,6 +278,11 @@ namespace CppSharp.Passes
         {
             if (VisitDeclarationContext(@class))
             {
+                if (Options.VisitClassBases)
+                    foreach (var baseClass in @class.Bases)
+                        if (baseClass.IsClass)
+                            VisitClassDecl(baseClass.Class);
+
                 new PropertyGenerator(@class, Log).GenerateProperties();
             }
             return false;


ddobrev:
I changed that at https://github.com/mono/CppSharp/commit/ec0f6a3d63db330fab90967f7603a7d3750d5266 . There was a crash with QtCore, I believe. Could you please generate QtCore with your change and see what happens?

golddranks:
I see, thanks for information. I checked your fix... while it indeed prevents the crash in QtCore, it doesn't seem to fix the problem. After quick debugging (man, I hate these deep recursion walk debugs, but I'm starting to get the knack of it :P)

The actual problem before your fix seems to be:

Walk to QMetaObject, walk to its base, QObject, walk QObject's methods, walk one of the methods' return type, QThread, walk to one of QThread's method's return type, QAbstractEventDispatcher, walk to its base which is QObject, it deems QObject already visited, so start unwinding the recursion stack: return to QAbstractEventDispatcher, generate QAbstractEventDispatcher properties. One of the properties overrides QObjects property, but it isn't generated yet -> CRASH.

PR #406, opened 2015-01-30
Fixes for namespaces: exporting and importing names

golddranks:
I finally decided to do this. Implemented more testcases that test nested and overlapping namespaces, and an import & export mechanism for names. Now a library wrapped with C++# exports its symbols, and another library that is dependent of the first one, can import them allowing wrapper dependencies.


ddobrev:
I don't understand this pull. You continue not to use NativeLibrary.Dependencies but rather stuff dependencies in options (ExportNames, ImportNames)? This looks like testing rather than production code to me. Options have nothing to do with library dependencies. Please clarify what this pull is supposed to do.

golddranks:
Okay, let's talk about it more. I'm really tired now, so let's discuss it some other day, @tritao, could you  not merge it yet, please.

golddranks:
Okay, I'll explain now a bit about this. Note: I might have understood something wrong or made wrong assumptions, so correct me if you need :)

When I started doing this, I thought that maybe it could reference everything with fully qualified names, so if Derived class references Base class, rather than importing Base with "using NamespacesBase", reference it always by NamespacesBase.Base in code. This will prevent the problem with "using" directives: If two libraries define the same subnamespace, "using" can import both with the same name. (Of course it could import it using "as" and generate another name, but then we would have to take this into account when generating code also, so it wouldn't make things a bit easier?)

But: when generating the code, it has to know which symbols come from which library, to be able to decorate the names. Clang doesn't know this, since the decision about which constitutes a single library is done by C++#. I simply solved this making it write a dictionary of DeclInfo structs, that have the info: original name of an exported symbol and the name decorated with the output namespace name. This is why I added the Export and Import dictionaries - they can be passed from one library building run to the next - to export the dependent names.

If there's an obviously better way to do this, please tell me, it's not a big patch and quite easy to change.

golddranks:
We discussed this further in Skype, and @ddobrev came up with a better design. I'll work on it.

tritao:
OK that's good to know, I'll wait for the new pull.

PR #407, opened 2015-02-04
Fix namespaces 2

golddranks:
All right, what do you think about this?


ddobrev:
Looks much better, thank you for this work. I am afraid, though, that I can only review it in detail. I did notice something that arouse my interest: why do you need to change tunit.Name in your pass? We do have Declaration.OriginalName in case of need but I'm just wondering.
@tritao your opinion on this pull would be most useful.

ddobrev:
@golddranks please also remove using System.Runtime.Serialization.Formatters.Binary; and other using directives you don't need any more.

ddobrev:
OK, I can see now why you rename the translation unit. There's something interesting in the idea, I also like that it allows to to remove code in GetNestedQualifiedName rather than add more. Joao, what do you think?
 
Now that you do this anyway, please use full names to all needed cases. Now you only use it for base types which would cause incompilable code when a dependent type is used as a parameter or a return type, etc. In short, now that you've gone for qualified names instead of usings, any usage of a dependent type must be fully qualified.

PR #408, opened 2015-02-06
Fix namespaces 3

golddranks:
Round three.

I addressed all the comments I got in the earlier PR. Still got nits to pick?


ddobrev:
I'm afraid so. :) You have missed "Please use full names to all needed cases. Now you only use it for base types which would cause incompilable code when a dependent type is used as a parameter or a return type, etc. In short, now that you've gone for qualified names instead of usings, any usage of a dependent type must be fully qualified.".

golddranks:
Oh, dang :D I guess that there is a fair amount of "all needed cases"? Do you think I'll be able to catch them if I read through the CSharpTextTemplate.cs and wrap all the usages of `something.Name` with calls to `QualifiedIdentifier()` like with the case of base classes?

ddobrev:
I think so but you need to skip types residing in the current name-space.

golddranks:
I extended the tests, and found out that the types already get printed right; this seems to be because it already uses the machinery to print typenames with namespaces. I also implemented a function called QualifiedIdentifierIfNeeded, that skips the qualifications when they are not needed (i.e. when already in the same namespace).

ddobrev:
I can see you also tested properties, very nice. :) No further bugging from me on this one. I guess unless @tritao has some remarks on the code itself, this is ready to merge.

golddranks:
Oh damn, just ran the Basic tests and I'm getting some errors, I'll check if they are regressions.

golddranks:
Okay, two little tweaks: 1) When it processed the same file two times (creating CLI and CSharp versions), it produced strange results because it relied on the fact that same file would be visited only once during the lifetime of the script. 2) When it tried to strip superfluous namespace decorations, it should've stop when there was only one name left, but it continued until there was nothing left. Fixed these, but I'm running every test to catch if there's more regressions. Will post the commit soon.

ddobrev:
Great, I'll take a look at the code when you've sent it, to better understand both your solution and the problem.

golddranks:
Oh, btw. the patch above fixes those. I run every test, no problems. (There seems to be a regression with QtCore, but that was introduced already before this pull request. It first appears when building with commit f2de736f81473049ee36ef48c7b6227c3a4a5159, "Fixed cyclic dependency when parsing class templates with friends." Maybe I should try and fix that next.)

tritao:
Sorry for the delay in processing this. Code looks real solid and well-tested. Thanks for your contribution.

PR #425, opened 2015-03-05
Determine bitness on osx

stsundermann:
This reads the bitness from mono which is in path.

lipo -info gives the following output on a 64 bit mono:
```
localhost:build Stephan$ lipo -info `which mono`
Non-fat file: /usr/local/bin/mono is architecture: x86_64
```

With the executable from the Mono MDK the following is returned:
```
localhost:build Stephan$ lipo -info /Library/Frameworks/Mono.framework/Mono
Non-fat file: /Library/Frameworks/Mono.framework/Mono is architecture: i386
```

This patch checks for x86_64 in the output if found a 64 bit build is configured, else 32 bit.


tritao:
Thanks

PR #431, opened 2015-03-19
added automatic downloading of depends

grbd:
Hi,

I've added a small change to the build system to allow for automatic downloading of depends
I'm not sure if this will be accepted since it uses Nuget.exe (requires mono)
to handle the download of the prebuilt deps

    DownloadDepends.bat (windows)
    ./premake5-linux --file=premake4.lua depend_nuget (linux)
    ./premake5-linux --file=premake4.lua depend_gitclone (linux)

Note to get Nuget to work under linux you may need to run this before hand to load in the certificates for mono

    mozroots --import --sync

to add new deps, it's just a case of editing packages.config
the versions of clang / llvm it downloads via git is specified in DownloadDeps.lua
I've tried this on windows and kubuntu 14.10 (getting linux to build is another story)
not sure about Mac OsX

I'm thinking of using premake in my own projects since it looks a lot simpler to setup than cmake
but there are a couple of things I've spotted
these may have been fixed in the latest premake from the website, I haven't tried that one yet

1. os.execute under linux seems to have problems with http links
http:// becomes http:/ for some reason, perhaps this is a lua bug?
doesn't seem to affect windows

2. forward decleration of functions didn't seem to work when I tried it for some reason


tritao:
Thanks for working on this, there are multiple aspects to this pull I'd like to comment.

About LLVM and Clang, I like your approach but not everyone has Git on their path (on Windows), maybe the best bet would be to add them as Git submodules instead. What do you think?

About NuGet, honestly I'd rather not mess with it since it adds more complexity... but I can also see the advantages of moving to it for dependency management so I'm open to adopting your suggestions.

I've never experienced those kinds of problems on Premake but I've also not done much OS level calls before. You might be interested in this pull I've worked on a few months (but haven't had time to finish / get merged since): https://bitbucket.org/premake/premake-dev/pull-request/122/initial-http-support

That would allow direct HTTP downloads without calls to external processes.

And what do you mean by forward declaration of functions? That should not be needed with Lua.

grbd:
Hi,

For llvm / clang I've just had a different idea that might be better
With github we can do a direct download based on the commit id, with that patch you mentioned it might just be a case of a simple http download via

* https://github.com/llvm-mirror/llvm/archive/0e8abfa6ed986c892ec723236e32e78fd9c47b88.zip
* https://github.com/llvm-mirror/clang/archive/3457cd5516ac741fa106623d9578f5ac88593f4d.zip

I should be able to code that into lua simply enough. This has the added benifit of not having to download the entire git tree (unless you want to work on clang / llvm from a git point of view).
I'll look into adding this in as a default option, although I can always leave the git download there for anyone that has git in they're path / wants to work on clang / llvm

I'm afraid I don't have much experience with git submodules, most of what I've read about them is pretty bad, and suggests subtree's instead. The downside with subtree's though is that your storing a copy of the code in a seperate branch


For NuGet I could replace it with http links, it just seemed a lot more elegant to do it that way
It's basically a standalone .Net executable that's runnable via mono or under windows so in theory it should be cross platform. I figured since CppSharp needs mono anyway, any depends for it would already be provided. I'm in the process of installing gentoo onto a Rpi2, so I'm going to give it a go with that (my ultimate aim is to have cross platform Qt .Net apps that will run on windows or the rpi2). Just to see if it works

Just to mention, having the version number at the end of the directory name for the depend is optional btw. Also it's possible just to specify a specific version or the latest version of a depend

One idea I did have was to relocate some of the build files that are not called directly (like Helpers.lua)
into a seperate sub directory like build/bin or build/scripts for less clutter, and just keep those scripts that are called by the user directly under build, would this be a good idea? I didn't want to change too much without permission.
Forward declerations aren't needed really, I just noticed that normal lua supports them and kind of assumed it was supported in premake but it's no big deal.

ddobrev:
About your suggestion concerning LLVM/Clang: Git submodules point to a specific commit by definition. So you can see that there's nothing we would gain this way except more script to maintain.

tritao:
Thanks a lot for this idea of using Premake for dependencies. We have now switched to a variant I developed using your pull request as a base. It turned out to be a lot more complicated due to having to handle some details like CI infrastructure.

PR #440, opened 2015-04-13
Generate delegates for non-typedef'd function pointers to prevent runtime exceptions during marshalling

spencercw:
Function pointers are converted to delegates which are marshalled using GetDelegateForFunctionPointer and GetFunctionPointerForDelegate. If the function pointers are typedef'd this works fine because a delegate is generated for each typedef. If it's just a plain function pointer then Action<> or Func<> is used, which fails because the marshalling functions can't deal with generic types. I've added a new pass which generates delegates for these function pointers to work around this issue.

I'm obviously not particularly familiar with the code base, so any thoughts on how this might be done better are welcome.

### DIFF
--- a/build/Tests.lua
+++ b/build/Tests.lua
@@ -12,11 +12,11 @@ function SetupExampleProject()
   SetupParser()
 end
 
-function SetupTestProject(name, file, lib)
+function SetupTestProject(name, extraFiles)
   SetupTestGeneratorProject(name)
   SetupTestNativeProject(name)  
-  SetupTestProjectsCSharp(name)
-  SetupTestProjectsCLI(name)
+  SetupTestProjectsCSharp(name, nil, extraFiles)
+  SetupTestProjectsCLI(name, extraFiles)
 end
 
 function SetupTestCSharp(name)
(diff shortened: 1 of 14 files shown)


tritao:
Looks good, I'm only wondering if there might be any conflict problems when adding similar delegates in the same namespace by two different translation units.

spencercw:
Ah yeah, that doesn't work. I had assumed the translation units could just be treated separately. One option would be to generate a class with the name of the translation unit inside the namespace for any anonymous delegates that don't belong to a class. That would fail if two translation units happened to have the same name though (if that's even supported). Got any better ideas?

tritao:
We can try to lookup for a compatible synthetized delegate inside the namespace and if exists reference it directly instead of creating a new one.

spencercw:
Ok I've changed the pass to run over the entire library instead of doing each translation unit separately and matching declaration contexts by qualified name.

tritao:
Nice, looks good. Thanks for the thorough tests and making sure it works on both backends!

tritao:
By the way, it might be worth noting that the marshaling failure is only a problem on .NET 4 (according to @ddobrev, I didn't personally test it). Once that's obsolete we can remove the pass and use `Action/Func` again.

ddobrev:
I put both my heads on the line that it fails on .NET 4 and older. Marshalling generic delegates was introduced in Microsoft .NET 4.5 - that is, Mono had it years before. We've considered switching to generic delegates with XP users switching to Mono for Windows but that's of a low priority at present.

spencercw:
Thanks. Interesting that it works on .NET 4.5, I didn't know that.

Issue #445, opened 2015-05-06
Function signstures include non-const appendages after the closing parenthesis


ddobrev:
Add these to AST.h:

void func() MY_MACRO;
void func() const MY_MACRO;

The signature of the function should only include the const (if any) but it also includes any other symbols.

PR #446, opened 2015-05-08
Fixed the signatures of const functions

ddobrev:


ddobrev:
Fixes https://github.com/mono/CppSharp/issues/445 .

PR #450, opened 2015-05-25
Fixed code generation when using arrays in Value Type types...

genuinelucifer:
Just added array number to end of '__array' identifier for CLI generator.
And added fixed scoping for C# generator in getters and setters since the arrays are fixed size buffers.

### DIFF
--- a/src/Generator/Generators/CLI/CLIMarshal.cs
+++ b/src/Generator/Generators/CLI/CLIMarshal.cs
@@ -47,8 +47,7 @@ namespace CppSharp.Generators.CLI
             {
                 case ArrayType.ArraySize.Constant:
                     var supportBefore = Context.SupportBefore;
-                    string value = Generator.GeneratedIdentifier("array") 
-                                            + _arrayNum;
+                    string value = Generator.GeneratedIdentifier("array") + _arrayNum;
                     ++_arrayNum;
                     supportBefore.WriteLine("cli::array<{0}>^ {1} = nullptr;", array.Type, value, array.Size);
                     supportBefore.WriteLine("if ({0} != 0)", Context.ReturnVarName);
--- a/src/Generator/Generators/CLI/CLIMarshal.cs
+++ b/src/Generator/Generators/CLI/CLIMarshal.cs
@@ -40,15 +40,15 @@ namespace CppSharp.Generators.CLI
             return decl.Visit(this);
         }
 
-        private static int _arrayNum = 0;
+        private static int arrayNum = 0;
         public override bool VisitArrayType(ArrayType array, TypeQualifiers quals)
         {
             switch (array.SizeType)
             {
                 case ArrayType.ArraySize.Constant:
                     var supportBefore = Context.SupportBefore;
-                    string value = Generator.GeneratedIdentifier("array") + _arrayNum;
-                    ++_arrayNum;
+                    string value = Generator.GeneratedIdentifier("array") + arrayNum;
+                    ++arrayNum;
                     supportBefore.WriteLine("cli::array<{0}>^ {1} = nullptr;", array.Type, value, array.Size);
                     supportBefore.WriteLine("if ({0} != 0)", Context.ReturnVarName);
                     supportBefore.WriteStartBraceIndent();
--- a/src/Generator/Generators/CLI/CLIMarshal.cs
+++ b/src/Generator/Generators/CLI/CLIMarshal.cs
@@ -40,13 +40,16 @@ namespace CppSharp.Generators.CLI
             return decl.Visit(this);
         }
 
+        private static int _arrayNum = 0;
         public override bool VisitArrayType(ArrayType array, TypeQualifiers quals)
         {
             switch (array.SizeType)
             {
                 case ArrayType.ArraySize.Constant:
                     var supportBefore = Context.SupportBefore;
-                    string value = Generator.GeneratedIdentifier("array");
+                    string value = Generator.GeneratedIdentifier("array") 
+                                            + _arrayNum;
+                    ++_arrayNum;
                     supportBefore.WriteLine("cli::array<{0}>^ {1} = nullptr;", array.Type, value, array.Size);
                     supportBefore.WriteLine("if ({0} != 0)", Context.ReturnVarName);
                     supportBefore.WriteStartBraceIndent();
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -856,12 +856,29 @@ namespace CppSharp.Generators.CSharp
                 WriteStartBraceIndent();
 
                 var marshal = new CSharpMarshalManagedToNativePrinter(ctx);
-                ctx.ReturnVarName = string.Format("{0}{1}{2}",
+
+                var arrayType = field.Type as ArrayType;
+                
+                if(arrayType!=null && @class.IsValueType==true)
+                {
+                    CSharpTypePrinter _ctp = new CSharpTypePrinter(ctx.Driver);
+                    string type = arrayType.CSharpType(_ctp).Type;
+                    WriteLine(string.Format("fixed ({0} __arrPtr = {1}.{2})",
+                        type.Replace("[]","*"), Helpers.InstanceField,
+                        Helpers.SafeIdentifier(field.OriginalName)));
+                    WriteStartBraceIndent();
+                    ctx.ReturnVarName = string.Format("__arrPtr");
+                }
+                else
+                {
+                    ctx.ReturnVarName = string.Format("{0}{1}{2}",
                     @class.IsValueType
                         ? Helpers.InstanceField
                         : string.Format("((Internal*) {0})", Helpers.InstanceIdentifier),
                     @class.IsValueType ? "." : "->",
                     Helpers.SafeIdentifier(field.OriginalName));
+                }
+                
                 param.Visit(marshal);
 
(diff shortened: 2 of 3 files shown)


src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -856,12 +856,29 @@ private void GeneratePropertySetter<T>(QualifiedType returnType, T decl, Class @
                 WriteStartBraceIndent();
 
                 var marshal = new CSharpMarshalManagedToNativePrinter(ctx);
-                ctx.ReturnVarName = string.Format("{0}{1}{2}",
+
+                var arrayType = field.Type as ArrayType;
+                
+                if(arrayType!=null && @class.IsValueType==true)
```
ddobrev: Please put spaces around '!=' and do not use ==/!= for booleans: just use if (var)/if (!var).

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -872,6 +889,11 @@ private void GeneratePropertySetter<T>(QualifiedType returnType, T decl, Class @
                     WriteLine("{0} = {1};", ctx.ReturnVarName, marshal.Context.Return);                    
                 }
 
+                if (arrayType != null && @class.IsValueType == true)
```
ddobrev: Again: checking a Boolean with ==.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -962,6 +984,19 @@ private void GeneratePropertyGetter<T>(QualifiedType returnType, T decl, Class @
                     ReturnType = decl.QualifiedType
                 };
 
+                var arrayType = field.Type as ArrayType;
+
+                if (arrayType != null && @class.IsValueType == true)
```
ddobrev: Checking a Boolean with ==.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -962,6 +984,19 @@ private void GeneratePropertyGetter<T>(QualifiedType returnType, T decl, Class @
...
+                {
+                    CSharpTypePrinter _ctp = new CSharpTypePrinter(ctx.Driver);
```
ddobrev: We don't prefix locals with '_', that's only in the generated code so that we can avoid naming conflicts. Just 'ctp' is better.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -969,6 +1004,14 @@ private void GeneratePropertyGetter<T>(QualifiedType returnType, T decl, Class @
                     Write(marshal.Context.SupportBefore);
 
                 WriteLine("return {0};", marshal.Context.Return);
+
+
+                if (arrayType != null && @class.IsValueType == true)
```
ddobrev: Checking a Boolean with ==.

src/Generator/Generators/CLI/CLIMarshal.cs
```diff
@@ -40,13 +40,16 @@ public override bool VisitTagType(TagType tag, TypeQualifiers quals)
...
+        private static int _arrayNum = 0;
         public override bool VisitArrayType(ArrayType array, TypeQualifiers quals)
         {
             switch (array.SizeType)
             {
                 case ArrayType.ArraySize.Constant:
                     var supportBefore = Context.SupportBefore;
-                    string value = Generator.GeneratedIdentifier("array");
+                    string value = Generator.GeneratedIdentifier("array") 
+                                            + _arrayNum;
```
tritao: This should be on the previous line.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
tritao: Or better yet, `typePrinter`.

genuinelucifer:
Changed done...

genuinelucifer:
I had put that on next line because git diff could not show more than 80 chars on one line. 
Fixed it.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -962,6 +984,19 @@ private void GeneratePropertyGetter<T>(QualifiedType returnType, T decl, Class @
                     ReturnType = decl.QualifiedType
                 };
 
+                var arrayType = field.Type as ArrayType;
+
+                if (arrayType != null && @class.IsValueType)
+                {
+                    CSharpTypePrinter typePrinter = new CSharpTypePrinter(ctx.Driver);
+                    string type = arrayType.CSharpType(typePrinter).Type;
+                    WriteLine(string.Format("fixed ({0} __arrPtr = {1}.{2})",
```
ddobrev: Type printers can have their contexts changed. You should be able to give it a native context and have the type produced right away instead of replacing [] with *.

genuinelucifer: I didn't actually get it. Isn't ctx in the setter code a native Context?
because in setter too the typePrinter produces type with []

ddobrev: "native" means code interacting with C++, marshalling code. The context by default is managed, that is, regular C#.

genuinelucifer: Oh. At this point I have no idea about this. Will look into it.

genuinelucifer: I tries to push a Native context and then try. I also used VisitArrayType of typePrinter but nothing worked as required. Both the changes :
https://gist.github.com/genuinelucifer/079b00302845c486fb24
.
Both return 'fixed float' and required is 'float*'.... Now they don't even return []s....
Also, I tried CLI counterpart of typePrinter but it generates cli<> types which won't work...
.
If you could point me in the right direction, it would be quite helpful.

genuinelucifer: Also I saw that getters and setters are not generated for multidimentional arrays.
So, wouldn't it be good to simply use the substring of 'type' before last two characters as there will always be only one set of [].
This works perfectly fine too :
https://gist.github.com/genuinelucifer/45a29b36ec4f5254ea21

genuinelucifer:
@ddobrev Please have a look and help....
I tried 2 ways to use Native contexts both of which failed. Please tell me what is wrong with my approach. My tries -
https://gist.github.com/genuinelucifer/079b00302845c486fb24
.
Also, we are only marshalling all c++ arrays to 1-D array in C#. Moreover, getters and setters are genrated only for originally 1-D arrays. SO --
https://gist.github.com/genuinelucifer/45a29b36ec4f5254ea21
The above diff words perfectly fine and doesn't need to replace anything just remove the last two letters and use '*' sign.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -872,6 +889,11 @@ private void GeneratePropertySetter<T>(QualifiedType returnType, T decl, Class @
                     WriteLine("{0} = {1};", ctx.ReturnVarName, marshal.Context.Return);                    
                 }
 
+                if (arrayType != null && @class.IsValueType)
+                {
```
ddobrev: Please do not use braces with one-liners. I personally prefer then but @tritao doesn't so we'd better do without.

src/Generator/Generators/CLI/CLIMarshal.cs
```diff
@@ -40,13 +40,15 @@ public override bool VisitTagType(TagType tag, TypeQualifiers quals)
             return decl.Visit(this);
         }
 
+        private static int _arrayNum = 0;
```
ddobrev: Do not prefix fields, static or instance, with an underscore. We don't do that in this project.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -856,12 +856,29 @@ private void GeneratePropertySetter<T>(QualifiedType returnType, T decl, Class @
...
                 var marshal = new CSharpMarshalManagedToNativePrinter(ctx);
-                ctx.ReturnVarName = string.Format("{0}{1}{2}",
+
+                var arrayType = field.Type as ArrayType;
+                
+                if(arrayType != null && @class.IsValueType)
+                {
+                    CSharpTypePrinter typePrinter = new CSharpTypePrinter(ctx.Driver);
+                    string type = arrayType.CSharpType(typePrinter).Type;
+                    WriteLine(string.Format("fixed ({0} __arrPtr = {1}.{2})",
```
ddobrev: Please extract "__arrPtr" to a local and create it using Helpers.GeneratedIdentifier("arrPtr").

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
ddobrev: The native context was a guess of mine which turned out to be incorrect. I am sorry for wasting your time. Please note, however, that you need to avoid manipulating strings unless there's no other way, and that's why asked you to try the printer.

ddobrev: As usual, please combine your commits into one for the final version.

genuinelucifer:
Since this PR can't be merged due to merge conflicts (Github states that) so I am closing this and opening a new one with all the required changes...

PR #451, opened 2015-05-25
Fixed compilation issues when nesting more than one anonymous types.

genuinelucifer:
The [double underscore] identifier was generated for anonymous types and hence resulted in conflict when nesting more than one anonymous types.
I checked if an identifier had just '_' in it then i repeated the underscore 'numSafeIden' times which kept track of how many times such identifier has been generated.
.
Alternate solution (probably more elegant) can be if [double underscore] was to be suffixed with integer 'numSafeIden' instead of repeating the underscore so many times.

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -713,6 +713,11 @@ namespace CppSharp.Generators.CSharp
 
             var safeIdentifier = Helpers.SafeIdentifier(field.OriginalName);
 
+            if(safeIdentifier.All(c => c.Equals('_')))
+            {
+                safeIdentifier = Helpers.SafeIdentifier(field.Name);
+            }
+
             PushBlock(CSharpBlockKind.Field);
 
             WriteLine("[FieldOffset({0})]", field.OffsetInBytes);
--- a/src/Generator/Passes/CleanInvalidDeclNamesPass.cs
+++ b/src/Generator/Passes/CleanInvalidDeclNamesPass.cs
@@ -3,6 +3,7 @@ using System.Linq;
 using CppSharp.AST;
 using CppSharp.Generators.CLI;
 using CppSharp.Generators.CSharp;
+using System.Text;
 
 namespace CppSharp.Passes
 {
@@ -152,6 +153,21 @@ namespace CppSharp.Passes
         {
             item.Name = CheckName(item.Name);
             return base.VisitEnumItem(item);
+        }
+
+        public override bool VisitFieldDecl(Field field)
+        {
+            if (field.Class.Fields.Count(c => c.Name.Equals(field.Name)) > 1)
+            {
+                StringBuilder str = new StringBuilder();
+                str.Append(field.Name);
+                do
+                {
+                    str.Append('_');
+                } while (field.Class.Fields.Any(c => c.Name.Equals(str.ToString())));
+                field.Name = str.ToString();
+            }
+            return base.VisitFieldDecl(field);
         }
     }
 }
--- a/tests/Basic/Basic.h
+++ b/tests/Basic/Basic.h
@@ -647,6 +647,24 @@ public:
         {
         };
     };
+	struct
+	{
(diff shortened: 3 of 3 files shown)


src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -34,11 +34,25 @@ public static class Helpers
...
+            
+            id = new string(id.Select(c => char.IsLetterOrDigit(c) ? c : '_').ToArray());
+
+            if (id.Replace("_", "") == string.Empty)
+            {
+                id = string.Empty;
+                ++numSafeIden;
+                for (var i = 0; i < numSafeIden; ++i)
+                {
+                    id += '_';
```
ddobrev: Never concatenate ("+") strings in a loop. Creating strings is an expensive operation because strings are immutable which means that each concatenation creates a new string object. Use StringBuilder.Append instead.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -34,11 +34,25 @@ public static class Helpers
...
+        private static int numSafeIden = 0;
         public static string SafeIdentifier(string id)
         {
             if (id.All(char.IsLetterOrDigit))
                 return Keywords.Contains(id) ? "@" + id : id;
-            return new string(id.Select(c => char.IsLetterOrDigit(c) ? c : '_').ToArray());
+            
+            id = new string(id.Select(c => char.IsLetterOrDigit(c) ? c : '_').ToArray());
+
+            if (id.Replace("_", "") == string.Empty)
```
ddobrev: string.Replace creates a new string which is expensive. The fastest option here is:
if (id.All(c => c == '_'))

genuinelucifer:
I am new to C#. Thanks for the insights..
.
Improvement done.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -34,11 +34,27 @@ public static class Helpers
                 "void", "partial", "yield", "where"
             };
 
+        private static int numSafeIden = 0;
```
ddobrev: I paid a little more attention to this now. I think it may cause problems. It is static which means that it will keep increasing for the generation of the entire library, even if the anonymous types reside in different name-spaces. I think we'd better check if the parent name-space already has a type of the same name and then modify it.

genuinelucifer: Will do. :+1:

genuinelucifer: I guess now it is taken care of. Please review.

genuinelucifer:
Updated the code. Please review again. @tritao @ddobrev

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -711,6 +711,17 @@ private void GenerateClassInternalsField(Field field)
                 !(@class != null && @class.IsUnion)) || (@class != null && @class.TranslationUnit.IsSystemHeader))
                 return;
 
+            if (field.Class.Fields.Count(c => c.Name.Equals(field.Name)) > 1)
```
tritao: We should not modify `field.OriginalName`. So I'd advise to extract this piece of code to a static method and change the `Helpers.SafeIdentifier(field.OriginalName);` call to call the method instead.

genuinelucifer: We need to change atleast one member to keep track of the changes.
So if not OriginalName, should I change 'Name' field?

tritao:
This is looking good to me, but I think we should move this code to CleanInvalidDeclNamesPass, https://github.com/mono/CppSharp/blob/master/src/Generator/Passes/CleanInvalidDeclNamesPass.cs.

genuinelucifer:
Ohh. Yes that seems more logical.
Will do.

tritao:
Looks good.

PR #477, opened 2015-06-14
Fixed default args with template types; added an option for expressions to use in type maps

ddobrev:


tests/CSharpTemp/CSharpTemp.cs
```diff
@@ -41,6 +41,56 @@ public override void CSharpMarshalToManaged(MarshalContext ctx)
...
+    [TypeMap("QList")]
+    public class QList : TypeMap
+    {
+        public override bool IsIgnored
+        {
+            get
+            {
+                TemplateSpecializationType type = (TemplateSpecializationType) this.Type;
+                QualifiedType pointeeType = type.Arguments[0].Type;
+                TypeIgnoreChecker checker = new TypeIgnoreChecker(TypeMapDatabase);
```
tritao: We can use a var here.

tests/CSharpTemp/CSharpTemp.cs
```diff
@@ -41,6 +41,56 @@ public override void CSharpMarshalToManaged(MarshalContext ctx)
...
+                pointeeType.Visit(checker);
+                return checker.IsIgnored;
+            }
+        }
+
+        public override string CSharpSignature(CSharpTypePrinterContext ctx)
+        {
+            if (ctx.CSharpKind == CSharpTypePrinterContextKind.Native)
+            {
+                if (Type.IsAddress())
```
tritao: No need for the extra braces here, and add a newline after the if.

tests/CSharpTemp/CSharpTemp.cs
```diff
@@ -41,6 +41,56 @@ public override void CSharpMarshalToManaged(MarshalContext ctx)
...
+            }
+
+            TemplateSpecializationType templateSpecialization = (TemplateSpecializationType) ctx.Type.Desugar();
+            TemplateArgument templateArgument = templateSpecialization.Arguments[0];
+            if (templateArgument.Type.Type.IsPointerToPrimitiveType())
+            {
+                return string.Format("System.Collections.Generic.{0}<global::System.IntPtr>",
+                    ctx.CSharpKind == CSharpTypePrinterContextKind.DefaultExpression ? "List" : "IList");
+            }
+            return string.Format("System.Collections.Generic.{0}<{1}>",
```
tritao: Newline between the if block and the return.

PR #486, opened 2015-06-16
Fixed code generation when using arrays in Value Type types.

genuinelucifer:
Fixed all the changed in one commit and a mergeable PR...

### DIFF
--- a/src/Generator/Generators/CLI/CLIMarshal.cs
+++ b/src/Generator/Generators/CLI/CLIMarshal.cs
@@ -40,13 +40,15 @@ namespace CppSharp.Generators.CLI
             return decl.Visit(this);
         }
 
+        private static int arrayNum = 0;
         public override bool VisitArrayType(ArrayType array, TypeQualifiers quals)
         {
             switch (array.SizeType)
             {
                 case ArrayType.ArraySize.Constant:
                     var supportBefore = Context.SupportBefore;
-                    string value = Generator.GeneratedIdentifier("array");
+                    string value = Generator.GeneratedIdentifier("array") + arrayNum;
+                    ++arrayNum;
                     supportBefore.WriteLine("cli::array<{0}>^ {1} = nullptr;", array.Type, value, array.Size);
                     supportBefore.WriteLine("if ({0} != 0)", Context.ReturnVarName);
                     supportBefore.WriteStartBraceIndent();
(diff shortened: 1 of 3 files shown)


src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -856,12 +856,30 @@ private void GeneratePropertySetter<T>(QualifiedType returnType, T decl, Class @
...
                 var marshal = new CSharpMarshalManagedToNativePrinter(ctx);
-                ctx.ReturnVarName = string.Format("{0}{1}{2}",
+                
+                var arrayType = field.Type as ArrayType;
+
+                if (arrayType != null && @class.IsValueType)
+                {
+                    CSharpTypePrinter typePrinter = new CSharpTypePrinter(ctx.Driver);
+                    string type = arrayType.CSharpType(typePrinter).Type;
+                    string arrPtrIden = Helpers.SafeIdentifier("__arrPtr");
```
ddobrev: This works too but what I meant was that you could just do:

Helpers.SafeIdentifier("arrPtr")

The point is that SafeIdentifier prepends "__" for you, that's the point. No problem if you leave it this way, I just wanted you to know what the convenience of SafeIdentifier.

src/Generator/Generators/CLI/CLIMarshal.cs
```diff
@@ -40,13 +40,15 @@ public override bool VisitTagType(TagType tag, TypeQualifiers quals)
             return decl.Visit(this);
         }
 
+        private static int arrayNum = 0;
```
ddobrev: I believe I've already sent you a remark about this. A static value keeps increasing for the lifetime of the type - that is, for the whole run of the application. This means that for each new array ANYWHERE you'll increase this index.
We already have a mechanism to deal with this. Context has properties called ParameterIndex and MarshalVarPrefix. See how they are used and employ them in your code.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
genuinelucifer: Oh sorry I changed that without actually looking at Helpers.SafeIdentifier

src/Generator/Generators/CLI/CLIMarshal.cs
genuinelucifer: Thanks. I implemented something similar in CSharpTextTemplate but could not do same here. I will look into ParameterIndex and try to implement.

genuinelucifer:
Due to latest Merges this PR has now conflict merges and is incompilable on my machine too since i pulled the changes. So I am closing it and opening a new one with all the required changes.

Issue #402, opened 2015-01-07
Incorrect handling of virtual methods


tritao:
Consider this case:

C++:
```static Base getBase() { return Derived(); }```

C#:
```Base b = Base.getBase();
b.virtual();```

Since we're not generating code to go through the virtual table in `Base` we'll not have the expected polymorphic behaviour.

genuinelucifer:
@tritao I didn't quite understand it. Please explain this a bit...

ddobrev:
@genuinelucifer imagine that in the Base of the example you have a virtual function defines as follows:

int Base::virtualFunction()
{
    return 1;
}

Then in the Derived from the example you override this function:

int Base::virtualFunction()
{
    return 2;
}

At present, when you call b.virtualFunction() - as in the definition of the issue - you'll get 1. However, this is wrong - you should get 2 because the real C++ type that is returned is Derived.

ddobrev:
@genuinelucifer You need to use the same code we use for calls in abstract implementations:

https://github.com/mono/CppSharp/blob/master/src/Generator/Generators/CSharp/CSharpTextTemplate.cs#L2202

genuinelucifer:
@ddobrev Thanks. Will try to implement it.

genuinelucifer:
Please see #496. I have tried but the test is failing at one instance. It feels a bit weird to me.

PR #495, opened 2015-06-21
Added param to interface pass for Operator functions.

genuinelucifer:
Had to put one extra check for Conversion and Explicit conversion operators...
I don't know if this is okay or we should change the the parameterKind for such operators!

### DIFF
--- a/src/Generator/Passes/ParamTypeToInterfacePass.cs
+++ b/src/Generator/Passes/ParamTypeToInterfacePass.cs
@@ -1,4 +1,5 @@
-﻿using CppSharp.AST;
+﻿using System.Linq;
+using CppSharp.AST;
 using CppSharp.AST.Extensions;
 
 namespace CppSharp.Passes
@@ -7,20 +8,22 @@ namespace CppSharp.Passes
     {
         public override bool VisitFunctionDecl(Function function)
         {
-            if (!function.IsOperator)
+            if (!function.IsOperator || function.Parameters.Count > 1)
+                ChangeToInterfaceType(function.OriginalReturnType);
+
+            if (function.OperatorKind != CXXOperatorKind.Conversion && 
+                function.OperatorKind != CXXOperatorKind.ExplicitConversion)
             {
-                ChangeToInterfaceType(function.ReturnType);
-                foreach (Parameter parameter in function.Parameters)
-                {
+                foreach (var parameter in function.Parameters.Where(p => p.Kind != ParameterKind.OperatorParameter))
                     ChangeToInterfaceType(parameter.QualifiedType);
-                }
             }
+            
             return base.VisitFunctionDecl(function);
         }
 
         private static void ChangeToInterfaceType(QualifiedType type)
         {
-            var tagType = type.Type.GetFinalPointee() as TagType;
+            var tagType = (type.Type.GetFinalPointee() ?? type.Type) as TagType;
             if (tagType != null)
             {
                 var @class = tagType.Declaration as Class;
--- a/tests/CSharpTemp/CSharpTemp.Tests.cs
+++ b/tests/CSharpTemp/CSharpTemp.Tests.cs
@@ -261,4 +261,20 @@ public class CSharpTempTests : GeneratorTestFixture
             }
         }
     }
+
+    [Test]
+    public void TestParamTypeToInterfacePass()
+    {
+        var baseClass = new TestParamToInterfacePassBaseTwo();
+        baseClass++;
+        Assert.AreEqual(baseClass.M, 1);
+        ITestParamToInterfacePassBaseTwo baseInterface = new TestParamToInterfacePassBaseTwo();
+        var dervClass = new TestParamToInterfacePass();
+        dervClass.AddM(baseClass);
+        Assert.AreEqual(dervClass.M, 1);
+        dervClass = new TestParamToInterfacePass(dervClass + baseClass);
+        Assert.AreEqual(dervClass.M, 2);
+        dervClass = new TestParamToInterfacePass(dervClass + baseInterface);
+        Assert.AreEqual(dervClass.M, 2);
+    }
 }
--- a/tests/CSharpTemp/CSharpTemp.h
+++ b/tests/CSharpTemp/CSharpTemp.h
@@ -509,3 +509,27 @@ public:
     void VirtualMember();
     void setProperty(int value);
 };
+
+class DLL_API TestParamToInterfacePassBaseOne
+{
+};
+
+class DLL_API TestParamToInterfacePassBaseTwo
+{
+	int m;
+public:
+	int getM();
+	void setM(int n);
+	const TestParamToInterfacePassBaseTwo& operator++();
+	TestParamToInterfacePassBaseTwo();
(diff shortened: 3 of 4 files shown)


src/Generator/Passes/ParamTypeToInterfacePass.cs
```diff
@@ -7,20 +7,22 @@ public class ParamTypeToInterfacePass : TranslationUnitPass
     {
         public override bool VisitFunctionDecl(Function function)
         {
-            if (!function.IsOperator)
-            {
+            if (!function.IsOperator || function.Parameters.Count > 1)
```
ddobrev: Why have you left this check? What does it have to to with the return type?

src/Generator/Passes/ParamTypeToInterfacePass.cs
```diff
@@ -7,20 +7,22 @@ public class ParamTypeToInterfacePass : TranslationUnitPass
...
                 ChangeToInterfaceType(function.ReturnType);
```
ddobrev: It's not related to the current issue but now that you're changing the pass anyway, please use function.OriginalReturnType instead.

src/Generator/Passes/ParamTypeToInterfacePass.cs
```diff
@@ -7,20 +7,22 @@ public class ParamTypeToInterfacePass : TranslationUnitPass
...
-                foreach (Parameter parameter in function.Parameters)
-                {
-                    ChangeToInterfaceType(parameter.QualifiedType);
-                }
-            }
+
+            if(function.OperatorKind != CXXOperatorKind.Conversion &&
```
ddobrev: Please align this whole block.

src/Generator/Passes/ParamTypeToInterfacePass.cs
```diff
@@ -7,20 +7,22 @@ public class ParamTypeToInterfacePass : TranslationUnitPass
...
+                                    function.OperatorKind != CXXOperatorKind.ExplicitConversion)
+                     foreach (var parameter in function.Parameters)
```
ddobrev: I don't understand, why have you not added the check for OperatorParameter. If you had a test - please add one - it would've failed.

src/Generator/Passes/ParamTypeToInterfacePass.cs
genuinelucifer: The unary operators must return the containing type while Binary operators can return other types.
Without this check Build was failing in CSharpTemp.cs at operator ++ in Bar.

src/Generator/Passes/ParamTypeToInterfacePass.cs
genuinelucifer: Ok. Will do.

src/Generator/Passes/ParamTypeToInterfacePass.cs
genuinelucifer: Ok. Will do.

src/Generator/Passes/ParamTypeToInterfacePass.cs
genuinelucifer: Sorry, i copied few changes between branches and that part didn't got copied. I have corrected it now.

genuinelucifer:
@ddobrev I have added the classes and fixed the ParamTypeToInterfacePass code but when writing the test how should I check for the types to be an interface?
Also, it would be much more easy to write test in CSharpTemp.tests because from there i can directly access all the classes. Writing the tests in Generator.Tests is proving dificult.
Can I write the test in CSharpTemp.tests?

src/Generator/Passes/ParamTypeToInterfacePass.cs
ddobrev: I meant not just the conditional but its body too.

genuinelucifer: :+1:   WIll do in the next commit.

ddobrev:
@genuinelucifer I was actually going to ask you myself to move the test to CSharpTemp. We only use Generator.Tests when we cannot check what we need at run-time - for example line numbers or dependencies. For this issue you'd get incompilable code if there's a problem so using the regular test projects is enough.

genuinelucifer:
@ddobrev Thanks. Will move it. I thought Generator Tests were for all passes. Thanks for the info.

genuinelucifer:
@ddobrev Please check it again. I finally merged everything into one commit and added a test too.

tests/CSharpTemp/CSharpTemp.h
```diff
@@ -509,3 +509,27 @@ class DLL_API TestOverrideFromSecondaryBase : public Foo, public SecondaryBase
...
+
+class DLL_API TestParamToInterfacePassBaseOne
+{
+};
+
+class DLL_API TestParamToInterfacePassBaseTwo
+{
+	int m;
+public:
+	int getM() { return m; }
```
ddobrev: Could you please move all inlined bodies to the cpp file?

tests/CSharpTemp/CSharpTemp.Tests.cs
```diff
@@ -261,4 +261,20 @@ public void TestNotDestroyingForeignObjects()
             }
         }
     }
+
+    [Test]
+    public void TestParamTypeToInterfacePass()
+    {
+        var baseClass = new TestParamToInterfacePassBaseTwo();
+        baseClass++;
+        Assert.AreEqual(baseClass.M,1);
```
ddobrev: Please put a space after each comma.

ddobrev:
Thanks. I've told @tritao to merge it. I would've done so earlier but you only changed the pull without adding a comment so that I was notified.

genuinelucifer:
Oh. Sorry. After doing the changes you told, I rebased both commits to
single. I should have commented. Will remember in future. :)

ddobrev:
@genuinelucifer @tritao 's (finally :)) restoring the build server and he'd like to complete that before merging this pull. So please move on to https://github.com/mono/CppSharp/issues/402 and he'll merge this when done.

genuinelucifer:
Great! I will start the other issue.

tests/CSharpTemp/CSharpTemp.h
```diff
@@ -509,3 +509,27 @@ class DLL_API TestOverrideFromSecondaryBase : public Foo, public SecondaryBase
...
     void setProperty(int value);
 };
+
+class DLL_API TestParamToInterfacePassBaseOne
+{
+};
+
+class DLL_API TestParamToInterfacePassBaseTwo
+{
+	int m;
```
tritao: I think you need to initialize m to 0 since you're checking its value later with an assert.

ddobrev: You don't initialise this when using the empty ctor. For some reason it happens to work but we cannot rely on that. Please assign it to 0 in the empty ctor.

genuinelucifer: Done :)
I assumed zero as default value for integers. But, I remember now that it's only for global variables. :P

Issue #365, opened 2014-11-03
Static methods can create name-space conflicts when wrapped with C++/CLI


ddobrev:
To reproduce, go to CLITemp.h and move testStdString from its class to a free function.

genuinelucifer:
I reproduced the issue and fixed it by generating a 'FileName + "Class" ' for all the free functions instead of the 'FileName' used earlier. I made the changes only for CLI generator.
But, I guess it would be better to be uniform across CSharp and CLI generators regarding naming conventions. So, please give review my work and if it is acceptable i will add the changes to CSharp generator as well.
My work:
https://github.com/genuinelucifer/CppSharp/commit/8b01d4a9702c5702af309c3824f86de6fb8a1095

genuinelucifer:
@ddobrev also please see if FIleName + "Class" solution is acceptable only as a change in C++/CLI generator! SInce, the C# generator works fine without any changes so is it fine to leave it be?

ddobrev:
I am afraid @tritao has the final say here as well. I am not really familiar with the C++/CLI end. What I do know, I've already mentioned.

tritao:
We try to keep the behavior uniform between the C# and C++/CLI backends so it would be better to fix this so it works the same between backends. 

Just going from the description I cannot easily see what the problem is though. What is the actual problem here @ddobrev?

ddobrev:
@tritao I don't remember, I would need to see some generated code. I hit that by accident when writing a test for something completely unrelated and filed this so that it's not forgotten. What I meant was that the "Class" appendage is not appropriate because the functions in question are classless.

genuinelucifer:
@tritao The problem is that for free functions (not contained in any class)
in the header a new class is created with same name as the file and since
there is a top level namespace created with the same name as the file so
these both names result in conflict.
Ex:
If ClassA and ClassB are in Basic.h and a free function "Func" is also
there then we generate a namespace Basic which contains the two classes AND
a class named Basic which has the "Func".
Now, we reference ClassB from ClassA with Basic::ClassB which results in
the error that ClassB is not a member of Basic. Because now the class Basic
is in scope NOT the namespace.
So, either we rename the namespace or the clas that is generated for free
function. (this seems to be one obvious solution)

tritao:
https://github.com/mono/CppSharp/commit/cf472f526c87e9807a9f9868e7cbc07376efcbb6 should have fixed this.

PR #496, opened 2015-06-22
Added support for Virtual function handling and a test. TEST FAILING!!

genuinelucifer:
I am opening this for discussion. I have no experience with C# delegates so I maybe missing something very basic.
@ddobrev @tritao Please review and help.

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -2114,6 +2114,10 @@ namespace CppSharp.Generators.CSharp
                 {
                     GenerateAbstractImplCall(method, @class);
                 }
+                else if (method.IsVirtual)
+                {
+                    GenerateVirtualFunctionCall(method, @class);
+                }
                 else
                 {
                     GenerateInternalFunctionCall(method);
@@ -2197,6 +2201,40 @@ namespace CppSharp.Generators.CSharp
                     return method.IsOverride ?
                         ((Class) method.Namespace).GetRootBaseMethod(method).Access : method.Access;
             }
+        }
+
+        private void GenerateVirtualFunctionCall(Function function, Class @class)
+        {
+            string delegateId, @delegate;
+            Write(GetVirtualCallDelegate(function, @class, out delegateId, out @delegate));
+            WriteLine("if({0} == {1}.{2})", @delegateId, @class.QualifiedName, @delegate + "Instance");
+            WriteStartBraceIndent();
+            GenerateInternalFunctionCall(function);
+            WriteCloseBraceIndent();
+            WriteLine("else");
+            WriteStartBraceIndent();
+            GenerateFunctionCall(delegateId, function.Parameters, function);
+            WriteCloseBraceIndent();
+        }
+
+        public string GetVirtualCallDelegate(Function function, Class @class,
+            out string delegateId,out string @delegate)
+        {
+            var virtualCallBuilder = new StringBuilder();
+            var i = VTables.GetVTableIndex(function, @class);
+            virtualCallBuilder.AppendFormat("void* slot = *(void**) ((({0}.Internal*) {1})->vfptr0 + {2} * {3});",
+                (@class.BaseClass ?? @class).Name, Helpers.InstanceIdentifier, i, Driver.TargetInfo.PointerWidth / 8);
+            virtualCallBuilder.AppendLine();
+
+            @delegate = GetVTableMethodDelegateName(function.OriginalFunction ?? function);
+            delegateId = Generator.GeneratedIdentifier(@delegate);
+
+            virtualCallBuilder.AppendFormat(
+                "var {1} = ({0}) Marshal.GetDelegateForFunctionPointer(new IntPtr(slot), typeof({0}));",
+                @delegate, delegateId);
+
+            virtualCallBuilder.AppendLine();
+            return virtualCallBuilder.ToString();
         }
 
         private void GenerateAbstractImplCall(Function function, Class @class)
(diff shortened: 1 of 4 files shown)


genuinelucifer:
What I did is that I am checking if the delegate to be called is same as that of the current class. If so, then I am simply calling the normally generated function call with GenerateInternalFunctionCall() else it is calling the delegate function generated from GenerateFunction()

genuinelucifer:
I think this line:
```C#
return BaseClassVirtual.__CreateInstance(__ret);
```
generated in Base property is causing test failure. Because if an instance of BaseClassVirtual is created then it would return delegate of same class only. (wild guess)

ddobrev:
What is the failure? Does it crash or is it just incorrect?

genuinelucifer:
The test simply fails because the value don't match at last comparision. Expected 2 was 1 [sort of]
.
There is no build failure.

ddobrev:
That's really strange. @tritao , can you help here?
@genuinelucifer in the mean time could you please try reusing the GenerateAbstractImplCall method? It seems to me it has the same body as the new method you've written.

genuinelucifer:
I simply copied the method and made a few changes. Using exactly that method gave build failure due to NullPointerException.

ddobrev:
OK, we can see about that later. Let's first see if @tritao has any idea about the error.

genuinelucifer:
@ddobrev Till @tritao  sees this issue, could you pls see #365 I did some work few days ago. I would like to finish that too.

ddobrev:
Can you please paste me the generated code you get for Base::retInt()?

genuinelucifer:
For BaseClassVirtual::RetInt
```C#
public virtual int retInt()
        {
            void* slot = *(void**) (((BaseClassVirtual.Internal*) __Instance)->vfptr0 + 0 * 8);
            var ___retInt_0Delegate = (_retInt_0Delegate) Marshal.GetDelegateForFunctionPointer(new IntPtr(slot), typeof(_retInt_0Delegate));
            if(___retInt_0Delegate.GetType().Equals(new  BaseClassVirtual))
            {
                var __ret = Internal.retInt_0(__Instance);
                return __ret;
            }
            else
            {
                var __ret = ___retInt_0Delegate(__Instance);
                return __ret;
            }
        }
```

ddobrev:
Please remove the conditional. The whole idea is that this should work correctly just with the block in your "else".

genuinelucifer:
I used the conditional like this:
If function from other class is to be called then--
   call the code as if this function was pure virtual
If this class has to be called then--
   call the code as if this function is normal function...
.
Well, i will try to remove the conditional and then try...

ddobrev:
@genuinelucifer any better now?

genuinelucifer:
@ddobrev
Umm not yet. I could not think of a way to remove the if conditions and
make it work. Removing them caused stack overflow...
Because if a virtual function is called it keeps recursively calling itself.
How to decide if it is being called from the same class and no need for
furthur virtuak table check?
We need atleast some condition to decide when to call the function body. I
tried few other things but no success yet.

ddobrev:
Right, I didn't think of that. I am afraid I don't have other ideas right away and I am busy at the moment. Perhaps @tritao can help.

tritao:
I cannot check this at the moment, I will try to check later tonight.

ddobrev:
@genuinelucifer you've basically correctly judged that you need to check for the bottom of the recursion. However, you cannot do it by type. You'd rather need some way of checking if the delegate you obtained points to the same method you're in which might not be possible.

ddobrev:
I think the problem is actually at:

(((BaseClassVirtual.Internal*) __Instance)->vfptr0 + 0 * 8)

This returns the pointer to the v-table of BaseClassVirtual. However, for the approach to work, we need to get the v-table of DerivedClassVirtual in this case. In other words, we need to extract from __Instance, which is the only clue we have to the real object, its v-table pointer. However, I have no idea at the moment how we should do it.

tritao:
I think @ddobrev's on to something. We need to adjust the vtable pointer to point to the correct sub type. I think for MS ABI we could use `VFTableInfo.VFPtrFullOffset`, though for Itanium I don't remember very well how it works. I'll read up on it and think of the correct solution.

ddobrev:
@genuinelucifer while @tritao checks it, you could try this:

            void* slot = *(void**) (__Instance + 0 * 8);

The idea is that the v-table pointer of a class is always at offset 0.

genuinelucifer:
@ddobrev That resulted in some 
```C#
System.AccessViolationException : Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
```
Also, i tried many other things to compare delegates but after reading Microsoft's documentation it seems we can't test delegates for equality in our case.
@tritao I could not think of how to use VFTableInfo.VFPtrFullOffset ... Please provide a little more info on how to utilise it.

ddobrev:
@genuinelucifer could you please try debugging the generated code and see what __Instance is like. How many v-table pointers (vfptr-like fields) it has, at what offsets, etc. It would help if you cast __Instance to DerivedClassVirtual* so that you can inspect all fields. Basically, search for any clue how you can be sure you get the correct v-table ptr.

genuinelucifer:
@ddobrev I had not used Visual Studio debugger in past. So, I had to learn about the basics of it.
I will debug the generated code and let you know.

tritao:
Hey Abhinav, I took your pull as a base and spent some time ironing out the final issues.

It turns out that the test case you were stuck on is actually expected behavior on C++, see http://stackoverflow.com/questions/3479712/virtual-functions-object-slicing.

Apart from that there were a few bugs to fix in our vtable saving code that made us return wrong results for some calls.

Thanks for your work on this.

genuinelucifer:
Great :)
Feels relieving to know that I was actually on the right path. Thanks to you for this!

PR #501, opened 2015-06-26
Structible classes

genuinelucifer:
I guess everything else is fine. I just couldn't find how to check if a class has been defined by 'class' keyword or 'struct'!! Please review and help. @ddobrev @tritao


ddobrev:
There's a function in Clang we do not use yet that can give you that information. Please add a new property Class.IsStruct and then fill its value by using clang::RecordDecl::isStruct() in Parser::WalkRecord. Make it work this way for the time being. Then we'll think about whether to use both that and IsValueType or stick to just one.

genuinelucifer:
@ddobrev Thanks. Will look into the clang docs. Also, will try to implement
it this way.

genuinelucifer:
Closing this since changing the way of implementation...

PR #505, opened 2015-06-28
Added option for generating 'struct' for types defined with 'struct' keyword wherever possible.

genuinelucifer:
Added everything and as faar as I can see it is working fine.
Please review @ddobrev @tritao

### DIFF
--- a/src/AST/Class.cs
+++ b/src/AST/Class.cs
@@ -188,7 +188,7 @@ namespace CppSharp.AST
 
         public bool IsValueType
         {
-            get { return Type == ClassType.ValueType || IsUnion; }
+            get { return Type == ClassType.ValueType || IsUnion || IsStruct; }
         }
 
         public bool IsRefType
--- a/src/AST/Class.cs
+++ b/src/AST/Class.cs
@@ -102,7 +102,7 @@ namespace CppSharp.AST
         // True if the class represents a static class.
         public bool IsStatic;
 
-        // True is declared with 'struct' keyword
+        // True if declared with 'struct' keyword
         public bool DeclaredStruct;
 
         public Class()
--- a/src/AST/Class.cs
+++ b/src/AST/Class.cs
@@ -193,7 +193,7 @@ namespace CppSharp.AST
 
         public bool IsRefType
         {
-            get { return Type == ClassType.RefType && !IsUnion; }
+            get { return Type == ClassType.RefType && !IsUnion && !IsStruct; }
         }
 
         public bool IsInterface
--- a/src/AST/Class.cs
+++ b/src/AST/Class.cs
@@ -102,9 +102,6 @@ namespace CppSharp.AST
         // True if the class represents a static class.
         public bool IsStatic;
 
-        // True if declared with 'struct' keyword
-        public bool DeclaredStruct;
-
         public Class()
         {
             Bases = new List<BaseClassSpecifier>();
--- a/src/AST/Class.cs
+++ b/src/AST/Class.cs
@@ -102,6 +102,9 @@ namespace CppSharp.AST
         // True if the class represents a static class.
         public bool IsStatic;
 
+        // True is declared with 'struct' keyword
+        public bool DeclaredStruct;
+
         public Class()
         {
             Bases = new List<BaseClassSpecifier>();
@@ -113,6 +116,7 @@ namespace CppSharp.AST
             IsUnion = false;
             IsOpaque = false;
             IsPOD = false;
+            IsStruct = false;
             Type = ClassType.RefType;
             Layout = new ClassLayout();
         }
@@ -137,6 +141,7 @@ namespace CppSharp.AST
             HasNonTrivialCopyConstructor = @class.HasNonTrivialCopyConstructor;
             HasNonTrivialDestructor = @class.HasNonTrivialDestructor;
             IsStatic = @class.IsStatic;
+            IsStruct = @class.IsStruct;
         }
 
         public bool HasBase
@@ -147,6 +152,12 @@ namespace CppSharp.AST
         public bool HasBaseClass
         {
             get { return BaseClass != null; }
+        }
+
+        public bool IsStruct
+        {
+            get;
+            set;
         }
 
         public Class BaseClass
--- a/src/AST/PropertyExtensions.cs
+++ b/src/AST/PropertyExtensions.cs
@@ -14,7 +14,7 @@ namespace CppSharp.AST
             type = type ?? p.Field.Type;
 
             Class decl;
-            return type.TryGetClass(out decl) && (decl.IsValueType || decl.IsStruct);
+            return type.TryGetClass(out decl) && decl.IsValueType;
         }
     }
 }
--- a/src/AST/PropertyExtensions.cs
+++ b/src/AST/PropertyExtensions.cs
@@ -14,7 +14,7 @@ namespace CppSharp.AST
             type = type ?? p.Field.Type;
 
             Class decl;
-            return type.TryGetClass(out decl) && decl.IsValueType;
+            return type.TryGetClass(out decl) && (decl.IsValueType || decl.IsStruct);
         }
     }
 }
--- a/src/CppParser/AST.h
+++ b/src/CppParser/AST.h
@@ -716,7 +716,7 @@ public:
     bool HasNonTrivialCopyConstructor;
     bool HasNonTrivialDestructor;
     bool IsExternCContext;
-    bool DeclaredStruct;
+    bool IsStruct;
 
     ClassLayout* Layout;
 };
--- a/src/CppParser/AST.h
+++ b/src/CppParser/AST.h
@@ -716,6 +716,7 @@ public:
     bool HasNonTrivialCopyConstructor;
     bool HasNonTrivialDestructor;
     bool IsExternCContext;
+	bool DeclaredStruct;
 
     ClassLayout* Layout;
 };
--- a/src/CppParser/AST.h
+++ b/src/CppParser/AST.h
@@ -716,7 +716,7 @@ public:
     bool HasNonTrivialCopyConstructor;
     bool HasNonTrivialDestructor;
     bool IsExternCContext;
-	bool DeclaredStruct;
+    bool DeclaredStruct;
 
     ClassLayout* Layout;
 };
--- a/src/CppParser/Bindings/CLI/AST.cpp
+++ b/src/CppParser/Bindings/CLI/AST.cpp
@@ -2835,9 +2835,9 @@ void CppSharp::Parser::AST::Class::IsExternCContext::set(bool value)
     ((::CppSharp::CppParser::AST::Class*)NativePtr)->IsExternCContext = value;
 }
 
-bool CppSharp::Parser::AST::Class::DeclaredStruct::get()
+bool CppSharp::Parser::AST::Class::IsStruct::get()
 {
-    return ((::CppSharp::CppParser::AST::Class*)NativePtr)->DeclaredStruct;
+    return ((::CppSharp::CppParser::AST::Class*)NativePtr)->IsStruct;
 }
 
 CppSharp::Parser::AST::ClassLayout^ CppSharp::Parser::AST::Class::Layout::get()
--- a/src/CppParser/Bindings/CLI/AST.cpp
+++ b/src/CppParser/Bindings/CLI/AST.cpp
@@ -2837,7 +2837,7 @@ void CppSharp::Parser::AST::Class::IsExternCContext::set(bool value)
 
 bool CppSharp::Parser::AST::Class::DeclaredStruct::get()
 {
-	return ((::CppSharp::CppParser::AST::Class*)NativePtr)->DeclaredStruct;
+    return ((::CppSharp::CppParser::AST::Class*)NativePtr)->DeclaredStruct;
 }
 
 CppSharp::Parser::AST::ClassLayout^ CppSharp::Parser::AST::Class::Layout::get()
--- a/src/CppParser/Bindings/CLI/AST.cpp
+++ b/src/CppParser/Bindings/CLI/AST.cpp
@@ -2835,6 +2835,11 @@ void CppSharp::Parser::AST::Class::IsExternCContext::set(bool value)
     ((::CppSharp::CppParser::AST::Class*)NativePtr)->IsExternCContext = value;
 }
 
+bool CppSharp::Parser::AST::Class::DeclaredStruct::get()
+{
+	return ((::CppSharp::CppParser::AST::Class*)NativePtr)->DeclaredStruct;
+}
+
 CppSharp::Parser::AST::ClassLayout^ CppSharp::Parser::AST::Class::Layout::get()
 {
     return (((::CppSharp::CppParser::AST::Class*)NativePtr)->Layout == nullptr) ? nullptr : gcnew CppSharp::Parser::AST::ClassLayout((::CppSharp::CppParser::AST::ClassLayout*)((::CppSharp::CppParser::AST::Class*)NativePtr)->Layout);
--- a/src/CppParser/Bindings/CLI/AST.h
+++ b/src/CppParser/Bindings/CLI/AST.h
@@ -1840,7 +1840,7 @@ namespace CppSharp
                     void set(bool);
                 }
 
-                property bool DeclaredStruct
+                property bool IsStruct
                 {
                     bool get();
                 }
--- a/src/CppParser/Bindings/CLI/AST.h
+++ b/src/CppParser/Bindings/CLI/AST.h
@@ -1840,6 +1840,11 @@ namespace CppSharp
                     void set(bool);
                 }
 
+				property bool DeclaredStruct
+				{
+					bool get();
+				}
+
                 property CppSharp::Parser::AST::ClassLayout^ Layout
                 {
                     CppSharp::Parser::AST::ClassLayout^ get();
--- a/src/CppParser/Bindings/CLI/AST.h
+++ b/src/CppParser/Bindings/CLI/AST.h
@@ -1840,10 +1840,10 @@ namespace CppSharp
                     void set(bool);
                 }
 
-				property bool DeclaredStruct
-				{
-					bool get();
-				}
+                property bool DeclaredStruct
+                {
+                    bool get();
+                }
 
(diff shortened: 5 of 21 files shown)


genuinelucifer:
Had to change two tests from 'struct' to 'class' because they were generating erroneous code when marshalled with 'struct' keyword.

src/AST/PropertyExtensions.cs
```diff
@@ -14,7 +14,7 @@ public static bool IsInRefTypeAndBackedByValueClassField(this Property p)
             type = type ?? p.Field.Type;
 
             Class decl;
-            return type.TryGetClass(out decl) && decl.IsValueType;
+            return type.TryGetClass(out decl) && (decl.IsValueType || decl.IsStruct);
```
ddobrev: It seems to me that you use this || expression any time IsValueType is used. In that case why don't you move || decl.IsStruct to the getter of IsValueType analogous to what you did for IsRefType? This way we can continue checking just IsValueType and we'll also have many less changes in this pull.

src/CppParser/AST.h
```diff
@@ -716,6 +716,7 @@ class CS_API Class : public DeclarationContext
     bool HasNonTrivialCopyConstructor;
     bool HasNonTrivialDestructor;
     bool IsExternCContext;
+	bool DeclaredStruct;
```
ddobrev: This is aligned with a tab instead of with 4 spaces as per convention.

src/CppParser/Bindings/CLI/AST.cpp
```diff
@@ -2835,6 +2835,11 @@ void CppSharp::Parser::AST::Class::IsExternCContext::set(bool value)
     ((::CppSharp::CppParser::AST::Class*)NativePtr)->IsExternCContext = value;
 }
 
+bool CppSharp::Parser::AST::Class::DeclaredStruct::get()
+{
+	return ((::CppSharp::CppParser::AST::Class*)NativePtr)->DeclaredStruct;
```
ddobrev: Misaligned.

src/CppParser/Bindings/CLI/AST.h
```diff
@@ -1840,6 +1840,11 @@ namespace CppSharp
                     void set(bool);
                 }
 
+				property bool DeclaredStruct
```
ddobrev: Misaligned.

src/CppParser/Parser.cpp
```diff
@@ -685,6 +685,7 @@ void Parser::WalkRecord(clang::RecordDecl* Record, Class* RC)
     RC->IsUnion = Record->isUnion();
     RC->IsDependent = Record->isDependentType();
     RC->IsExternCContext = Record->isExternCContext();
+	RC->DeclaredStruct = Record->isStruct();
```
ddobrev: Misaligned.

src/Generator/Generators/CLI/CLIHeadersTemplate.cs
```diff
@@ -283,6 +286,34 @@ public void GenerateClass(Class @class)
...
+
+             if (@class.IsValueType)
+                 return true;
+             if (@class.IsInterface)
+                 return false;
+             if (@class.IsStatic)
+                 return false;
+             if (@class.IsAbstract)
+                 return false;
+             if (@class.HasBase)
```
ddobrev: Are you sure this is necessary? C# structs can inherit from other classes, they just cannot be inherited from.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -353,6 +353,12 @@ public void GenerateInlineSummary(RawComment comment)
 
         public void GenerateClass(Class @class)
         {
+            if(@class.DeclaredStruct)
```
ddobrev: Please add a space after the 'if'.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -353,6 +353,12 @@ public void GenerateInlineSummary(RawComment comment)
...
+            {
+                WriteLine("//Declared Struct");
```
ddobrev: Not that it's a problem but what does this help with?

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -434,6 +440,34 @@ private void GenerateClassMarshals(Class @class)
             NewLine();
         }
 
+         private bool CheckClassIsStructible(Class @class)
```
ddobrev: You have written this method for the second time. Please move it to some common to both generators location, such as ClassExtensions.

tests/Basic/Basic.h
```diff
@@ -486,8 +486,9 @@ DLL_API int Function()
 }
 
 // Tests properties
-struct DLL_API TestProperties
+class DLL_API TestProperties
```
ddobrev: Please add a few words explaining why this type has to be a class. I just want to be sure there's no hidden bug revealed by this test.

tests/CSharpTemp/CSharpTemp.h
```diff
@@ -175,8 +175,9 @@ class DLL_API P : Proprietor
 };
 
 // Tests destructors
-struct DLL_API TestDestructors
+class DLL_API TestDestructors
```
ddobrev: Same as the previous test: please add a few words to explain this change.

genuinelucifer:
Instead of changing the tests I thought it's better to change the allocation of 'IsStruct' property.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
genuinelucifer: Sorry for this one, I forgot to erase it. I used it just to test if 'DeclaredStruct' was properly initialised.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -434,6 +436,38 @@ private void GenerateClassMarshals(Class @class)
...
+                 return true;
+             if (@class.IsInterface)
+                 return false;
+             if (@class.IsStatic)
+                 return false;
+             if (@class.IsAbstract)
+                 return false;
+             if (@class.HasBase)
+                 return false;
+             if (@class.Methods.Any(m => m.IsStatic))
```
ddobrev: C# structs are allowed to have static members and operators so this fix is incorrect. If the only problem with those tests is their static members, there's some other bug that needs fixing.

src/Generator/Generators/CLI/CLIHeadersTemplate.cs
genuinelucifer: Sorry. I don't even remember now why I added that at first. Will remove it.

src/AST/PropertyExtensions.cs
genuinelucifer: I first thought IsValueType and IsStruct were mean to be in different contexts. But now yes they seem to be same. Will move it.

genuinelucifer:
@ddobrev Yes, the build was failing due to obe of the static members in ValueType..
Will try to fix it...

src/Generator/Generators/ExtensionMethods.cs
```diff
@@ -23,5 +24,35 @@ public static Interop.CallingConvention ToInteropCallConv(this CallingConvention
...
+                return false;
+            if (@class.Methods.Any(m => m.IsStatic))
+                return false;
+            if (@class.Methods.Any(m => m.IsOperator))
+                return false;
+
+            var allTrUnits = Driver.ASTContext.TranslationUnits;
+            foreach (var trUnit in allTrUnits)
+            {
+                foreach (var cls in trUnit.Classes)
```
genuinelucifer: Please verify once that it is correct way to loop through ALL the classes. I am not sure but have found no error so far.

src/AST/Class.cs
```diff
@@ -102,6 +102,9 @@ public class Class : DeclarationContext
         // True if the class represents a static class.
         public bool IsStatic;
 
+        // True if declared with 'struct' keyword
+        public bool DeclaredStruct;
```
ddobrev: You haven't renamed this to "IsStruct" as I asked.

src/Generator/Generators/ExtensionMethods.cs
ddobrev: Looks fine to me.

src/AST/Class.cs
genuinelucifer: I have used 'DeclaredStruct' which is filled by AST just to check if while declaring this 'struct' keyword was used.
'IsStruct' is used to determine if the type follows all rules and should be marshalled to 'struct'...
.
Should I combine them both?

ddobrev: A good question. I think it it all works, you'd better combine them. Also, now that you're at it, please make it a property (best if automatic) and not a field. I know the rest are field but it's a bad practice and we should eliminate it over time.

src/AST/Class.cs
```diff
@@ -149,6 +154,12 @@ public bool HasBaseClass
             get { return BaseClass != null; }
         }
 
+        public bool IsStruct
+        {
+            get;
+            set;
```
genuinelucifer: @ddobrev By automatic property, do you mean something like this?

ddobrev: Exactly.

src/AST/Class.cs
genuinelucifer: Wouldn't it seem erroneous to see some code like:
```C#
@class.IsStruct = CheckClassIsStructible(@class, Driver);
```
while the function itself uses the same property as:
```C#
if(!@class.IsStruct)
    return false;
```

ddobrev: I wouldn't say so. Basically, C# structs are limited compared to C++ structs so this code simply means that we apply these limitations on top if the type is a (C++) struct.

genuinelucifer: Ohh. Fine.

genuinelucifer:
Although i combined them but now the CLI generator is generating erroneous code in Basic.cpp 'Exception' and STL.cpp
I am looking for the cause.

ddobrev:
In case you have problem with this, wouldn't setting of Class.Type to ClassType.ValueType instead of setting IsStruct to true, work better?

genuinelucifer:
@ddobrev Thanks. Using that and a few other bits reduced the errors a lot. I am working on removing the left errors. For that we need some changes which I am wroking on in #506 .

genuinelucifer:
Once #506 is merged, i will open a fresh pull with all the working changes.
But I had to change the TestDestructors test from 'struct' to 'class' because structs can't have destructors in C#.
Should I improve the check to include this or we could provide a new macro CS_REF_TYPE.. as CS_VALUE_TYPE marshals a C++ class to C# struct.
CS_REF_TYPE could be used to force marshaling a C++ struct to C# class...

src/Generator/Generators/CLI/CLIHeadersTemplate.cs
```diff
@@ -137,6 +137,9 @@ public void GenerateForwardRefs()
             var typeReferences = typeReferenceCollector.TypeReferences;
             var @namespace = ConvertForwardReferencesToNamespaces(typeReferences);
 
+            foreach (var cls in @namespace.Classes)
+                cls.IsStruct = ExtensionMethods.CheckClassIsStructible(cls, Driver);
```
tritao: We should not be modifying the AST in the text generators. This should be done in a pass.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -353,6 +353,8 @@ public void GenerateInlineSummary(RawComment comment)
 
         public void GenerateClass(Class @class)
         {
+            @class.IsStruct = ExtensionMethods.CheckClassIsStructible(@class, Driver);
```
tritao: We should not be modifying the AST in the text generators. This should be done in a pass.
 Add a line note

src/Generator/Generators/ExtensionMethods.cs
```diff
@@ -23,5 +24,30 @@ public static Interop.CallingConvention ToInteropCallConv(this CallingConvention
...
+        public static bool CheckClassIsStructible(Class @class, Driver Driver)
+        {
+            if (!@class.IsStruct)
+                return false;
+            if (@class.IsInterface || @class.IsStatic || @class.IsAbstract)
+                return false;
+            if (@class.Methods.Any(m => m.IsOperator))
+                return false;
+
+            if (@class.IsValueType)
```
tritao: I think this check should be done first and have priority to the other checks.

src/Generator/Generators/ExtensionMethods.cs
```diff
@@ -23,5 +24,30 @@ public static Interop.CallingConvention ToInteropCallConv(this CallingConvention
...
             return Interop.CallingConvention.Winapi;
         }
+
+        public static bool CheckClassIsStructible(Class @class, Driver Driver)
+        {
+            if (!@class.IsStruct)
+                return false;
+            if (@class.IsInterface || @class.IsStatic || @class.IsAbstract)
+                return false;
+            if (@class.Methods.Any(m => m.IsOperator))
```
tritao: Why can't it be a struct if there are any operators? As far as I know C# structs also support operators.

tritao:
Overall I don't think this approach makes sense, we can't just assume something is a struct just by looking if it was originally declared as a struct in C++. We should be instead checking its structure (operators, inheritance chain, members) and if it's a POD type.

ddobrev:
What doesn't make sense? He does make all of the mentioned checks.

src/Generator/Generators/ExtensionMethods.cs
ddobrev: Right, I've told already told him about it, not sure why he hasn't removed it yet.

src/Generator/Generators/CLI/CLIHeadersTemplate.cs
genuinelucifer: Should I add a new pass? I can't find any pass that would be fit to accommodate this.
Also, I can add a new file for the pass but do I actually get it working? Please tell me the file where it is defined as to which passes will be applicable and which not for a certain test project.

src/Generator/Generators/ExtensionMethods.cs
genuinelucifer: Yes, I actually have changed many things from this pull but in a separate branch. I will do that here as well.

src/Generator/Generators/CLI/CLIHeadersTemplate.cs
ddobrev: I think you'd best move the whole logic about checking whether a type can be represented as a struct, to a new pass.

src/Generator/Generators/ExtensionMethods.cs
genuinelucifer: Sorry, I was experimenting with different options and changes in a separate branch. I will apply the changes here as well.

genuinelucifer:
@tritao I have pushed a couple of changes here. But, I will need the changes of #506 for these to work correctly. I will submit another PR with fresh changes.
.
Also, I too don't see why it is not making sense?
We have all the proper checks here before deciding. The 'struct' keyword is just one of those checks!

src/Generator/Generators/CLI/CLIHeadersTemplate.cs
genuinelucifer: Ok. Will do.
But how do I make sure that pass is applied to say Basic.h in the tests!?

tests/CSharpTemp/CSharpTemp.h
```diff
@@ -175,8 +175,9 @@ class DLL_API P : Proprietor
 };
 
 // Tests destructors
-struct DLL_API TestDestructors
+class DLL_API TestDestructors                                  //Value Types can't have destructors
 {
+public:
```
genuinelucifer: @ddobrev @tritao I know I should not have changed this. But, I tried a few things but could not find how to check if a 'destructor' or a 'zero argument Constructor' has been "explicitly defined" by the user! Please help.

tests/CSharpTemp/CSharpTemp.h
```diff
@@ -175,8 +175,9 @@ class DLL_API P : Proprietor
 };
 
 // Tests destructors
-struct DLL_API TestDestructors
+class DLL_API TestDestructors                                  //Value Types can't have destructors
```
genuinelucifer: @ddobrev @tritao I know I should not have changed this. But, I tried a few things but could not find how to check if a 'destructor' or a 'zero argument Constructor' has been "explicitly defined" by the user! Please help.

src/Generator/Generators/CLI/CLIHeadersTemplate.cs
ddobrev: You just add the pass to the list of passes in Driver.SetupPasses.

genuinelucifer: Thanks. Will try to implement it.

tests/CSharpTemp/CSharpTemp.h
tritao: I think you can use clang's `Decl::isImplicit()` to check if the declaration was explicitly declared by the user.

tritao:
How a type was written in native code (whether with `class` or `struct` keyword) should not make any difference to how its classified (`ref` vs `value` type). We should solely look at its structure and usage to make that determination.

tests/CSharpTemp/CSharpTemp.h
genuinelucifer: Thanks... will try that.

genuinelucifer:
@tritao I have added a new pass to work for that and also I have removed the checking of 'struct' keyword in a separate branch. Please merge #506 first if you have no objections on the way that is done.
Then I will open the pull for this because I really need the changes on that pull.
Till then I am closing it since it is obsolete I looked at the changes and they are confusing here.

ddobrev:
@tritao I think it should make a difference. The way we do it now does not check if a type's been written as a struct, it's actully the other way around. It's checked if it's a class - in which case it's left alone. If it's written as a struct, only then we apply the checks to see if it can be a struct.
@genuinelucifer when you get back to this issue, please add an additional check - if the type has any virtual functions.

PR #506, opened 2015-06-29
Fixed marshalling base members from ref type to value type.

genuinelucifer:
The code added in Basic.h generated incompilable code.
I have fixed the CSharp side. But the CLI generator still gives "value types can only inherit from interface classes" error.
Trying to fix it because it will be helpful in the other pull of marshalling structs to struct.

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -765,7 +765,7 @@ namespace CppSharp.Generators.CSharp
             return Tuple.Create(library, decl.Mangled);
         }
 
-        private void GeneratePropertySetter<T>(T decl, Class @class, bool isAbstract = false)
+        private void GeneratePropertySetter<T>(T decl, Class @class, bool isAbstract = false, Class originalClass = null)
             where T : Declaration, ITypedDecl
         {
             if (!(decl is Function || decl is Field) )
@@ -773,6 +773,7 @@ namespace CppSharp.Generators.CSharp
                 return;
             }
 
+            originalClass = originalClass ?? @class;
             PushBlock(CSharpBlockKind.Method);
             Write("set");
 
@@ -821,12 +822,12 @@ namespace CppSharp.Generators.CSharp
                         else
                         {
                             var parameters = new List<Parameter> { param };
-                            GenerateInternalFunctionCall(function, parameters);
+                            GenerateInternalFunctionCall(function, parameters, null, originalClass);
                         }
                     }
                     else
                     {
-                        GenerateInternalFunctionCall(function, new List<Parameter> { param });
+                        GenerateInternalFunctionCall(function, new List<Parameter> { param }, null, originalClass);
                     }
                 }
                 WriteCloseBraceIndent();
@@ -844,7 +845,7 @@ namespace CppSharp.Generators.CSharp
 
                 var arrayType = field.Type as ArrayType;
 
-                if (arrayType != null && @class.IsValueType)
+                if (arrayType != null && originalClass.IsValueType)
                 {
                     CSharpTypePrinter typePrinter = new CSharpTypePrinter(ctx.Driver);
                     string type = arrayType.CSharpType(typePrinter).Type;
@@ -858,10 +859,10 @@ namespace CppSharp.Generators.CSharp
                 else
                 {
                     ctx.ReturnVarName = string.Format("{0}{1}{2}",
-                    @class.IsValueType
+                    originalClass.IsValueType
                         ? Helpers.InstanceField
                         : string.Format("((Internal*) {0})", Helpers.InstanceIdentifier),
-                    @class.IsValueType ? "." : "->",
+                    originalClass.IsValueType ? "." : "->",
                     Helpers.SafeIdentifier(field.OriginalName));
                 }
                 param.Visit(marshal);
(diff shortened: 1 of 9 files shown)


genuinelucifer:
@ddobrev @tritao
I am confused to see that multiple Inheritance when in "Basic.h" generates incompilable code while when in "CSharpTemp.h" it works perfectly fine.
And "CS_VALUE_TYPE" macro when in Basic.h retains its usability while when in CSharpTemp.h it looses its use!!!
I have no clue why it is so?
Please help!

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -922,9 +922,10 @@ private void GenerateIndexerSetter(Function function)
         }
 
         private void GeneratePropertyGetter<T>(QualifiedType returnType, T decl,
-            Class @class, bool isAbstract = false)
+            Class @class, bool isAbstract = false, Class OrigClass = null)
```
ddobrev: Please use lower case for "OrigClass". You can actually rename it to "originalClass" for more clarity.

tests/Basic/Basic.h
```diff
@@ -762,3 +762,15 @@ struct CS_VALUE_TYPE ValueTypeArrays
 	char thirdValueTypeArray[ARRAY_LENGTH];
 	size_t size;
 };
+
+class DLL_API RefTypeToValTypeCheckBase
+{
+protected:
+	void doSmthProt() { }
```
ddobrev: Please write all function bodies in the cpp file. The issue with inlined bodies is that they are not compiled by default on Linux and OS X and we want the tests to work there as well.

ddobrev:
The multiple inheritance depends on an option which is on for CSharpTemp and off for Basic. So if you see anything like MI in Basic, that's a bug. About C++/CLI, I am afraid @tritao will have to help there, I have little idea.

tests/Basic/Basic.h
genuinelucifer: Oops. I didn't know that. Will remember in future.

genuinelucifer:
@ddobrev @tritao I have fixed the CLI side by forcing class to be ref type if it is inherited from another ref type.
I don't know why the Travis CI build is failing. It is compiling successfully on my machine.
Please review...

ddobrev:
@genuinelucifer don't worry about Travis, it's just not fully set up yet.
@tritao please review the C++/CLI changes.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -915,16 +916,23 @@ private void GenerateIndexerSetter(Function function)
...
-                    Helpers.InstanceIdentifier, function.Parameters[0].Name);
+                Class @class;
+                if(type.TryGetClass(out @class) && @class.IsValueType)
+                    WriteLine("*({0}.Internal*) Internal.{1}({2}, {3}) = ({0}.Internal) value.{2};",
+                        type.ToString(), internalFunction,
+                        Helpers.InstanceIdentifier, function.Parameters[0].Name);
+                else
+                    WriteLine("*({0}.Internal*) Internal.{1}({2}, {3}) = *({0}.Internal*) value.{2};",
+                        type.ToString(), internalFunction,
+                        Helpers.InstanceIdentifier, function.Parameters[0].Name);
```
genuinelucifer: @ddobrev  @tritao Should I use braces here or is it fine like for some of the other one liners?

tritao: Looks fine to me since they're just single statements.

genuinelucifer:
I have completed pushing changes to this pull. Please merge it if it's acceptable. @tritao

src/Generator/Generators/CLI/CLIHeadersTemplate.cs
```diff
@@ -172,6 +172,8 @@ public void GenerateDeclContext(DeclarationContext decl)
                     continue;
 
                 PushBlock(CLIBlockKind.Class, @class);
+                if (@class.HasBaseClass && @class.BaseClass.IsRefType)
```
tritao: We should not be modifying the AST in the text generator.

src/Generator/Generators/CLI/CLITypeReferences.cs
```diff
@@ -191,7 +191,8 @@ public override bool VisitClassDecl(Class @class)
             if (@class.IsIncomplete && @class.CompleteDeclaration != null)
                 @class = (Class) @class.CompleteDeclaration;
 
-            var keywords = @class.IsValueType? "value struct" : "ref class";
+            var keywords = (@class.IsValueType && !(@class.HasBaseClass && @class.BaseClass.IsRefType))
```
tritao: This doesn't look right to me. We should be changing the value of `@class.IsValueType` in a pass instead if the class is not supposed to be a value type.

genuinelucifer:
@tritao Sorry, i tried that earlier but could not find how to do the changes only for CLi and not for C#. Well, I found it now.

src/Generator/Generators/CLI/CLITypeReferences.cs
genuinelucifer: @triao I have one last change for this pull. I will push that tomorrow. If
you see it before then, please don't merge it yet.

src/Generator/Library.cs
```diff
@@ -175,10 +175,15 @@ public static void SetClassBindName(this ASTContext context, string className, s
...
+        public static void SetClassAsValueType(this ASTContext context, string className, Driver driver = null)
         {
             foreach (var @class in context.FindClass(className))
-                @class.Type = ClassType.ValueType;
+            {
+                if (driver != null && driver.Options.IsCLIGenerator && @class.HasBaseClass && @class.BaseClass.IsRefType)
+                    @class.Type = ClassType.RefType;
+                else
+                    @class.Type = ClassType.ValueType;
+            }
```
genuinelucifer: @ddobrev @tritao 
I am not sure about this! Should we leave it upto the user to decide not to make such a class a value type while using CLI generator or is it fine to take the Driver as an argument?
Also, should I leave the default value or shall I change all the files to reflect the changes?

src/Generator/Library.cs
```diff
@@ -175,10 +175,15 @@ public static void SetClassBindName(this ASTContext context, string className, s
                 @class.Name = name;
         }
 
-        public static void SetClassAsValueType(this ASTContext context, string className)
+        public static void SetClassAsValueType(this ASTContext context, string className, Driver driver)
```
genuinelucifer: I changed this to behave similar to CS_VALUE_TYPE macro and not to make value type for specific classes when in CLI mode.
@tritao Please review this part and if it is acceptable now (As the AST is changed only in this function and the macro) please merge it.

ddobrev: Please change this to only pass the information you actually need: Driver.Options.IsCLIGenerator. That is, use a Boolean and not a Driver parameter.

genuinelucifer:
@ddobrev Changed the parameter to bool.

genuinelucifer:
Closing this because of merge conflicts...

Issue #509, opened 2015-06-30
Marshal parameters of type pointers to primitives (int*) as ref params (ref int)


ddobrev:
Should be pretty straightforward: get the addresses of ref params using unsafe code and pass that to the native code. The only corner case is default argument values because C# ref types cannot have them.

genuinelucifer:
@ddobrev Please review #512 Is it acceptable? I merely changed the parameter usage to InOut for the pointers to primitives without a default value!

tritao:
Fixed by https://github.com/mono/CppSharp/commit/dd941d9191dc8d8e97db07828f2804cba72b7237.

PR #512, opened 2015-07-06
Added marshalling of primitive type pointers as ref types.

genuinelucifer:
Deals with #509

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -2581,6 +2581,17 @@ namespace CppSharp.Generators.CSharp
             return paramMarshal;
         }
 
+        private CSharpTypePrinterResult ReturnTypeCheckIsParamRefConvertible(Parameter param)
+        {
+            if(!param.HasDefaultValue && !param.IsOut && !param.IsInOut
+               && param.Type.IsPointerToPrimitiveType() && !param.Type.IsPointerToPrimitiveType(PrimitiveType.Char)
+               && !param.Type.IsPointerToPrimitiveType(PrimitiveType.Void))
+            {
+                param.Usage = ParameterUsage.InOut;
+            }
+            return param.CSharpType(TypePrinter);
+        }
+
         static string GetParameterUsage(ParameterUsage usage)
         {
             switch (usage)
@@ -2599,7 +2610,7 @@ namespace CppSharp.Generators.CSharp
             return string.Join(", ",
                 from param in @params
                 where param.Kind != ParameterKind.IndirectReturnType && !param.Ignore
-                let typeName = param.CSharpType(TypePrinter)
+                let typeName = ReturnTypeCheckIsParamRefConvertible(param)
                 select string.Format("{0}{1} {2}", GetParameterUsage(param.Usage),
                     typeName, param.Name +
                         (param.DefaultArgument == null || !Options.GenerateDefaultValuesForArguments ?
--- a/tests/CSharpTemp/CSharpTemp.h
+++ b/tests/CSharpTemp/CSharpTemp.h
@@ -550,3 +550,7 @@ public:
 	TestParamToInterfacePass(TestParamToInterfacePassBaseTwo b);
 	TestParamToInterfacePass();
 };
+
+DLL_API class SomeRandomClassObject { };
+DLL_API int* pdefnull = 0;
+DLL_API void funcParamPrimitiveRefTypeCheck(int* pOne, char* pTwo, SomeRandomClassObject* pThree, int* pFour = pdefnull) { }


tests/CSharpTemp/CSharpTemp.h
```diff
@@ -550,3 +550,7 @@ class DLL_API TestParamToInterfacePass : public TestParamToInterfacePassBaseOne,
 	TestParamToInterfacePass(TestParamToInterfacePassBaseTwo b);
 	TestParamToInterfacePass();
 };
+
+DLL_API class SomeRandomClassObject { };
+DLL_API int* pdefnull = 0;
+DLL_API void funcParamPrimitiveRefTypeCheck(int* pOne, char* pTwo, SomeRandomClassObject* pThree, int* pFour = pdefnull) { }
```
ddobrev: Please move this function body to the CPP as I told you before.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -2581,6 +2581,17 @@ public struct ParamMarshal
             return paramMarshal;
         }
 
+        private CSharpTypePrinterResult ReturnTypeCheckIsParamRefConvertible(Parameter param)
```
ddobrev: Please move this to a new pass since it changes the AST.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -2581,6 +2581,17 @@ public struct ParamMarshal
...
+        {
+            if(!param.HasDefaultValue && !param.IsOut && !param.IsInOut
```
ddobrev: Please handle default arguments by generating overloads rather than leaving pointers. See the pass for default arguments.

ddobrev:
Please also add tests. One simple to check it works, one that tests passing a returned pointer:

int* p = ReturnsPrimitivePointer();
TakesPrimitivePointer(ref *p);

and one that works with a pointer to an array, for example:

int[] array = {1, 2, 3};
fixed (int* p = array)
{
    TakesPrimitiveArrayPointer(ref *p); // change the array in native code
}
// assert that 'array' contains the new values as modified in C++

tests/CSharpTemp/CSharpTemp.h
genuinelucifer: Sorry. It was empty body so slipped my mind.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
genuinelucifer: Ok. WIll do.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
genuinelucifer: Will do. :+1:

genuinelucifer:
@ddobrev Will add the test.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
genuinelucifer: @ddobrev I tried to copy the HandleDeafultValuesPass to a new pass but adding that to any test project failed the build.
I also tried adding the existing pass to the passes in both Basic.cs and CSharpTemp.cs but both times the build failed without any specific error messages.
Am I missing something? How is it supposed to generate overloads? I mean how to call it to generate overloads for default values?

genuinelucifer:
@tritao @ddobrev I tried to copy the HandleDeafultValuesPass to a new pass but adding that to any test project failed the build.
I also tried adding the existing pass to the passes in both Basic.cs and CSharpTemp.cs but both times the build failed without any specific error messages.
Am I missing something? How is it supposed to generate overloads? I mean how to call it to generate overloads for default values?

ddobrev:
@genuinelucifer you shouldn't make a new pass but rather reuse the existing one. You only need a check if the parameter is ref and fall back to generating overloads in that case.

genuinelucifer:
@tritao @ddobrev I used the following code in basic.h:
```C++
DLL_API class CheckOverloadGen
{
public:
	void FuncCheckDef(int a, int b = 5, char c = 'c', float f = 0.0f);
};
```
and added the existing pass without any edits in basic.cs as:
```C#
public override void SetupPasses(Driver driver)
        {
            driver.AddTranslationUnitPass(new HandleDefaultParamValuesPass());
```
I expected that there will be 4 'FuncCheckDef' generated in the output file basic.cs
But, i found just one!
AM i still missing something?

ddobrev:
@genuinelucifer I will help you with some more explanations:
1. You don't need to add the pass, it's already been added. There's an option for it which is false by default. Search for its usages;
2. The tests need to be added to Basic.h, however, this particular test about a default value of a ref param needs to be added to CSharpTemp. The reason is that Basic.h is also used for testing our C++/CLI back-end and C++/CLI does not support default values of parameters;
3. The pass uses default values of parameters whenever C# supports the particular defaults. Overloads are only generated as a fall-back. Since your example only uses simple primitives, you correctly get just one function.

HandleDefaultValuesPass has various checks whether a default value for a parameter is supported. What you need to do is add one more check about whether the parameter is ref.

genuinelucifer:
@ddobrev Thank you so much for the explaination. It really helped a lot. I finally got it working. Only a few bits left. :)

PR #513, opened 2015-07-07
Fixed marshalling base members from ref type to value type.

genuinelucifer:
If a value type class inherited from a ref type base then there were errors due to scope of methods and   properties in C# and it was not possible in C++/CLI. Fixed these.


genuinelucifer:
@ddobrev @tritao Please review and merge it!

tritao:
Closing as https://github.com/mono/CppSharp/pull/518 is a superset of this.

Issue #500, opened 2015-06-25
Wrap C++ structs as C# structs where appropriate


ddobrev:
Structs and classes are essentially the same in C++. However, the way they are used is almost identical as they way they are in C#. So it would be a good idea if a C++ struct is mapped to a C# struct. That is not always possible - for example, the struct must not be inherited from because that's forbidden in C#. So the algorithm should:
1. Only consider C++ structs, that is, types declared with the "struct" keyword;
2. Apply a series of checks, for example searching the AST for derived types as stated above, to determine if it's possible to represent it as a C# struct.

Note: manual marking of types as structs must be left to give freedom to binders.

genuinelucifer:
@ddobrev Please review #501

genuinelucifer:
@ddobrev Please review #518

genuinelucifer:
@ddobrev 
There is a corner case for this which can't be checked directly. As in NamespaceDerived and NamespaceBase Test in which both are compiled separately but the Derived one inherits from the Base.
We have practically no way of knowing that a future type (rather compiled separately) will inherit this or not. So, if we marshal NamespaceBase to struct then build will fail for NamespaceDerived.
I could think that if the user enables MarshalStructToStructWhereverPossible() Pass then he must also mark the C++ struct(s) with some macro to prohibit value type marshalling. (ex: STRICT_REF_TYPE)
What do you suggest as a workaround for this?

genuinelucifer:
@ddobrev Any comment?!

tomspilman:
Getting this working would be a huge benefit to many APIs cases we've run into.  Currently we end up writing a dummy class to wrap the struct with getters/setters to access the fields.  It is a lot of wasted effort when a C++ struct can be easily be represented in C#.

PR #518, opened 2015-07-12
Added a pass to allow struct to struct marshalling if possible

genuinelucifer:
@tritao @ddobrev Please review. It is based on one of my earlier pulls and the first commit is from that only.
This pull tries to address #500 .

### DIFF
--- a/src/Generator/Passes/MarshalStructAsValTypeIfPossiblePass.cs
+++ b/src/Generator/Passes/MarshalStructAsValTypeIfPossiblePass.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using CppSharp.AST;
+
+namespace CppSharp.Passes
+{
+    public class MarshalStructAsValTypeIfPossiblePass : TranslationUnitPass
+    {
+        
+        public override bool VisitClassDecl(Class @class)
+        {
+            @class.Type = CheckClassIsStructible(@class, Driver) ? ClassType.ValueType : @class.Type;
+            return base.VisitClassDecl(@class);
+        }
+
+        private bool CheckClassIsStructible(Class @class, Driver Driver)
+        {
+            if (@class.IsValueType)
+                return true;
+
+            if (!@class.IsStruct)
+                return false;
+            if (@class.IsInterface || @class.IsStatic || @class.IsAbstract)
+                return false;
+            if (@class.Declarations.Any(decl => decl.Access == AccessSpecifier.Protected))
+                return false;
+            if (@class.Methods.Any(m => m.IsVirtual))
+                return false;
+            if (Driver.Options.IsCLIGenerator && @class.HasBaseClass && @class.BaseClass.IsRefType)
+                return false;
+
+            var allTrUnits = Driver.ASTContext.TranslationUnits;
+            foreach (var trUnit in allTrUnits)
+            {
+                foreach (var cls in trUnit.Classes)
+                {
+                    if (cls.Bases.Any(clss => clss.IsClass && clss.Class == @class))
+                        return false;
+                }
+            }
+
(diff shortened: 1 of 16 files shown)


genuinelucifer:
As far as I have checked it, now it has to have a check that if there is an explicitly defined constructor/destructor [like in 'TestDestructor' in CSharpTemp.h] then too it must NOT be marshalled as a struct.

src/Generator/Passes/MarshalStructAsValTypeIfPossiblePass.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using CppSharp.AST;
+
+namespace CppSharp.Passes
+{
+    public class MarshalStructAsValTypeIfPossiblePass : TranslationUnitPass
+    {
+
```
tritao: Extra newline here.

src/Generator/Passes/MarshalStructAsValTypeIfPossiblePass.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using CppSharp.AST;
+
+namespace CppSharp.Passes
+{
+    public class MarshalStructAsValTypeIfPossiblePass : TranslationUnitPass
```
tritao: I think we should be more consistent here and rename this to something like our existing passes, maybe `CheckValueTypeClassesPass` or something.

src/Generator/Passes/MarshalStructAsValTypeIfPossiblePass.cs
```diff
@@ -0,0 +1,46 @@
...
+            @class.Type = CheckClassIsStructible(@class, Driver) ? ClassType.ValueType : @class.Type;
+            return base.VisitClassDecl(@class);
+        }
+
+        private bool CheckClassIsStructible(Class @class, Driver Driver)
+        {
+            if (@class.IsValueType)
+                return true;
+
+            if (!@class.IsStruct)
```
tritao: Me and @ddobrev discussed this and we think it's better to replace all this logic with `@class.IsPOD`.

genuinelucifer:
if we have code like:
```C++
struct RefType
{
protected:
    int n;
public:
    virtual int getN() { return n; }
};

struct CS_VALUE_TYPE ValType : public RefType
{
};
```
Then uncompilable code was generated. We simply copy members from the base into the derived type if the derived type is a value type. But value types can't have protected &/or virtual members. So, i just wanted to fix that.

ddobrev:
@genuinelucifer your answer shows that you've ignored part of @tritao 's questions. He's asked you if the code is STILL necessary when you consider our new discovery - value types cannot derive from ref types in C# either.

src/Generator/Passes/MarshalStructAsValTypeIfPossiblePass.cs
genuinelucifer: Will change it.

src/Generator/Passes/MarshalStructAsValTypeIfPossiblePass.cs
genuinelucifer: will do.

src/Generator/Passes/MarshalStructAsValTypeIfPossiblePass.cs
genuinelucifer: will remove.

genuinelucifer:
@ddobrev @tritao Sorry. I was leaving for some place and wrote that in a hurry.
Since it is not supported so i don't think there is any need of keeping it. I will remove it.

PR #519, opened 2015-07-14
Clang bindings

ocdmessiah:


tritao:
This pull is failing the AppVeyor build, which means there are some regressions in your refactoring. Please fix them so we can merge it.

tritao:
ping

src/AST/ASTVisitor.cs
```diff
@@ -63,6 +63,12 @@ public bool AlreadyVisited(Declaration decl)
 
         #region Type Visitors
 
+        public virtual void VisitAll<D>(IEnumerable<D> decs) where D:Declaration
+        {
+            foreach (var d in decs)
```
tritao: This should be `decls.`

src/AST/IMember.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace CppSharp.AST
+{
+    public interface IMember<M> where M : Declaration, IMember<M>
```
tritao: I'd prefer for M to be named `D` or `Decl` since it's more readable..

src/AST/IMember.cs
```diff
@@ -0,0 +1,10 @@
...
+    {
+        String Name { get; }
+        bool canOverride(M m);
```
tritao: This method does not follow naming conventions (Uppercase) and I'd prefer for the parameter to have a meaningful name, like `member` or `decl`.

src/AST/Property.cs
```diff
@@ -28,40 +30,50 @@ public Type Type
 
         public QualifiedType QualifiedType { get; set; }
 
+
+        public bool MemberVerifies<M>(M m, Predicate<M> pred) where M:Declaration, IMember<M>
```
tritao: IMHO these should be moved out of the declaration and into extension methods.

src/Generator/Passes/CheckAbiParameters.cs
```diff
@@ -26,6 +26,9 @@ public class CheckAbiParameters : TranslationUnitPass
     {
         public override bool VisitClassDecl(Class @class)
         {
+            if(@class.Name.Contains("build_index_impl"))
+                System.Console.WriteLine("CLASS\t" + @class.Name);
```
ddobrev: Remove this anywhere you've forgotten to, it's test code.

src/CppParser/Bootstrap/Bootstrap.cs
```diff
@@ -62,13 +62,13 @@ public void Preprocess(Driver driver, ASTContext ctx)
...
+            var exprClass = ctx.FindCompleteClass("clang::Expr");
 
             var exprUnit = ctx.TranslationUnits [0];
-            var subclassVisitor = new SubclassVisitor (exprClass);
-            exprUnit.Visit (subclassVisitor);
+            var subclassVisitor = new SubclassVisitor(exprClass);
+            exprUnit.Visit(subclassVisitor);
 
-            var subclasses = subclassVisitor.Classes;
+            //var subclasses = subclassVisitor.Classes;
```
ddobrev: Do not comment out redundant lines, just delete them.

src/CppParser/Bootstrap/Bootstrap.cs
```diff
@@ -42,15 +42,15 @@ public void Setup(Driver driver)
             options.MicrosoftMode = false;
             options.TargetTriple = "i686-apple-darwin12.4.0";
 
-            options.addDefines ("__STDC_LIMIT_MACROS");
-            options.addDefines ("__STDC_CONSTANT_MACROS");
+            options.addDefines("__STDC_LIMIT_MACROS");
+            options.addDefines("__STDC_CONSTANT_MACROS");
 
-            var llvmPath = Path.Combine (GetSourceDirectory ("deps"), "llvm");
+            var llvmPath  = Path.Combine(GetSourceDirectory("deps"), "llvm");
```
ddobrev: There's an extra space before the assignment.

tritao: We usually do not align stuff like this.

src/AST/TypeExtensions.cs
```diff
@@ -207,13 +207,7 @@ public static Type SkipPointerRefs(this Type t)
...
         public static Type GetPointee(this Type t)
         {
-            var ptr = t as PointerType;
-            if (ptr != null)
-                return ptr.Pointee;
-            var memberPtr = t as MemberPointerType;
-            if (memberPtr != null)
-                return memberPtr.QualifiedPointee.Type;
-            return null;
+            return (t as PointerType)?.Pointee ?? (t as MemberPointerType)?.QualifiedPointee.Type;
```
tritao: We don't support C# 6. And IMHO this one-liner is less readable than what we had before.

ddobrev: We haven't switched to C# 6 yet so do not use ?. I think you'd best revert this whole piece.

src/AST/Property.cs
```diff
@@ -125,5 +129,10 @@ public override T Visit<T>(IDeclVisitor<T> visitor)
             return visitor.VisitProperty(this);
         }
 
+        public bool canOverride(Property p)
```
ddobrev: Rename to CanOverride to keep the standard Mono/.NET naming convention.

src/AST/SymbolContext.cs
```diff
@@ -97,37 +97,21 @@ public bool FindSymbol(ref string symbol)
...
-            alternativeSymbol = symbol.TrimStart('_');
-            if (FindLibraryBySymbol(alternativeSymbol, out lib))
-            {
-                symbol = alternativeSymbol;
-                return true;
-            }
-
-            alternativeSymbol = "_imp_" + symbol;
-            if (FindLibraryBySymbol(alternativeSymbol, out lib))
+            var alternativeSymbols = new []
```
tritao: While this makes for nicer code, it will do a bunch of needless string allocations. I think we can easily fix by building a regex to do the matching. See https://github.com/mono/CppSharp/blob/master/src%2FGenerator%2FLibrary.cs#L129.

src/AST/ClassExtensions.cs
```diff
@@ -54,99 +54,69 @@ public static Class GetNonIgnoredRootBase(this Class @class)
...
-                b => b.Type.IsClass() && b.Class.IsDeclared))
-            {
-                property = baseClassSpecifier.Class.GetPropertyByName(propertyName);
-                if (property != null)
-                    return property;
-            }
-            return null;
+        public static M GetMemberByName<M>(this Class c, string memberName, Func<Class, IEnumerable<M>> members, Func<BaseClassSpecifier, bool> considerBase) where M : Declaration, IMember<M>
+        {
+            var member = members(c).FirstOrDefault(m => m.Name == memberName);
```
ddobrev: You have changed the behaviour. Mine was depth-first, that is, the most basic member was returned. You need to flip the two parts of your new code: try the recursive call first and then the members.

afesteves: Is this about GetPropertyByName? It seems to me that function (both before and after changes) will not even recurse if it finds a property on the subclass.
https://github.com/mono/CppSharp/blob/35a5f7d4b9dee2c44b4d239e980bd0ef044a12fb/src/AST/ClassExtensions.cs#L93

ddobrev: Sorry, my mistake. I was talking about GetRoot... but I got confused. I can see GetRoot... is actually OK so please ignore my above comment.

PR #520, opened 2015-07-19
Fixed namespace conflict in C++/CLI.

genuinelucifer:
Just was going through the codes and found the TODO in UTF16 Test....
Added CLI generation for the test and one of the problems that i encountered was same as #365 
.
This pull solves that. But, it is totally a naming convention thing. So please see if this would be carried out or not. @tritao


tritao:
Well, like you said this is a naming convention thing. I don't think we want to add "WrapperClass" to the class that is generated automatically.

Me and @ddobrev fixed the original in a different way in https://github.com/mono/CppSharp/commit/cf472f526c87e9807a9f9868e7cbc07376efcbb6.

PR #522, opened 2015-07-23
closes #514.

### LINKED ISSUE #514: Passing null to a function pointer causes a crash
We don't check what we pass to Marshal.GetFunctionPointerForDelegate. To reproduce, set CSharpTemp.Foo.AttributedFunctionPtr to null.

xistoso:
closes #514.
test added to tests/CSharpTemp/CSharpTemp.Tests.cs

### DIFF
--- a/tests/CSharpTemp/CSharpTemp.Tests.cs
+++ b/tests/CSharpTemp/CSharpTemp.Tests.cs
@@ -327,4 +327,14 @@ public class CSharpTempTests : GeneratorTestFixture
         dervClass = new TestParamToInterfacePass(dervClass + baseInterface);
         Assert.AreEqual(dervClass.M, 2);
     }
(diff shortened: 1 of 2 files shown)


tests/CSharpTemp/CSharpTemp.Tests.cs
```diff
@@ -327,4 +327,13 @@ public void TestParamTypeToInterfacePass()
         dervClass = new TestParamToInterfacePass(dervClass + baseInterface);
         Assert.AreEqual(dervClass.M, 2);
     }
+
+    [Test]
+    public void TestNullAttributedFunctionPtr()
+    {
+        var foo = new Foo();
```
tritao: Use `using` here to properly dispose of the instance.

PR #523, opened 2015-07-26
Added primitive type to ref type pass and improved support for multi overloads…

genuinelucifer:
The test is failing. One of the reasons may be that when we pass ref int to C# code it might not allow some thing like:
```C#
public void TakePrimTypePtr(ref int ptr)
{
int _ptr = ptr;
var arg0 = &_ptr;
Internal.TakePrimTypePtr_0(__Instance, arg0);
ptr = _ptr;
}
```
The arg0 might not be pointing to the same memory as ptr and hence the mismatch.
But again, I am just begineer in C# so I might be wrong.
@ddobrev @tritao Insights?

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -2035,6 +2035,13 @@ namespace CppSharp.Generators.CSharp
             if (method.IsPure)
                 Write("abstract ");
 
+            if(method.SynthKind == FunctionSynthKind.DefaultValueOverload)
+            {
+                var needUnsafe = method.Parameters.Any(p => p.Kind == ParameterKind.Regular && p.Ignore
+                             && p.Type.IsPointerToPrimitiveType() && p.Usage == ParameterUsage.InOut && p.HasDefaultValue);
+                Write("unsafe ");
+            }
+
             var functionName = GetMethodIdentifier(method);
 
             if (method.IsConstructor || method.IsDestructor)
@@ -2135,24 +2142,23 @@ namespace CppSharp.Generators.CSharp
             PopBlock(NewLineKind.BeforeNextBlock);
         }
 
+        private string OverloadParamNameWithDefValue(Parameter p, ref int j)
+        {
+            return (p.Type.IsPointerToPrimitiveType() && p.Usage == ParameterUsage.InOut && p.HasDefaultValue)
+                   ? "ref param" + j++ : p.DefaultArgument.String;
+        }
+
         private void GenerateOverloadCall(Function function)
         {
-            var needUnsafe = function.Parameters.Any(p => p.Kind == ParameterKind.Regular && p.Ignore
-                             && p.Type.IsPointerToPrimitiveType() && p.Usage == ParameterUsage.InOut && p.HasDefaultValue);
-            if(needUnsafe)
+            int i = 0;
+            foreach(Parameter param in function.Parameters.Where(
+                        p => p.Kind == ParameterKind.Regular && p.Ignore && p.Type.IsPointerToPrimitiveType()
+                             && p.Usage == ParameterUsage.InOut && p.HasDefaultValue))
             {
-                WriteLine("unsafe");
-                WriteStartBraceIndent();
-                int i = 0;
-                foreach(Parameter param in function.Parameters.Where(
-                            p => p.Kind == ParameterKind.Regular &&
-                                 p.Ignore && p.Type.IsPointerToPrimitiveType() && p.Usage == ParameterUsage.InOut && p.HasDefaultValue))
-                {
-                    var strtype = param.Type.CSharpType(TypePrinter).ToString();
-                    WriteLine("{0} arg{1} = ({0}){2};", strtype, i, param.DefaultArgument.String);
-                    WriteLine("{0} param{1} = *arg{1};", strtype.Substring(0, strtype.Length - 1), i);
-                    ++i;
-                }
+                var strtype = param.Type.CSharpType(TypePrinter).ToString();
+                WriteLine("{0} arg{1} = ({0}){2};", strtype, i, param.DefaultArgument.String);
+                WriteLine("{0} param{1} = *arg{1};", strtype.Substring(0, strtype.Length - 1), i);
+                ++i;
             }
 
             int j = 0;
@@ -2163,10 +2169,7 @@ namespace CppSharp.Generators.CSharp
                 string.Join(", ",
                     function.Parameters.Where(
                         p => p.Kind == ParameterKind.Regular).Select(
-                        p => p.Ignore ? ((p.Type.IsPointerToPrimitiveType() && p.Usage == ParameterUsage.InOut && p.HasDefaultValue) ? "ref param" + j++ : p.DefaultArgument.String)
-                                        : (p.Usage == ParameterUsage.InOut ? "ref " : string.Empty) + p.Name)));
-            if(needUnsafe)
-                WriteCloseBraceIndent();
+                        p => p.Ignore ? OverloadParamNameWithDefValue(p, ref j) : (p.Usage == ParameterUsage.InOut ? "ref " : string.Empty) + p.Name)));
         }
 
         private void GenerateEquals(Function method, Class @class)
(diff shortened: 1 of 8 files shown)


tests/CSharpTemp/CSharpTemp.Tests.cs
```diff
@@ -340,4 +340,35 @@ public void TestNullAttributedFunctionPtr()
...
+                obj.TakePrimTypePtr(ref *p1);
+                Assert.AreEqual(100, p1[0]);          //Pass!
+                //Assert.AreEqual(200, p1[1]);          //Fail!
+                //Assert.AreEqual(300, p1[2]);          //Fail!
+            }
+
+            Assert.AreEqual(100, array[0]);           //Pass!!!
+            //Assert.AreEqual(200, array[1]);           //FAIL!!!
+            //Assert.AreEqual(300, array[2]);           //FAIL!!!
+        }
```
genuinelucifer: The tests are passing for 1st element while they are failing for the rest. Might be int* and ref int are that exchangeable after all?!
@ddobrev @tritao Any mistakes on my side? Or any editions?!

ddobrev: int _ptr = ptr;

This makes a copy so you pass to C++ the address of _ptr and not of ptr. Could you change that to use ptr directly:
{
    var arg0 = &ptr;
    Internal.TakePrimTypePtr_0(__Instance, arg0);
}

There's a chance this will also fix your tests. As a beginning you could try changing the code manually (you can disable the pre-build event which regenerates the code) and see if it works.

genuinelucifer: @ddobrev I changed the generated function to
```C#
public void TakePrimTypePtr(ref int ptr)
{
    fixed (int* arr = &ptr)
    {
        Internal.TakePrimTypePtr_0(__Instance, arr);
    }
}
```
And now all the tests pass!!! 
Should I change the code generation for all ref type parameters?
Moreover, what in case of more than 1 ref type params to be passed?
Can we use nested fixed statements??

genuinelucifer: Simply using
```C#
var arg = &ptr;
```
generated error
```
You can only take the address of an unfixed expression inside of a fixed statement initializer
```

ddobrev: You can safely nest fixed statements, it's fine.

genuinelucifer:
@tritao @ddobrev  Please review again. I have checked it and now it seems to me that everything is working. (No errors and all tests passed.)

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -2137,6 +2137,25 @@ public void GenerateMethod(Method method, Class @class)
 
         private void GenerateOverloadCall(Function function)
         {
+            var needUnsafe = function.Parameters.Any(p => p.Kind == ParameterKind.Regular && p.Ignore
+                             && p.Type.IsPointerToPrimitiveType() && p.Usage == ParameterUsage.InOut && p.HasDefaultValue);
+            if(needUnsafe)
+            {
+                WriteLine("unsafe");
```
ddobrev: This does work but please move it to the method signature instead, it's less cumbersome.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -2590,16 +2621,26 @@ public struct ParamMarshal
 
             paramMarshal.Context = ctx;
 
-            var marshal = new CSharpMarshalManagedToNativePrinter(ctx);
-            param.CSharpMarshalToNative(marshal);
+            if (param.IsInOut && param.Type.IsPointerToPrimitiveType()
+                && ExtensionMethods.AllowedToHaveDefaultPtrVals.Any(primType => param.Type.IsPointerToPrimitiveType(primType)))
```
ddobrev: Since you always use this list the same way, why don't you just create a function that takes a primitive type and returns a Boolean?

tests/CSharpTemp/CSharpTemp.Tests.cs
```diff
@@ -340,4 +340,35 @@ public void TestNullAttributedFunctionPtr()
         }
     }
 
+    [Test]
+    public void TestMultiOverLoadNPtrToRefTypeGen()
+    {
+        unsafe
```
ddobrev: Please move this to the signature instead, less nesting of code.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -2144,7 +2163,10 @@ private void GenerateOverloadCall(Function function)
                 string.Join(", ",
                     function.Parameters.Where(
                         p => p.Kind == ParameterKind.Regular).Select(
-                            p => p.Ignore ? p.DefaultArgument.String : p.Name)));
+                        p => p.Ignore ? ((p.Type.IsPointerToPrimitiveType() && p.Usage == ParameterUsage.InOut && p.HasDefaultValue) ? "ref param" + j++ : p.DefaultArgument.String)
```
ddobrev: This has got too long and complicated. Please extract it to a separate function.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -2487,6 +2509,10 @@ into context
 
             if (needsFixedThis && operatorParam == null)
                 WriteCloseBraceIndent();
+
+            var numFixedBlocks = @params.Count(param => param.HasFixedBlock);
+            for(var i=0; i<numFixedBlocks; ++i)
```
ddobrev: You missed the spaces around the signs.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
genuinelucifer: Will move to the method signature.
Well, it is required because if there is a ref param with default value then it will make a native pointer with that value and then take its address to a managed type and pass the ref to managed type to the function call.

genuinelucifer:
@ddobrev Changes done.
Pls review again.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -2035,6 +2035,13 @@ public void GenerateMethod(Method method, Class @class)
             if (method.IsPure)
                 Write("abstract ");
 
+            if(method.SynthKind == FunctionSynthKind.DefaultValueOverload)
+            {
+                var needUnsafe = method.Parameters.Any(p => p.Kind == ParameterKind.Regular && p.Ignore
+                             && p.Type.IsPointerToPrimitiveType() && p.Usage == ParameterUsage.InOut && p.HasDefaultValue);
+                Write("unsafe ");
```
ddobrev: I read and understood your explanation. However, I removed all "unsafe" from your generated code and it still compiled and worked properly. If you really need this, please add a test that exposes the need. Otherwise please remove it.
This is the last major problem with this pull. Once you're done with it, we'll merge even if you decide to ignore the other minor remarks.

src/Generator/Generators/ExtensionMethods.cs
```diff
@@ -1,10 +1,17 @@
...
+using System.Collections.Generic;
+using CppSharp.AST;
+using CppSharp.AST.Extensions;
 using Interop = System.Runtime.InteropServices;
 
 namespace CppSharp.Generators
 {
     public static class ExtensionMethods
     {
+        public static List<PrimitiveType> AllowedToHaveDefaultPtrVals = new List<PrimitiveType> {PrimitiveType.Bool, PrimitiveType.Double, PrimitiveType.Float,
```
ddobrev: Now that you have a dedicated function, you could make this private or better yet, make it local for the function.

genuinelucifer:
@ddobrev Sorry. I am still trying out C#. I didn't see that if the class is marked 'unsafe' then we don't need another unsafe in the function. Removed it now.
Should i open a new PR for this now? Because github is showing merge conflicts and i don't know if they can be resolved without a new PR.

ddobrev:
Thank you, please do file a new request. As usual, please merge the sub-commits (cleaning up, subsequent small fixes) into the main one.

PR #525, opened 2015-07-26
Added marshalling to struct...

genuinelucifer:
Mostly done... But, one other error is exposed due to this. And moreover it now also marshals few C++ classes to C# struct.
Was that intended while using ``` @class.IsPOD ``` @tritao @ddobrev 

Also, I did not add the check for is declared as struct this time.

### DIFF
--- a/src/Generator/Passes/CheckValueTypeClassesPass.cs
+++ b/src/Generator/Passes/CheckValueTypeClassesPass.cs
@@ -1,0 +1,42 @@
+﻿using System.Linq;
+using CppSharp.AST;
+
+namespace CppSharp.Passes
+{
+    public class CheckValueTypeClassesPass : TranslationUnitPass
+    {
+        public CheckValueTypeClassesPass()
+        {
+        }
+
+        public override bool VisitClassDecl(Class @class)
+        {
+            @class.Type = CheckClassIsStructible(@class, Driver) ? ClassType.ValueType : @class.Type;
+            return base.VisitClassDecl(@class);
+        }
+
+        private bool CheckClassIsStructible(Class @class, Driver Driver)
+        {
+            if (@class.IsUnion || @class.Namespace.Templates.Any(tmp => tmp.Name.Equals(@class.Name)))
+                return false;
+            if (@class.IsInterface || @class.IsStatic || @class.IsAbstract)
+                return false;
+            if (@class.Declarations.Any(decl => decl.Access == AccessSpecifier.Protected))
+                return false;
+            if (@class.Methods.Any(m => m.IsVirtual))
+                return false;
+            if (@class.HasBaseClass && @class.BaseClass.IsRefType)
+                return false;
+
+            var allTrUnits = Driver.ASTContext.TranslationUnits;
+            if (allTrUnits.Any(trUnit => trUnit.Classes.Any(
+                                  cls => cls.Bases.Any(clss => clss.IsClass && clss.Class == @class))))
+                return false;
+
+            if (@class.IsPOD || @class.IsValueType)
+                return true;
+
+            return false;
+        }
+    }
+}
--- a/tests/Basic/Basic.cs
+++ b/tests/Basic/Basic.cs
@@ -32,6 +32,7 @@ namespace CppSharp.Tests
         {
             driver.AddTranslationUnitPass(new GetterSetterToPropertyPass());
             driver.AddTranslationUnitPass(new CheckMacroPass());
+            driver.AddTranslationUnitPass(new CheckValueTypeClassesPass());
             ctx.SetClassAsValueType("Bar");
             ctx.SetClassAsValueType("Bar2");
             ctx.IgnoreClassWithName("IgnoredType");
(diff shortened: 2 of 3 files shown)


src/Generator/Passes/CheckValueTypeClassesPass.cs
```diff
@@ -0,0 +1,42 @@
...
+                return false;
+            if (@class.HasBaseClass && @class.BaseClass.IsRefType)
+                return false;
+
+            var allTrUnits = Driver.ASTContext.TranslationUnits;
+            if (allTrUnits.Any(trUnit => trUnit.Classes.Any(
+                                  cls => cls.Bases.Any(clss => clss.IsClass && clss.Class == @class))))
+                return false;
+
+            if (@class.IsPOD || @class.IsValueType)
```
ddobrev: What @tritao had in mind was to use IsPOD instead of IsValueType. So you'd best change it to just:

return @class.IsPOD.

src/Generator/Passes/CheckValueTypeClassesPass.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        private bool CheckClassIsStructible(Class @class, Driver Driver)
+        {
+            if (@class.IsUnion || @class.Namespace.Templates.Any(tmp => tmp.Name.Equals(@class.Name)))
+                return false;
+            if (@class.IsInterface || @class.IsStatic || @class.IsAbstract)
+                return false;
+            if (@class.Declarations.Any(decl => decl.Access == AccessSpecifier.Protected))
+                return false;
+            if (@class.Methods.Any(m => m.IsVirtual))
```
ddobrev: That's OK but just so you know, there's a short cut: @class.IsDynamic.

tests/Basic/Basic.cs
```diff
@@ -32,6 +32,7 @@ public override void Preprocess(Driver driver, ASTContext ctx)
         {
             driver.AddTranslationUnitPass(new GetterSetterToPropertyPass());
             driver.AddTranslationUnitPass(new CheckMacroPass());
+            driver.AddTranslationUnitPass(new CheckValueTypeClassesPass());
```
ddobrev: This is supposed to be added to Driver, not just to the test.

ddobrev:
Marshalling some classes as structs is OK, this is what @tritao 's idea is: marshal types based on their usage, not on the key-word which means almost nothing in C++.

src/Generator/Passes/CheckValueTypeClassesPass.cs
genuinelucifer: Will do.

src/Generator/Passes/CheckValueTypeClassesPass.cs
genuinelucifer: Ohh. great! Thanks.

tests/Basic/Basic.cs
genuinelucifer: I will have to see how to do that. WIll try.

genuinelucifer:
@ddobrev The code marshals a class which was intended to be a Ref type as I had written it for testing a fix when CS_OUT was applied to a ref type argument.
Now it is value type and it is not being properly initialising before use.

genuinelucifer:
@ddobrev Also i wanted to ask about the case when a class in completely different assembly inherits from a class which was previously marshalled as Value Type.
I described the problem on #500 a few days back,

genuinelucifer:
@ddobrev Adding this pass to generator didn't seem to be good. Because of two main reasons:
```1.``` If user has 1 class in one project inheriting from a second class in another project. Then if he uses CppSharp on the two projects separately (As in Namespace Test) then it causes error if the Base has laready been marshalled to value type.
```2.``` Some classes are used in Tests like they are ref type and after marshalling them to value type the tests break.
.
For (1) I have introduced a new macro ```CS_REF_TYPE``` which will allow the user to force marshalling to ref type even when CheckClassAsValueTypePass is used just on the condition that the CheckMacroPass is used before this pass.

genuinelucifer:
@tritao @ddobrev Please review this.

genuinelucifer:
@ddobrev @tritao Any comments?!

ddobrev:
Hello @genuinelucifer . Please accept my apologies for this huge delay. I've been really busy with Qt#, I am almost ready to build GUI examples with it and it needs a last effort. Meanwhile please proceed with the task about delegates, it's more important than this one anyway.

genuinelucifer:
@ddobrev That's great! :) I had a look into Qt# and it really seems that you have put much much effort into it. Best of Luck! 

I'm actually unable to proceed with the delegates task because I actually don't know much about them and actually couldn't figure out how to do that. I will post my real problems on that issue soon.

ddobrev:
@genuinelucifer I am glad you have started thinking about the task. Feel free to ask any questions. I still don't have time to review this one but I have more than enough for questions.

tritao:
Hey @ddobrev we should review this soon.

ddobrev:
(sigh) I agree. It's just that the problems keep popping up...

PR #530, opened 2015-08-01
Added marshalling of ptrs to ref type.

genuinelucifer:


src/Generator/Generators/ExtensionMethods.cs
```diff
@@ -23,5 +26,14 @@ public static Interop.CallingConvention ToInteropCallConv(this CallingConvention
 
             return Interop.CallingConvention.Winapi;
         }
+
+        public static bool IsParamPrimToRefTypeConvertible(Parameter param, bool checkParamUsage = true)
+        {
+            List<PrimitiveType> AllowedToHaveDefaultPtrVals = new List<PrimitiveType> {PrimitiveType.Bool, PrimitiveType.Double, PrimitiveType.Float,
```
ddobrev: @tritao said he wanted to review your code too a little later. In the mean time - since you have to wait anyway - could you please change this to lower-case ('allowedToHaveDefaultPtrVals ') as per our convention for local variables.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -2573,11 +2598,11 @@ public struct ParamMarshal
         {
             if (param.Type is BuiltinType)
             {
-                return new ParamMarshal { Name = param.Name, Param = param };
+                return new ParamMarshal { Name = param.Name, Param = param, HasFixedBlock = false };
```
ddobrev: @tritao said he wanted to review your code too a little later. In the mean time - since you have to wait anyway - you might skip setting HasFixedBlock to 'false' because Boolean fields are false by default.

ddobrev:
I added some really minor remarks while we wait for @tritao 's review. If you decide to make the changes, you don't have to file separate commits, then merge them, then file a new pull request. Instead, just amend your commit and then force-push.

genuinelucifer:
I couldn't find a clear way to ammend and force push commits from command line. (I prefer cmd.)
So i pushed changes and rebased instead.

ddobrev:
Thank you very much for putting up with us and mostly for your excellent work on this. I wanted to add it for more than a year and a half.

ddobrev:
Just a note for you to know. This code you generated for overloads:

            int* arg0 = (int*)0;
            int param0 = *arg0;

actually causes a crash at the second line - you try dereferencing an empty pointer. I fixed that for you and tested it by invoking your own test methods.

genuinelucifer:
Oh. Thanks :+1:

ddobrev:
You're welcome. Could you please read https://github.com/mono/CppSharp/commit/dd941d9191dc8d8e97db07828f2804cba72b7237#commitcomment-12514435 and deal with it?

Issue #498, opened 2015-06-24
Generate delegates in a separate file and reuse them instead of generating them each time


ddobrev:
At the moment when we need a delegate - v-tables, abstract impls, etc, we generate it in the same file. This leads to repetition, bloated API-s and binaries, etc. So we need to generate such delegates in a location independent of the bound API, and reuse them if possible. That is, maintain a dictionary of those delegates, and whenever we need one for a function, check by its signature - if a match is found, use it, if not, add it. This dictionary needs to be static so that it lasts for the lifetime of the generator app. This would allow it to be reused in dependent modules. When generating code at the end, the list of these delegates has to be generated too, best in a separate file.

We could use a convention similar to .NET's:

for functions returning a value: Func_<return_type>_<parameter_type1>_<parameter_type2>_...
for functions not returning a value: Action_<parameter_type1>_<parameter_type2>_...

The reason we cannot use the real Func<> and Action<> is that .NET 4.0, which we want to support because of Windows XP, cannot marshal generic delegates. On the other hand, Mono for Windows dropped XP support not long ago. So we need to stick to generated delegates.

genuinelucifer:
It might be silly. But, I tried a lot to see how to generate a new file using the already present classes and methods but could not do so.
Please help @ddobrev @tritao

genuinelucifer:
@ddobrev I am not certain but I read somewhere online that generic delegates were introduced in .NET 3.5 so i guess .NET 4.0 would be able to Marshal them. Isn't it the case?

ddobrev:
I am afraid that regardless of the time of their introduction, they cannot be marshalled. I myself have wondered how they could add this feature in half but they did.
If you have trouble with a new file, we can start without one. You could, for example, collect all delegates you're going to need and print them at the end of the last code file.

genuinelucifer:
@ddobrev Thanks. Then I will try to generate all the delegates at one place
first...

genuinelucifer:
@ddobrev I am really quite confused on this task. I read much about delegates but that didn't help.
Do we wish to move all the instantiations of the delegates to a separate file/class and call them from each class when needed? If so, how does it help? I know I am missing something here but I actually am unable to get what to do here.

ddobrev:
Hello @genuinelucifer , glad to hear back from you. :)
You have understood it correctly - we want all delegates moved to a separate file and reused as needed. Just in case, here's an example.
What we now do is generate delegates:

protected delegate void _delegate_sampleFunction();

[...]

PR #532, opened 2015-08-02
Marshalling char* to Stringbuilder.

genuinelucifer:
@ddobrev @tritao It is my initial attempt. I am completely new to encoding of characters. Also i have delved into the world of code 'marshalling' only with this project. So, I just wanted to ask if I am going in right direction with this.

### DIFF
--- a/tests/CSharpTemp/CSharpTemp.Tests.cs
+++ b/tests/CSharpTemp/CSharpTemp.Tests.cs
@@ -349,25 +349,26 @@ public class CSharpTempTests : GeneratorTestFixture
     public unsafe void TestMarshallingOfCharPtr()
     {
         var obj = new CheckMarshllingOfCharPtr();
-        var str = obj.FuncRetCharPtr;
-        Assert.AreEqual('S', str[0]);
+
+        var wstr = obj.FuncRetWideCharPtr;
+        Assert.AreEqual('S', wstr[0]);                                     //PASS!
+        var wstrBuilt = new StringBuilder(new string(wstr));
+        wstrBuilt[0] = 'j';
+        obj.FuncWithWideCharPtr(wstrBuilt);
+        wstr = obj.FuncRetWideCharPtr;
+        Assert.AreEqual('j', wstr[0]);                                     //PASS!
+
         
+        var str = obj.FuncRetCharPtr;
+        //Assert.AreEqual('S', str[0]);                                    //FAIL!
         var obj2 = new CheckMarshllingOfCharPtr();
-        var strBuilt = new StringBuilder(str);
+        var strBuilt = new StringBuilder(new string(str));
         strBuilt[0] = 'j';
         obj2.FuncWithCharPtr(strBuilt);                      
         var astr = obj2.FuncRetCharPtr;
-        /*Assert.AreEqual('j', astr[0]);*/
-        /*
-        var wstr = obj.FuncRetWideCharPtr;
-        Assert.AreEqual('S', wstr[0]);
-        var wstrBuilt = new StringBuilder(wstr);
-        wstrBuilt[0] = 'j';
-        obj.FuncWithWideCharPtr(wstrBuilt);                  //THIS LINE
-        wstr = obj.FuncRetWideCharPtr;                  //          causes this one to Abort
-        Assert.AreEqual('j', wstr[0]);
-
-        var fstr = CSharpTemp.CSharpTemp.FreeFuncWithCharPtrRet(strBuilt);    //EVEN this aborts
-        Assert.AreEqual('t', fstr[1]);*/
+        //Assert.AreEqual('j', astr[0]);                                    //FAIL!
+        
+        var fstr = CSharpTemp.CSharpTemp.FreeFuncWithCharPtrRet(strBuilt); 
+        //Assert.AreEqual('t', fstr[1]);                                    //FAIL!
     }
 }
--- a/tests/CSharpTemp/CSharpTemp.Tests.cs
+++ b/tests/CSharpTemp/CSharpTemp.Tests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Reflection;
+using System.Text;
 using CppSharp.Utils;
 using CSharpTemp;
 using NUnit.Framework;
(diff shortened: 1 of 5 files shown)


ddobrev:
Looks fine to me. I just don't know why you check for constancy, const char* is already marshalled as string. StringBuilder is just for char*. Perhaps the title of the issue is misleading, I'll correct it.

genuinelucifer:
If i don't check for non-const char* then it is messing up with marshalling of const char* and breaking the build!

ddobrev:
I see, I had misread your code. Could you please simplify it to:

if (ptrType != null && !CSharpTypePrinter.IsConstCharString(param.Type as PointerType) && ...)

genuinelucifer:
Is it clearer now?
And should we marshal functions return types to StringBuilder too? And what about class fields?
I just changed it for class properties now.
Also, should we change the fields in Internal class to StringBuilder from char*?
Moreover, the tests are failing! Will have to look into that.
Sorry for so many questions. I just wanted to leave it as a note too. Will try to work these out tomorrow.
Any suggestions?

ddobrev:
It's better, thanks. Shouldn't you ignore non-pointers altogether, though?
For return types you should check. I may have read something about returning StringBuilders being impossible but I am not sure. If it's possible, we should have it, yes.
Fields would be nice too but at the moment I am not sure how we can marshal them. If it's little effort, change it just for parameters and - if possible - return types. If it's too difficult to make a difference for the type of a field and a parameter, let me know and I'll help you research this problem.
Internal fields must stay the same, the code won't compile otherwise.
About the failing test I can only give a tip if you tell me what the error is.

genuinelucifer:
Yes. I will edit that bit to ignore non- pointers.
I read a few places and found that StringBuilders CAN be returned and didn't encounter any error in the build due to this. SO i am assuming that they can be returned.
Yes, it seems it will be dificult to marshal the fields.
I didn't try the internal fields yet. Won't touch them now.
.
The test is failing at assert. I am storing 'S','t','r','i','n','g' in a pointer and returning the pointer. But the test is failing on comparing..

tests/CSharpTemp/CSharpTemp.cpp
```diff
@@ -672,3 +672,43 @@ int TypeMappedWithOperator::operator |(int i)
...
+CheckMarshllingOfCharPtr::CheckMarshllingOfCharPtr()
+{
+	str = new char[25];
+	str[0] = 'S';
+	str[1] = 't';
+	str[2] = 'r';
+	str[3] = 'i';
+	str[4] = 'n';
+	str[5] = 'g';
+	str[6] = '\0';
```
genuinelucifer: Stored values here.

tests/CSharpTemp/CSharpTemp.Tests.cs
```diff
@@ -344,4 +345,12 @@ public void TestNullAttributedFunctionPtr()
         }
     }
 
+    [Test]
+    public unsafe void TestMarshallingOfCharPtr()
+    {
+        var obj = new CheckMarshllingOfCharPtr();
+        var str = obj.FuncRetCharPtr;
+        Assert.AreEqual('S', str[0]);
+        Assert.AreEqual('i', str[3]);
```
genuinelucifer: The tests fails to acknowledge any value. Both asserts fail.

ddobrev: What values do you get instead? Also, please try this in native code instead: str = "String". It should be the same but it's simpler and it's actually the way C/C++ strings are used in practice.

genuinelucifer: I was getting 253/254 instead of 83 (for 'S') that was because I had done typo while checking return types to mention the return type marshalling. Now the tests have passed. Will push that and a few other changes tomorrow.

ddobrev: Excellent. :)

genuinelucifer:
@tritao @ddobrev Now when I run the tests then Resharper simply aborts and doesn't show any error message. Might be something related to memory addresses being passed between codes causing SIGSEV. I have marked the lines with "//THIS LINE" in the tests which cause the tests to abort.
Insights?!

ddobrev:
@genuinelucifer could you try replacing the array assignment with a string one? That is, str = "String". I think you'll get the same behaviour but we want it anyway. If you do get the same, please try removing the "s[0] = j;" line and see if the test still crashes.

genuinelucifer:
@ddobrev i did both but it still crashes. :(

genuinelucifer:
@ddobrev If i use ```str="String";``` then the test crashes at the first call to any function.
While if I assign each elements of 'str' then atleast 1 call is survived.
Also, as far as i can remember doing ```str="String";``` should have cause some error related to 'LValue' becuase str is NOT a string object. It is a pointer.

ddobrev:
@genuinelucifer it makes sense. Sorry about this, I should've guessed earlier. In that case str is a local variable which gets destroyed and its address becomes invalid. So let's stick to your array assignment after all. Just please add a destructor to your class which deletes the two arrays, we don't want to leak memory.
I am afraid I don't have other ideas at the moment so let's do the following. Make sure we work properly with char* parameters. Once that is fully dealt with, we'll get back to returning char* as well.

genuinelucifer:
@ddobrev The tests pass now!
But, is it fine to change the Encoding in managed code?
I mean I thought that we should get the marshalled code in required encoding...

tests/CSharpTemp/CSharpTemp.Tests.cs
```diff
@@ -344,4 +345,34 @@ public void TestNullAttributedFunctionPtr()
...
+        var wstrBuilt = new StringBuilder(new string(wstr));
+        wstrBuilt[0] = 'j';
+        obj.FuncWithWideCharPtr(wstrBuilt);
+        wstr = obj.FuncRetWideCharPtr;
+        Assert.AreEqual('j', wstr[0]);
+
+        var obj2 = new CheckMarshllingOfCharPtr();
+        var str = obj2.FuncRetCharPtr;
+        byte[] bytes = Encoding.UTF32.GetBytes(new string(str));
+        var myString = Encoding.ASCII.GetString(bytes);
```
genuinelucifer: This is what is worrying me!

ddobrev:
@genuinelucifer from what I see it looks like to me that my suspicions about the impossibility to marshal returning char* as StringBuilders might be true. I am not certain these encoding fixes are correct either. So I would suggest implementing the marshalling of just char* parameters as the first half of this task. That is, similar to the task for ref, where you only marshalled parameters. When that's all done and tested, we'll merge it and then continue on top with returning char*.

tests/CSharpTemp/CSharpTemp.Tests.cs
ddobrev: The correct way to do it is with Marshal.PtrToStringUni:

    var myString = Marshal.PtrToStringUni(obj2.FuncRetCharPtr);

Let me know if you have problems with it.

genuinelucifer:
@ddobrev Using ```PtrToStringAnsi``` makes 2 Asserts fail. While if I use them with ```Encoding.UTF32.GetBytes``` and ```Encoding.ASCII.GetString``` then they pass!!
What I actually wanted to ask previously was that even though they are giving required results, do we want our users to do such tedious jobs of getting the encoding right?! Or should we rather add these lines to the functions themselves while returning and return either a ```string``` or a ```StringBuilder``` in Native code. No need to change the Marshalling code.

ddobrev:
@genuinelucifer this task seems more complicated than I had thought. I think I'd better take over. I appreciate your work and I am definitely going to reuse your code. However, I think I need to take a look myself.
In the mean time, could you please redirect your efforts to https://github.com/mono/CppSharp/issues/498 ? It is quite important too, more important than this one actually.

genuinelucifer:
@ddobrev Thanks. I had already started work on that a few weeks ago. Will resume work on that.

genuinelucifer:
Stopping work Moreover has merge conflicts. SO closing...

PR #536, opened 2015-08-09
Fixed code gen for Value Type CS_OUT params

genuinelucifer:
Also polished (sort of) a previous test written by me.
And added an entry in .gitignore to ignore temp files.
.
Also I could not move the function definitions to .cpp file else the CLI part could not find the definition of these functions.

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -2547,7 +2547,6 @@ namespace CppSharp.Generators.CSharp
             foreach (var paramInfo in @params)
             {
                 var param = paramInfo.Param;
-                Class @class;
                 if (!(param.IsOut || param.IsInOut)) continue;
                 if (param.IsPrimitiveParameterConvertibleToRef())
                     continue;
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -2616,8 +2616,7 @@ namespace CppSharp.Generators.CSharp
                 var paramType = param.Type;
 
                 Class @class;
-                if ( (paramType.GetFinalPointee() ?? paramType).Desugar().TryGetClass(out @class)
-                      && @class.IsRefType)
+                if ((paramType.GetFinalPointee() ?? paramType).Desugar().TryGetClass(out @class))
                 {
                     WriteLine("{0} = new {1}();", param.Name, paramType);
                 }
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -2547,6 +2547,7 @@ namespace CppSharp.Generators.CSharp
             foreach (var paramInfo in @params)
             {
                 var param = paramInfo.Param;
+                Class @class;
                 if (!(param.IsOut || param.IsInOut)) continue;
                 if (param.IsPrimitiveParameterConvertibleToRef())
                     continue;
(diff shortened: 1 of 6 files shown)


genuinelucifer:
@tritao @ddobrev 
Pls review.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -2616,10 +2616,10 @@ public struct ParamMarshal
                 var paramType = param.Type;
 
                 Class @class;
-                if ( (paramType.GetFinalPointee() ?? paramType).Desugar().TryGetClass(out @class)
-                      && @class.IsRefType)
+                if ( (paramType.GetFinalPointee() ?? paramType).Desugar().TryGetClass(out @class))
```
ddobrev: A redundant space  - '( ('.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -2616,10 +2616,10 @@ public struct ParamMarshal
...
                 {
-                    WriteLine("{0} = new {1}();", param.Name, paramType);
+                    if(@class.IsRefType || @class.IsValueType)
```
ddobrev: A missing space after the 'if'. Also, doesn't this check actually cover all cases and is therefore redundant? If it's not, wouldn't !@class.IsInterface be better?

ddobrev:
Looks OK to me save for some minor remarks. However, I am not familiar with this part of C++# so @tritao you should check it out.

tritao:
I think it looks good, but we should be trying to call these new test functions from the NUnit tests. Just making sure they generate and compile doesn't tell us that they actually work when called at runtime.

genuinelucifer:
@tritao Done. I actually thought that generation would gaurantee functioning. But, after thinking for a while it now makes much sense to test them. Thanks.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -2616,8 +2616,8 @@ public struct ParamMarshal
                 var paramType = param.Type;
 
                 Class @class;
-                if ( (paramType.GetFinalPointee() ?? paramType).Desugar().TryGetClass(out @class)
-                      && @class.IsRefType)
+                if ((paramType.GetFinalPointee() ?? paramType).Desugar().TryGetClass(out @class)
+                    && !@class.IsInterface)
```
tritao: Do we have any test for the `!IsInterface` case?

tests/Basic/Basic.Tests.cs
```diff
@@ -522,5 +522,17 @@ public void TestMarshallingEmptyType()
     {
         var empty = new ReturnsEmpty().Empty;
     }
+
+    [Test]
+    public void TestClassesPassTry()
+    {
+        RefTypeClassPassTry refTypeClassPassTry;
```
tritao: You're still not asserting anything. We should have some fields and make sure they're being marshaled correctly.

tritao:
Nevermind about the asserts, I think we're already well-covered. I'd just like to see one test for the interface case.

genuinelucifer:
@tritao I used following in CSharpTemp.h
```C++
void DLL_API funcTryInterfaceTypePtrOut(CS_OUT TestParamToInterfacePassBaseTwo* classTry) { }
void DLL_API funcTryInterfaceTypeOut(CS_OUT TestParamToInterfacePassBaseTwo classTry) { }
```
In the generated code parameter types were changed to respective Interfaces but they were NOT out parameters!
So, if we are not generating out parameters for Interface types then we don't need any checks. So i removed the check ```!@class.IsInterface```
Although, I guess we should generate out parameters for interface types too...
Or is there any limitation/hindrance to this?

genuinelucifer:
@ddobrev @tritao Should i try to generate 'out' qualifier for interfaces too?

ddobrev:
@genuinelucifer yes. Interfaces do not exist on their own, they are added in to support multiple inheritance. This means that existing classes implement those interfaces and if there's any mismatch in signatures, the generated code won't compile.

genuinelucifer:
@tritao @ddobrev I added the proper code for interface types too. Please review the changes. Will submit a new PR (Removing the conflicts) within a couple of hours.

genuinelucifer:
After pulling the recent commits, I have not been able to build the master branch. I am upgrading llvm and clang. And rebuilding everything. Will submit a PR after solving these issues.

ddobrev:
Sorry about that @genuinelucifer, we forgot to tell you. We were given the opportunity to have an OS X build and decided that since we had to build binaries for OS X anyway, we might as well update to the then latest LLVM/Clang.

genuinelucifer:
@ddobrev No problem. The errors revealed themselves that there was some
problem with my llvm. And good to hear about the new OS X build!  :)

PR #539, opened 2015-08-13
Added marshalling of fixed size non primitive type arrays.

genuinelucifer:
Test left. Will add soon.

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpMarshal.cs
+++ b/src/Generator/Generators/CSharp/CSharpMarshal.cs
@@ -111,6 +111,7 @@ namespace CppSharp.Generators.CSharp
             if (!VisitType(array, quals))
                 return false;
 
+            Class @class;
             switch (array.SizeType)
             {
                 case ArrayType.ArraySize.Constant:
@@ -124,6 +125,9 @@ namespace CppSharp.Generators.CSharp
                     if (array.Type.IsPointerToPrimitiveType(PrimitiveType.Void))
                         supportBefore.WriteLineIndent("{0}[i] = new global::System.IntPtr({1}[i]);",
                             value, Context.ReturnVarName);
+                    else if (array.Type.Desugar().TryGetClass(out @class) && @class.IsRefType)
+                        supportBefore.WriteLineIndent("{0}[i] = {1}.{2}(*(({1}.Internal*)&({3}[i * sizeof({1}.Internal)])));", 
+                            value, array.Type, Helpers.CreateInstanceIdentifier, Context.ReturnVarName);
                     else
                         supportBefore.WriteLineIndent("{0}[i] = {1}[i];", value, Context.ReturnVarName);
                     supportBefore.WriteCloseBraceIndent();
@@ -382,16 +386,27 @@ namespace CppSharp.Generators.CSharp
             if (!VisitType(array, quals))
                 return false;
 
+            Class @class;
             switch (array.SizeType)
             {
                 case ArrayType.ArraySize.Constant:
                     var supportBefore = Context.SupportBefore;
                     supportBefore.WriteLine("if ({0} != null)", Context.ArgName);
                     supportBefore.WriteStartBraceIndent();
+                    if (array.Type.Desugar().TryGetClass(out @class) && @class.IsRefType)
+                    {
+                        supportBefore.WriteLine("if (value.Length != {0})", array.Size);
+                        supportBefore.WriteLineIndent("throw new ArgumentOutOfRangeException(\"{0}\", \"The provided array's dimensions doesn't match the required size.\");",
+                            Context.Parameter.Name);
+                    }
                     supportBefore.WriteLine("for (int i = 0; i < {0}; i++)", array.Size);
-                    supportBefore.WriteLineIndent("{0}[i] = {1}[i]{2};",
-                        Context.ReturnVarName, Context.ArgName,
-                        array.Type.IsPointerToPrimitiveType(PrimitiveType.Void) ? ".ToPointer()" : string.Empty);
+                    if (@class != null && @class.IsRefType)
+                        supportBefore.WriteLineIndent("{0}[i * sizeof({2}.Internal)] = *((byte*)({2}.Internal*){1}[i].__Instance);",
+                            Context.ReturnVarName, Context.ArgName, array.Type);
+                    else
+                        supportBefore.WriteLineIndent("{0}[i] = {1}[i]{2};",
+                            Context.ReturnVarName, Context.ArgName,
+                            array.Type.IsPointerToPrimitiveType(PrimitiveType.Void) ? ".ToPointer()" : string.Empty);
                     supportBefore.WriteCloseBraceIndent();
                     break;
                 default:
--- a/tests/CSharpTemp/CSharpTemp.Tests.cs
+++ b/tests/CSharpTemp/CSharpTemp.Tests.cs
@@ -393,4 +393,33 @@ public class CSharpTempTests : GeneratorTestFixture
         obj.FuncPrimitivePtrToRefWithDefVal(ref refInt, null, null, ref refInt);
         obj.FuncPrimitivePtrToRefWithMultiOverload(ref refInt, null, null, ref refInt);
     }
+
+    [Test]
+    public void TestFixedArrayRefType()
+    {
+        Foo[] foos = new Foo[4];
+        foos[0] = new Foo();
+        foos[0].A = 5;
+        foos[1] = new Foo();
+        foos[1].A = 6;
+        foos[2] = new Foo();
+        foos[2].A = 7;
+        foos[3] = new Foo();
+        foos[3].A = 8;
+        Bar bar = new Bar();
+        bar.Foos = foos;
+
+        Foo[] retFoos = bar.Foos;
+        Assert.AreEqual(5, retFoos[0].A);
+        Assert.AreEqual(6, retFoos[1].A);
+        Assert.AreEqual(7, retFoos[2].A);
+        Assert.AreEqual(8, retFoos[3].A);
+
+        Foo[] foosMore = new Foo[2];
+        foosMore[0] = new Foo();
+        foosMore[1] = new Foo();
+        var ex = Assert.Throws<ArgumentOutOfRangeException>(() => bar.Foos = foosMore);
+        Assert.AreEqual("value", ex.ParamName);
+        Assert.AreEqual("The provided array's dimensions doesn't match the required size.\r\nParameter name: value", ex.Message);
+    }
 }
--- a/src/Generator/Passes/CheckIgnoredDecls.cs
+++ b/src/Generator/Passes/CheckIgnoredDecls.cs
@@ -341,11 +341,13 @@ namespace CppSharp.Passes
                 return true;
             }
 
+            Class @class;
             var arrayType = type as ArrayType;
             PrimitiveType primitive;
             if (arrayType != null && arrayType.SizeType == ArrayType.ArraySize.Constant &&
                 !arrayType.Type.IsPrimitiveType(out primitive) &&
-                !arrayType.Type.Desugar().IsPointerToPrimitiveType())
+                !arrayType.Type.Desugar().IsPointerToPrimitiveType() &&
+                !(arrayType.Type.Desugar().TryGetClass(out @class) && @class.IsRefType))
             {
                 msg = "unsupported";
                 return true;
--- a/tests/CSharpTemp/CSharpTemp.h
+++ b/tests/CSharpTemp/CSharpTemp.h
@@ -65,11 +65,11 @@ public:
     const Bar& operator++();
     Bar operator++(int i);
     void* arrayOfPrimitivePointers[1];
+    Foo foos[4];
 
 private:
     int index;
     Foo m_foo;
-    Foo foos[4];
 };
 
 Bar::Bar() {}


src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -124,6 +125,8 @@ public override bool VisitArrayType(ArrayType array, TypeQualifiers quals)
                     if (array.Type.IsPointerToPrimitiveType(PrimitiveType.Void))
                         supportBefore.WriteLineIndent("{0}[i] = new global::System.IntPtr({1}[i]);",
                             value, Context.ReturnVarName);
+                    else if(array.Type.TryGetClass(out @class) && @class.IsRefType)
```
ddobrev: A missing space after the "if".
More importantly, I think we may need some desugaring/final pointee/etc on array.Type.

genuinelucifer: Could you please explain what does desugaring does?! Sorry I looked at the code but couldn't fully understand its use case.

ddobrev: It gets the real type under a type def, for example.

genuinelucifer:
Can we support 2D float arrays with same method? I think C# does support multidimentional arrays. Why don't we support it yet?

genuinelucifer:
My Resharper license expired yesterday so I haven't been able to run the tests. I hope they are working. I am trying to get rid of the error "NUnit Test RUnner" is giving. Once i figure it out i will run the tests and comment here...

genuinelucifer:
The Test is failing when the getter is called at the line
```C#
Foo[] retFoos = bar.Foos;
```
with the error
```
Test Name:	TestFixedArrayRefType
Test FullName:	CSharpTempTests.TestFixedArrayRefType
Test Source:	e:\GSoC\MonoProject_2015\CppSharp_New\CppSharp\tests\CSharpTemp\CSharpTemp.Tests.cs : line 399
Test Outcome:	Failed
Test Duration:	0:00:00.042

Result Message:	System.NullReferenceException : Object reference not set to an instance of an object.
Result StackTrace:	
at CSharpTemp.Bar.get_Foos() in e:\GSoC\MonoProject_2015\CppSharp_New\CppSharp\build\vs2013\gen\CSharpTemp\CSharpTemp.cs:line 920
at CSharpTempTests.TestFixedArrayRefType() in e:\GSoC\MonoProject_2015\CppSharp_New\CppSharp\tests\CSharpTemp\CSharpTemp.Tests.cs:line 412
```

genuinelucifer:
The generated code for the getter is
```C#
                Foo[] __value = null;
                if (((Internal*) __Instance)->foos != null)
                {
                    __value = new Foo[4];
                    for (int i = 0; i < 4; i++)
                        __value[i] = Foo.__CreateInstance(*((Foo.Internal*)(((Internal*) __Instance)->foos[i * sizeof(Foo.Internal)])));
                }
                return __value;
```
Also the error is at the line where ```value[i] = Foo.__CreateInstance(...``` is called.
@ddobrev @tritao Any suggestions as to what might be going wrong?

ddobrev:
@genuinelucifer as a start, could you please debug it and determine which piece in this long line is null?

ddobrev:
Also, why do you use i * sizeof(Foo.Internal) as the index? foos only has 4 elements, you are almost guaranteed to crash.

genuinelucifer:
foos is a 128 bytes array.
Here is the internal field definition:
```C#
public fixed byte foos[128];
```

genuinelucifer:
Moreover, @tritao also used the same indexing on the issue discussion. So, i merely copied that.

ddobrev:
Oh, I see. Sorry about that. This leaves my previous suggestion, debug to see what exactly is null.

genuinelucifer:
I changed the getter to (manually)
```C#
                Foo[] __value = null;
                if (((Internal*) __Instance)->foos != null)
                {
                    __value = new Foo[4];
                    for (int i = 0; i < 4; i++)
                    {
                        var __ptr = ((Internal*)__Instance);
                        var val = __ptr->foos[i * sizeof(Foo.Internal)];
                        var rar = (Foo.Internal*)val;
                        var ins = Foo.__CreateInstance((System.IntPtr)val);
                        __value[i] = ins;
                        //__value[i] = Foo.__CreateInstance(*((Foo.Internal*)(((Internal*) __Instance)->foos[i * sizeof(Foo.Internal)])));
                    }
                        
                }
                return __value;
```
still the nullreference is in the line ```var ins = Foo.__CreateInstance((System.IntPtr)val);```

genuinelucifer:
Now it gives a weird nullreference exception for ```Foo.get_A()``` in the Assert line ```Assert.AreEqual(5, retFoos[0].A);``` and references this getter in the Foo class:
```C#
return ((Internal*) __Instance)->A;
```

genuinelucifer:
@tritao @ddobrev Finally everything is working and the tests are passing... Pls review again.

ddobrev:
@genuinelucifer I've added a single note, everything else looks fine to me.

genuinelucifer:
@ddobrev Done.

tritao:
Hey this looks good to me. I'm just wondering if we should be checking the size of the array when setting it. Since we now the maximum size of the array (4 in the test) then we can check that the user is not setting an array with a different size.

genuinelucifer:
That would probably be better. Should we throw some exception in that case?

tritao:
Yeah an exception would be nice.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -382,16 +385,26 @@ public override bool VisitArrayType(ArrayType array, TypeQualifiers quals)
...
             switch (array.SizeType)
             {
                 case ArrayType.ArraySize.Constant:
                     var supportBefore = Context.SupportBefore;
                     supportBefore.WriteLine("if ({0} != null)", Context.ArgName);
                     supportBefore.WriteStartBraceIndent();
+                    if (array.Type.Desugar().TryGetClass(out @class) && @class.IsRefType)
+                    {
+                        supportBefore.WriteLine("if (value.Length != {0})", array.Size);
+                        supportBefore.WriteLineIndent("throw new ArrayTypeMismatchException(\"The provided array's dimentions don't match the requred size.\");");
```
genuinelucifer: @tritao Is this ok. Or should we change the message? Or should i remove the message and throw empty exception?

tritao: Small typos: `dimensions`, `required`, and `doesn't`.

tritao:
That's perfect, let's just get a test that we throw the exception when the array size is not correct.

genuinelucifer:
@tritao Done.
Merging all commits now.

ddobrev:
@genuinelucifer could you please wait for a minute? I'd like to make some remarks.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -124,6 +125,8 @@ public override bool VisitArrayType(ArrayType array, TypeQualifiers quals)
                     if (array.Type.IsPointerToPrimitiveType(PrimitiveType.Void))
                         supportBefore.WriteLineIndent("{0}[i] = new global::System.IntPtr({1}[i]);",
                             value, Context.ReturnVarName);
+                    else if (array.Type.Desugar().TryGetClass(out @class) && @class.IsRefType)
+                        supportBefore.WriteLineIndent("{0}[i] = {1}.__CreateInstance(*(({1}.Internal*)&({2}[i * sizeof({1}.Internal)])));", value, array.Type, Context.ReturnVarName);
```
tritao: Actually, another small thing, use CreateInstanceIdentifier here instead of hardcoding the value.

tests/CSharpTemp/CSharpTemp.Tests.cs
```diff
@@ -393,4 +393,39 @@ public unsafe void TestMultiOverLoadPtrToRef()
...
+        Bar bar = new Bar();
+        bar.Foos = foos;
+
+        Foo[] retFoos = bar.Foos;
+        Assert.AreEqual(5, retFoos[0].A);
+        Assert.AreEqual(6, retFoos[1].A);
+        Assert.AreEqual(7, retFoos[2].A);
+        Assert.AreEqual(8, retFoos[3].A);
+
+        try
```
ddobrev: There are methods in NUnit such as Assert.Throws, Assert.Catch and similar, they are preferable to the manual solution.

tritao:
Check the inline comments I added, they're small nitpicks but worth fixing.

src/Generator/Generators/CSharp/CSharpMarshal.cs
ddobrev: The message is OK, I think, however we should choose a better exception type. Let me check...

src/Generator/Generators/CSharp/CSharpMarshal.cs
genuinelucifer: I'm sorry. CreateInstanceIdentifier for what exactly?

tritao: Instead of hardcoding `__CreateInstance`.

src/Generator/Generators/CSharp/CSharpMarshal.cs
ddobrev: I believe ArgumentOutOfRangeException makes a little more sense.

genuinelucifer:
@ddobrev @tritao Sorry if i missed some comments. I just merged the commits and can see only 1 comment from each of you.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -382,16 +385,26 @@ public override bool VisitArrayType(ArrayType array, TypeQualifiers quals)
...
             switch (array.SizeType)
             {
                 case ArrayType.ArraySize.Constant:
                     var supportBefore = Context.SupportBefore;
                     supportBefore.WriteLine("if ({0} != null)", Context.ArgName);
                     supportBefore.WriteStartBraceIndent();
+                    if (array.Type.Desugar().TryGetClass(out @class) && @class.IsRefType)
+                    {
+                        supportBefore.WriteLine("if (value.Length != {0})", array.Size);
+                        supportBefore.WriteLineIndent("throw new ArrayTypeMismatchException(\"The provided array's dimensions doesn't match the required size.\");");
```
ddobrev: I believe ArgumentOutOfRangeException makes a little more sense for the exception type.

ddobrev:
@genuinelucifer no problem, I've readded mine. @tritao should I readd yours too?

src/Generator/Generators/CSharp/CSharpMarshal.cs
genuinelucifer: From my point of view, ArgumentOutOfRangeException somewhat tells that the given array was ```larger``` than the one to be assigned to. While here we are throwing exception either if the array was smaller or larger. Your view on this?

tests/CSharpTemp/CSharpTemp.Tests.cs
genuinelucifer: Thanks. Done.

src/Generator/Generators/CSharp/CSharpMarshal.cs
ddobrev: Out of range does not imply which side of the range the value is out of. For example, if your List<int> has a size of 5 and you request the value at index -1, you'd still get an IndexOutOfRangeException even though -1 < 5.

genuinelucifer: Hmmm. That is okay. What I want to say that suppose (like the test) the array is of size 4 and the user gave array of size 2. Then the indexes 0 & 1 are all within range. But exception will be thrown.

ddobrev: The example I gave you involves a different exception, IndexOutOfRangeException. I am suggesting ArgumentOutOfRangeException.

src/Generator/Generators/CSharp/CSharpMarshal.cs
genuinelucifer: Done.

src/Generator/Generators/CSharp/CSharpMarshal.cs
genuinelucifer: Ohh. Soryy I misunderstood that with IndexOutOfBounds. Yes, ```ArgumentOutOfRangeException``` seems better.

ddobrev:
Everything looks perfect now. @tritao he's also fixed the typos so you can merge now.

ddobrev:
@genuinelucifer I may have been too quick with my last message, have you pushed the updated exception type as well as @tritao 's "__CreateInstance" remark.

genuinelucifer:
I haven't pushed yet. I will in a couple of minutes.

tritao:
Value is fine, no point losing valuable time over this.

genuinelucifer:
@tritao :+1: 
I am pushing all changes merged into one now...

genuinelucifer:
@tritao Done

tests/CSharpTemp/CSharpTemp.Tests.cs
```diff
@@ -393,4 +393,33 @@ public unsafe void TestMultiOverLoadPtrToRef()
...
+        Assert.AreEqual(6, retFoos[1].A);
+        Assert.AreEqual(7, retFoos[2].A);
+        Assert.AreEqual(8, retFoos[3].A);
+
+        Foo[] foosMore = new Foo[2];
+        foosMore[0] = new Foo();
+        foosMore[1] = new Foo();
+        var ex = Assert.Throws<ArgumentOutOfRangeException>(() => bar.Foos = foosMore);
+        Assert.AreEqual("value", ex.ParamName);
+        Assert.AreEqual("The provided array's dimensions doesn't match the required size.\r\nParameter name: value", ex.Message);
```
ddobrev: This line fails the Linux build. Please use Environment.NewLine instead of the hard-coded "\r\n".

Issue #438, opened 2015-04-10
CS_OUT arguments not assigned


jamesford42:
Trying to use CS_OUT on an argument that is a class marked with CS_API and passed in as a pointer (intending that the method call will dereference and assign a new value), results in the error:
Use of unassigned out parameter 'ExampleType'.

In particular this line is the problem. This testing of if the parameter is null prior to deciding to allocate a new one is invalid for an 'out' parameter in C#. 
```C#
var arg0 = packet == (ExampleType) null ? global::System.IntPtr.Zero : packet.__Instance;
```

genuinelucifer:
I recreated the issue with :
https://gist.github.com/genuinelucifer/9e23701dee397e5c41c3
.
What i guess is both the functions should generate the same C# code for ref type class.. I am not clear as what a value type 'out' argument would act as! Also i am not completely sure about what should be in C++/CLI code for these.
Please guide me as to what should be the expected output in each case.

tritao:
You're right, we should generate the same code for ref type class. Well it might have minor differences but it should mostly look the same.

And in C++/CLI case I think it needs to use a tracking reference (written as `%` type modifier). See 
https://msdn.microsoft.com/en-us/library/6f63s5b1.aspx.

genuinelucifer:
Thanks. Will read about the tracking reference and then try to implement it.

genuinelucifer:
I fixed the CSharp side of the code generation with :
https://gist.github.com/genuinelucifer/d73aa4039903b0c06e6d
.
I assume it was the expected solution.
.
But, i am NOT able to implement on the C++/CLI side because I am unable to figure out how the C++/CLI side is differentiating between "CS_OUT RefClass param" and "CS_OUT RefClass* param" when it is generating only "RefClass^% param" for both!
I just know that the difference comes at the point where param.Visit is called at line number 996 in CLISOurcesTemplate.cs
```C#
                var marshal = new CLIMarshalNativeToManagedPrinter(ctx);
                param.Visit(marshal);
```
Please provide furthur guidance on how to implement it,

jamesford42:
Unforunately, I am not familiar C++/CLI so can offer no guidance there.

tritao:
@genuinelucifer I didn't understand your question about CLI. What is the problem?

genuinelucifer:
[...]

PR #548, opened 2015-08-27
Fixed code gen for Out params (Value type and interfaces)

genuinelucifer:
@tritao @ddobrev Please review.

### DIFF
--- a/tests/Basic/Basic.h
+++ b/tests/Basic/Basic.h
@@ -759,16 +759,10 @@ public:
     } ct;
 };
 
-#define CS_API
-class CS_API ClassPassTry
-{
-public:
-	int n;
-	char c;
-};
+class DLL_API RefTypeClassPassTry { };
 
-void funcTry(CS_OUT ClassPassTry* classTry) { }
-void funcTry2(CS_OUT ClassPassTry classTry) { }
+void DLL_API funcTryRefTypePtrOut(CS_OUT RefTypeClassPassTry* classTry);
+void DLL_API funcTryRefTypeOut(CS_OUT RefTypeClassPassTry classTry);
 
 #define ARRAY_LENGTH 5
 #define CS_VALUE_TYPE
@@ -805,3 +799,8 @@ class DLL_API ReturnsEmpty
 public:
     Empty getEmpty();
 };
+
+class DLL_API CS_VALUE_TYPE ValueTypeClassPassTry { };
+
+void DLL_API funcTryValTypePtrOut(CS_OUT ValueTypeClassPassTry* classTry);
+void DLL_API funcTryValTypeOut(CS_OUT ValueTypeClassPassTry classTry);
--- a/tests/CSharpTemp/CSharpTemp.h
+++ b/tests/CSharpTemp/CSharpTemp.h
@@ -700,3 +700,10 @@ class DLL_API TestVariableWithFixedArrayType
 public:
 	static Foo variableWithFixedArrayType[2];
 };
+
+class DLL_API TestOutTypeInterfaces
+{
+public:
+	void funcTryInterfaceTypePtrOut(CS_OUT TestParamToInterfacePassBaseTwo* classTry);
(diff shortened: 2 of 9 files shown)


tests/CSharpTemp/CSharpTemp.h
```diff
@@ -700,3 +700,10 @@ class DLL_API TestVariableWithFixedArrayType
 public:
 	static Foo variableWithFixedArrayType[2];
 };
+
+class DLL_API TestOutTypeInterfaces
```
ddobrev: I don't see how this test code alone causes the generation of interfaces, they are only created with multiple inheritance.

genuinelucifer: It alone doesn't create the interfaces. I am using ```TestParamToInterfacePassBaseTwo``` which  is secondary base for ```TestParamToInterfacePass``` class and hence is marshalled as interface by paramToInterfacePass...
I thought it would be better to use existing classes instead of creating dummy classes.

ddobrev: Right. Sorry, I was a little tired last night. @tritao I am OK with this pull.

tests/Basic/Basic.h
```diff
@@ -760,15 +760,10 @@ class TestNamingAnonymousTypesInUnion
 };
 
 #define CS_API
-class CS_API ClassPassTry
-{
-public:
-	int n;
-	char c;
-};
+class CS_API RefTypeClassPassTry { };
```
tritao: What's the purpose of adding `CS_API` here and not `DLL_API`?

genuinelucifer: When I wrote this test the first time then it was to address an issue which
specifically mentioned failure on using CS_API, so I left it as it was.
Should I change it to DLL_API?!
On 28-Aug-2015 5:52 PM, "João Matos" <notifications@github.com> wrote:

In tests/Basic/Basic.h
<https://github.com/mono/CppSharp/pull/548#discussion_r38194070>:

> @@ -760,15 +760,10 @@ class TestNamingAnonymousTypesInUnion
>  };
>
>  #define CS_API
> -class CS_API ClassPassTry
> -{
> -public:
> -	int n;
> -	char c;
> -};
> +class CS_API RefTypeClassPassTry { };

What's the purpose of adding CS_API here and not DLL_API?

—
Reply to this email directly or view it on GitHub
<https://github.com/mono/CppSharp/pull/548/files#r38194070>.

tritao: Do you remember what the issue with `CS_API` was? At first I was thought that it was being specially processed by `CheckMacrosPass` or something but I checked the source that seemed not to be the case.

genuinelucifer: #438 .. I don't think the main problem was due to CS_API. It was just that the one who filed the issue said so. But, the main problem was wrong code generation.

tests/Basic/Basic.h
```diff
@@ -760,15 +760,10 @@ class TestNamingAnonymousTypesInUnion
...
 
-void funcTry(CS_OUT ClassPassTry* classTry) { }
-void funcTry2(CS_OUT ClassPassTry classTry) { }
+void DLL_API funcTryRefTypePtrOut(CS_OUT RefTypeClassPassTry* classTry) { }
```
ddobrev: As usual, could you please replace these inlined implementations with real ones (in the CPP)?

tests/Basic/Basic.h
```diff
@@ -805,3 +800,8 @@ class DLL_API ReturnsEmpty
 public:
     Empty getEmpty();
 };
+
+class DLL_API CS_VALUE_TYPE ValueTypeClassPassTry { };
+
+void DLL_API funcTryValTypePtrOut(CS_OUT ValueTypeClassPassTry* classTry) { }
```
ddobrev: Some more inlined impls.

genuinelucifer:
@tritao I changed that CS_API to DLL_API. I guess both default to ```__dllexport``` so not much change really.
@ddobrev Sorry I always forget to do that. Finaly moved the definitions to cpp file.

PR #564, opened 2015-10-02
Very basic try for delegate separation.

genuinelucifer:
@ddobrev 
I have just started the delegate separation. Sorry for being so sloow. If you wish, you may take over this task. Otherwise, I will be happy to do it with some help.
.
Currently I am stuck on a IndexOutOfBounds Exception which occurs for all test projects except for Namespace project. In that too it is populating the parameters as expected. I am looking into it.


src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -1592,11 +1675,13 @@ private void GenerateVTableMethodDelegates(Class @class, Method method)
...
                 string.Join(", ", @params));
+             * */
+            var vTableMethodDelegateName = GenerateDelegate(method);
 
-            WriteLine("private static {0} {0}Instance;", vTableMethodDelegateName);
+            WriteLine("private static {0} {1}Instance;", vTableMethodDelegateName, vTableMethodDelegateName.Substring(17));
             NewLine();
 
-            WriteLine("private static {0} {1}Hook({2})", retType, vTableMethodDelegateName,
+            WriteLine("private static {0} {1}Hook({2})", retType, vTableMethodDelegateName.Substring(17),
```
ddobrev: I would guess it's here that you get out of range. Could you please elaborate what this piece is for?
By the way, if you fell it would go faster, feel free to contact me over Skype.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,79 @@ public void GenerateHeader()
...
+            Type t;
+            string retT, pars;
+            for (int i = 0; i < AllDelegatesUsed.Count; ++i)
+            {
+                dd = AllDelegatesUsed.ElementAt(i);
+                retT = dd.param.ElementAt(0).CSharpType(TypePrinter).ToString();
+                pars = "";
+                for (int j = 1; j <= dd.numParams; ++j)
+                {
+                    t = dd.param.ElementAt(i);
```
genuinelucifer: I am getting IndexOutOfBounds at this line. Although i am pretty sure I have done enough to ensure this wouldn't throw exception.

ddobrev:
@genuinelucifer nice to see you making further progress. I would like to give you a suggestion to consider. You could change AllDelegatesUsed to a Dictionary<string, DelegateDef> where the key is the signature (return type + parameter types) + the calling convention. You should be able to construct that key nearly the same way we use to generate delegates at the moment, and then you just check in the dictionary.

genuinelucifer:
@ddobrev Thanks. Will try to do that. :)

ddobrev:
Hello @genuinelucifer . I can see you've added some more work, nice. Have you been of need for any help?

genuinelucifer:
@ddobrev Yes indeed. I am not able to get equality true when the keys are compared. I looked online and tried many things but to no avail. Whenever the function ```GenerateDelegate``` is called it always generates new delegate because the ```Equals``` method always returns ```false```!
I tried to change the ```CompareWith``` method but to no avail.

genuinelucifer:
Previously with list, that part worked upto a certain extent. Only one case was left. I used similar code for comparision but really i find it amusing that it doesn't work.

ddobrev:
I see. Your dictionary uses a DelegateSign as the key while I actually had a simple string in mind. What I meant is this: with our present code we construct a string consisting of the name of the method and its parameters. My idea was to construct a similar key, only replacing the name of the method with some genetic term (as you have used "_Delegate") and add the calling convention. Then when you need a new delegate, you construct the key from its signature and then check for a match.
In any case, if you feel this dictionary takes too much of your time, you could proceed until you get the code working, and then we can refactor it to use a list.

genuinelucifer:
@ddobrev Thanks a lot for the help. I changed the key and finally it works! The number of delegates is minimised. Please review.

genuinelucifer:
@tritao @ddobrev One of the tests is failing on Linux only and not on OSX or Windows. Please look into it because I have no experience with regards to checking code for platform specific errors.

ddobrev:
@genuinelucifer do not worry about the test, the failure's unrelated to your changes. I am going to review your code as soon as I can, I hope I'll be able to do so tomorrow.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,138 @@ public void GenerateHeader()
...
+                {
+                    return false;
+                }
+                return true;
+            }
+        }
+
+        private string GenerateDelegateSignature(Function function, out CSharpTypePrinterResult retType)
+        {
+            var @params = new List<string>();
```
ddobrev: All of this code seems an identical copy of GatherInternalParams (https://github.com/mono/CppSharp/blob/master/src/Generator/Generators/CSharp/CSharpTextTemplate.cs#L600). Please reuse that method instead of repeating the code. In case there are some minor differences, introduce a parameter.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,138 @@ public void GenerateHeader()
...
+                if (Options.IsMicrosoftAbi && @class.Layout.HasVirtualBases)
+                    @params.Add("int " + GeneratedIdentifier("forBases"));
+            }
+
+            TypePrinter.PopContext();
+
+            return string.Join(", ", @params);
+        }
+
+        private Dictionary<DelegateSignature, string> AllDelegates = new Dictionary<DelegateSignature, string>();
```
ddobrev: My idea was actually the other way around - a string as the key and your object as the value. It works this way too but with a string key the code will be much simpler and shorter because you won't need all of your code for hashes and comparison.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,138 @@ public void GenerateHeader()
...
+        {
+            return GenerateDelegate(func as Method, suppressWarning);
+        }
+
+        private void GenerateDelegatesClass()
+        {
+            if (AllDelegates.Count == 0)
+                return;
+            WriteLine("");
+            WriteLine("public unsafe class DelegatesStorage");
```
ddobrev: I'd rather not have the delegates in a class because people would wonder what the class itself is for. Please generate a name-space instead, or place them directly in the root (without a name-space) - I am not sure at the moment which would be better.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,138 @@ public void GenerateHeader()
...
+            WriteStartBraceIndent();
+            DelegateSignature ds;
+            string delegateName;
+            for (int i = 0; i < AllDelegates.Count; ++i)
```
ddobrev: Simpler, and a little faster, to use a "foreach" instead.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,138 @@ public void GenerateHeader()
...
+            CSharpTypePrinterResult retType;
+            ds.signature = GenerateDelegateSignature(func, out retType);
+            ds.retType = retType.ToString();
+            ds.suppressWarning = suppressWarning;
+
+            string delegateName;
+            if (AllDelegates.TryGetValue(ds, out delegateName))
+                return "DelegatesStorage." + delegateName;
+
+            if (ds.retType.Equals("void"))
```
ddobrev: Please do not compare types as strings. Change retType to a Type (and print it when necessary) and use IsPrimitiveType(Void).

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,138 @@ public void GenerateHeader()
...
+                delegateName = "Func_" + AllDelegates.Count;
```
ddobrev: Such names would result in a terrible API: Func_1, Action_2, etc. Instead, please use the signatures in the name, similar to what we had so far. Func_int_MyClass_double, Action_float_MyOtherClass, etc. As per the Mono/.NET convention, in the case of Func put the return type first and the parameters later.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,138 @@ public void GenerateHeader()
             PopBlock();
         }
 
+        private class DelegateSignature
+        {
+            public string retType, signature;
```
ddobrev: As per Mono/.NET conventions, please capitalise public members. Also, as I mentioned once, public fields are bad, properties should be used instead. As I additionally mentioned then, we do have public fields in other places but that doesn't mean it's right, and we've been replacing them over time.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,138 @@ public void GenerateHeader()
...
+            WriteLine("");
+            WriteLine("public unsafe class DelegatesStorage");
+            WriteStartBraceIndent();
+            DelegateSignature ds;
+            string delegateName;
+            for (int i = 0; i < AllDelegates.Count; ++i)
+            {
+                ds = AllDelegates.ElementAt(i).Key;
+                delegateName = AllDelegates.ElementAt(i).Value;
+                if(ds.suppressWarning)
```
ddobrev: Please put a space behind the "if"-s.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,138 @@ public void GenerateHeader()
...
+                    WriteLine("[SuppressUnmanagedCodeSecurity]");
+                if(ds.callingConvention != null)
+                    WriteLine("[UnmanagedFunctionPointerAttribute(global::System.Runtime.InteropServices.CallingConvention.{0})]", ds.callingConvention);
+                WriteLine("internal delegate {0} {1}({2});", ds.retType, delegateName, ds.signature);
+                if(i<(AllDelegates.Count-1))
```
ddobrev: No need for the internal parenthesis, also put spaces around the operators.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
genuinelucifer: Will do.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
genuinelucifer: Well, only string as a key is not sufficient. Because as we once discussed on skype. 2 delegates with different calling conventions need to be treated as separate delegates. So, we need other info too for distinguishing delegates.

ddobrev: My suggestion says "and add the calling convention". That is, your key would look like Action_int_MyClass_ThisCall.

genuinelucifer: Ohh. Sorry i missed that part. Will try to implement.

genuinelucifer:
@ddobrev I incorporated a few of the changes. Will do other changes too after i figure out a way for naming convention of the delegates. Currently I am a bit confused on how to do it programatically.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,111 @@ public void GenerateHeader()
...
+            {
+                return Equals(obj as DelegateSignature);
+            }
+            public bool Equals(DelegateSignature obj)
+            {
+                return obj != null && CompareWith(obj);
+            }
+            private bool CompareWith(DelegateSignature ds)
+            {
+                if (!retType.Equals(ds.retType) || !signature.Equals(ds.signature) || suppressWarning != ds.suppressWarning ||
```
genuinelucifer: This line is always becomes says that the 2 DelegateSignatures are NOT equal as ```retType.Equals(ds.retType)``` always returns false and hence number of delegates increased again! I encountered this once more in past but ignored. Am I missing something here? Is this expected behaviour?!

ddobrev: Hm, in theory we might've implemented only == and not Equals properly. Could you please check if you get a correct result with ==?

ddobrev: If this takes too much time, please migrate to the string keys first - this way you won't have to deal with this to begin with.

ddobrev:
@genuinelucifer I've got really excited about your work, the code looks much better than I had imagined. :) When can you continue?

genuinelucifer:
@ddobrev Thanks a lot.
I have been really busy with many other projects. I have pushed some changes which, i think, incorporate all your suggestions. And moreover, the number of delegates is minimized correctly. Please review.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,80 @@ public void GenerateHeader()
             PopBlock();
         }
 
+
+        private class DelegateSignature
+        {
+            public string Signature { get; set; }
+            public bool SuppressWarning { get; set; }
+            public string CallingConvention { get; set; }
```
ddobrev: Please use the higher level types of CallingConvention and Type and call ToString when needed instead of calling it in advance as you do now.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,80 @@ public void GenerateHeader()
...
+            OptionalOut<string> allTypes = new OptionalOut<string>();
+            allTypes.Result = new List<string>();
+            var @params = GatherInternalParams(function, out retType, false, allTypes);
+
+            DelegateSignature ds = new DelegateSignature();
+            ds.CallingConvention = function.CallingConvention.ToInteropCallConv().ToString();
+            ds.SuppressWarning = suppressWarning;
+
+            allTypes.Result.Add(ds.CallingConvention);
+            allTypes.Result.Add(suppressWarning.ToString());
```
ddobrev: We don't need the value of "suppressWarning" in the delegate signature. It only indicates whether to put the SuppressWarning attribute which is ignored when invalid unlike the calling convention which causes a crash.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,80 @@ public void GenerateHeader()
...
+            return strName;
+        }
+        private string GenerateDelegate(Function func, bool suppressWarning = true)
+        {
+            return GenerateDelegate(func as Method, suppressWarning);
+        }
+
+        private void GenerateDelegatesClass()
+        {
+            if (AllDelegates.Count == 0)
```
ddobrev: Please remove this, it's redundant because if there are 0 elements, the foreach below simply won't execute.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,80 @@ public void GenerateHeader()
...
+                return;
+            DelegateSignature ds;
```
ddobrev: Please move these within the loop using "var":
var delegateName = e.Key;
var ds = e.Value;

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -621,18 +712,28 @@ private List<string> GatherInternalParams(Function function, out CSharpTypePrint
...
             }
 
             if (method != null && method.IsConstructor)
             {
                 var @class = (Class) method.Namespace;
                 if (Options.IsMicrosoftAbi && @class.Layout.HasVirtualBases)
+                {
                     @params.Add("int " + GeneratedIdentifier("forBases"));
+                    if (allTypesInOne != null)
+                        allTypesInOne.Result.Add("int_forBases");
```
ddobrev: "forBases" is not part of the type but rather an identifier so I think just "int" would be enough.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -196,6 +198,80 @@ public void GenerateHeader()
...
+            else
+                strName = "Action_" + strName;
+            DelegateSignature delegateSign;
+            if (AllDelegates.TryGetValue(strName, out delegateSign))
+                return strName;
+
+            AllDelegates.Add(strName, signature);
+            return strName;
+        }
+        private string GenerateDelegate(Function func, bool suppressWarning = true)
```
ddobrev: You seem not to use any member exclusive to Method so you can only leave one overload of GenerateDelegate taking a Function.

ddobrev:
@genuinelucifer I've left some notes about minor improvements. There are some other things to do but I am not sure at the moment how they should be changed so I'll take care of them. If you are too busy, just leave it as it is now and I'll fix the remarks too - I have in a way left notes to myself. :) In either case, you've the overwhelming majority of the work so this feature will go in history under your name. :)

ddobrev:
Thanks, man. :) I'll take it from here.

genuinelucifer:
@ddobrev Thank you for being such a nice guide throughout. :)

ddobrev:
Superseded by fbb2f941c49ee5d034880cd38f1935964d128fa8.

PR #592, opened 2015-12-04
Initial patch to free allocated VTables.

genuinelucifer:
I have simply done the FreeHGlobal call for each AllocHGlobal in ```SetupVTables()```
.
I have yet to test it.

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -1304,7 +1304,8 @@ namespace CppSharp.Generators.CSharp
         private void DeAllocateVTablesMS(Class @class, IList<VTableComponent> wrappedEntries,
             bool destructorOnly)
         {
-            if (destructorOnly)
+            var hasVirtualDtor = wrappedEntries.Any(e => e.Method.IsDestructor);
+            if (hasVirtualDtor)
             {
                 WriteLine("if (__ManagedVTablesDtorOnly != null)");
                 WriteStartBraceIndent();
@@ -1322,7 +1323,8 @@ namespace CppSharp.Generators.CSharp
         private void DeAllocateVTablesItanium(Class @class, IList<VTableComponent> wrappedEntries,
             bool destructorOnly)
         {
-            if (destructorOnly)
+            var hasVirtualDtor = wrappedEntries.Any(e => e.Method.IsDestructor);
+            if (hasVirtualDtor)
             {
                 WriteLine("if (__ManagedVTablesDtorOnly != null)");
                 WriteStartBraceIndent();
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -1286,23 +1286,21 @@ namespace CppSharp.Generators.CSharp
             WriteStartBraceIndent();
             WriteLine("~Destructor()");
             WriteStartBraceIndent();
-            DeAllocateVTables(@class, wrappedEntries, false);
+            DeAllocateVTables(@class, wrappedEntries);
             WriteCloseBraceIndent();
             WriteCloseBraceIndent();
             NewLine();
         }
 
-        private void DeAllocateVTables(Class @class, IList<VTableComponent> wrappedEntries,
-            bool destructorOnly)
+        private void DeAllocateVTables(Class @class, IList<VTableComponent> wrappedEntries)
         {
             if (Options.IsMicrosoftAbi)
-                DeAllocateVTablesMS(@class, wrappedEntries, destructorOnly);
+                DeAllocateVTablesMS(@class, wrappedEntries);
             else
-                DeAllocateVTablesItanium(@class, wrappedEntries, destructorOnly);
+                DeAllocateVTablesItanium(@class, wrappedEntries);
         }
 
-        private void DeAllocateVTablesMS(Class @class, IList<VTableComponent> wrappedEntries,
-            bool destructorOnly)
+        private void DeAllocateVTablesMS(Class @class, IList<VTableComponent> wrappedEntries)
         {
             var hasVirtualDtor = wrappedEntries.Any(e => e.Method.IsDestructor);
             if (hasVirtualDtor)
@@ -1311,17 +1309,18 @@ namespace CppSharp.Generators.CSharp
                 WriteStartBraceIndent();
                 for (int tableIndex = 0; tableIndex < @class.Layout.VFTables.Count; tableIndex++)
                     WriteLine("Marshal.FreeHGlobal((System.IntPtr)__ManagedVTablesDtorOnly[{0}]);", tableIndex);
+                WriteLine("__ManagedVTablesDtorOnly = null;");
                 WriteCloseBraceIndent();
             }
             WriteLine("if (__ManagedVTables != null)");
             WriteStartBraceIndent();
             for (int tableIndex = 0; tableIndex < @class.Layout.VFTables.Count; tableIndex++)
                 WriteLine("Marshal.FreeHGlobal((System.IntPtr)__ManagedVTables[{0}]);", tableIndex);
+            WriteLine("__ManagedVTables = null;");
             WriteCloseBraceIndent();
         }
 
-        private void DeAllocateVTablesItanium(Class @class, IList<VTableComponent> wrappedEntries,
-            bool destructorOnly)
+        private void DeAllocateVTablesItanium(Class @class, IList<VTableComponent> wrappedEntries)
         {
             var hasVirtualDtor = wrappedEntries.Any(e => e.Method.IsDestructor);
             if (hasVirtualDtor)
@@ -1329,11 +1328,13 @@ namespace CppSharp.Generators.CSharp
                 WriteLine("if (__ManagedVTablesDtorOnly != null)");
                 WriteStartBraceIndent();
                 WriteLine("Marshal.FreeHGlobal((System.IntPtr)__ManagedVTablesDtorOnly[0]);");
+                WriteLine("__ManagedVTablesDtorOnly = null;");
                 WriteCloseBraceIndent();
             }
             WriteLine("if (__ManagedVTables != null)");
             WriteStartBraceIndent();
             WriteLine("Marshal.FreeHGlobal((System.IntPtr)__ManagedVTables[0]);");
+            WriteLine("__ManagedVTables = null;");
             WriteCloseBraceIndent();
         }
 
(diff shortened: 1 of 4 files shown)


src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -1273,10 +1273,51 @@ public void GenerateVTable(Class @class)
...
 
+            GenerateVTableCleanup(@class, wrappedEntries);
+
             WriteLine("#endregion");
             PopBlock(NewLineKind.BeforeNextBlock);
         }
 
+        private void GenerateVTableCleanup(Class @class, IList<VTableComponent> wrappedEntries)
+        {
+            WriteLine("private static readonly Destructor Finalise = new Destructor();");
```
ddobrev: This is a private local variable, please rename it to start with a small letter.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -1273,10 +1273,51 @@ public void GenerateVTable(Class @class)
...
+            if (Options.IsMicrosoftAbi)
+                DeAllocateVTablesMS(@class, wrappedEntries, destructorOnly);
+            else
+                DeAllocateVTablesItanium(@class, wrappedEntries, destructorOnly);
+        }
+
+        private void DeAllocateVTablesMS(Class @class, IList<VTableComponent> wrappedEntries,
+            bool destructorOnly)
+        {
+            var managedVTables = destructorOnly ? "__ManagedVTablesDtorOnly" : "__ManagedVTables";
```
ddobrev: The logic here is not the same as the patching logic. In here both managed tables might be full so you need to check each one and if not null, destroy it - not one or the other as it is at the moment.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -1273,10 +1273,51 @@ public void GenerateVTable(Class @class)
...
+            for (int tableIndex = 0; tableIndex < @class.Layout.VFTables.Count; tableIndex++)
+                WriteLine("Marshal.FreeHGlobal((System.IntPtr){0}[{1}]);", managedVTables, tableIndex);
+        }
+
+        private void DeAllocateVTablesItanium(Class @class, IList<VTableComponent> wrappedEntries,
+            bool destructorOnly)
+        {
+            var managedVTables = destructorOnly ? "__ManagedVTablesDtorOnly" : "__ManagedVTables";
```
ddobrev: Same as the MS case.

ddobrev:
@genuinelucifer Thank you for your work. I have left a few notes. Also, please note that the OS X crashes because of incorrect freeing - https://travis-ci.org/mono/CppSharp/jobs/94922301:

mono(4017,0xb0183000) malloc: *** error for object 0x7a65b9a8: pointer being freed was not allocated

*** set a breakpoint in malloc_error_break to debug

Stacktrace:

  at <unknown> <0xffffffff>

  at (wrapper managed-to-native) System.Runtime.InteropServices.Marshal.FreeHGlobal (intptr) <0xffffffff>

  at NamespacesDerived.HasVirtualInDependency/Destructor.Finalize () <0x0002b>

  at (wrapper runtime-invoke) object.runtime_invoke_virtual_void__this__ (object,intptr,intptr,intptr) <0xffffffff>

This will most probably go away when you've fixed the bottommost two of my remarks but I decided to paste it here anyway so that you have it in mind.

genuinelucifer:
@ddobrev I have tried to incorporate the changes and added a test.
I put a breakpoint on the line ```if (__ManagedVTables != null)``` in ```Bar.Destructor``` and it got executed. But, the code always seems to find that __ManagedVTables is null because the body of the 'if' is not executed. Any ideas as to why this would be so?

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -1273,10 +1273,68 @@ public void GenerateVTable(Class @class)
...
+            if (Options.IsMicrosoftAbi)
+                DeAllocateVTablesMS(@class, wrappedEntries, destructorOnly);
+            else
+                DeAllocateVTablesItanium(@class, wrappedEntries, destructorOnly);
+        }
+
+        private void DeAllocateVTablesMS(Class @class, IList<VTableComponent> wrappedEntries,
+            bool destructorOnly)
+        {
+            if (destructorOnly)
```
ddobrev: I may have not made myself clear: both tables have to always be checked and freed if necessary. So you don't have to check this parameter.

ddobrev:
@genuinelucifer  __ManagedVTables is only filled for user types, that is, not wrapped types but rather types created in managed code by the user which derive from wrapped types. You can find some in the tests.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
genuinelucifer: Actually ```__ManagedVTablesDtorOnly``` is not defined in some cases. That's why I did this. But, now I've changed it.

genuinelucifer:
@ddobrev It finally works. I tested the test in debug mode and the FreeHGlobal call was done exactly once.

src/Generator/Generators/CSharp/CSharpTextTemplate.cs
```diff
@@ -1273,10 +1273,70 @@ public void GenerateVTable(Class @class)
...
+            bool destructorOnly)
+        {
+            if (Options.IsMicrosoftAbi)
+                DeAllocateVTablesMS(@class, wrappedEntries, destructorOnly);
+            else
+                DeAllocateVTablesItanium(@class, wrappedEntries, destructorOnly);
+        }
+
+        private void DeAllocateVTablesMS(Class @class, IList<VTableComponent> wrappedEntries,
+            bool destructorOnly)
```
ddobrev: This parameter is no longer used, please delete it.

ddobrev:
The OS X build still crashes because of freeing. I'm looking at the error, I'll try helping you figure it out.

ddobrev:
I think you'd better debug the particular class that causes the crash. https://travis-ci.org/mono/CppSharp/jobs/95074638 says it's CSharp.Bar:

  at <unknown> <0xffffffff>

  at (wrapper managed-to-native) System.Runtime.InteropServices.Marshal.FreeHGlobal (intptr) <0xffffffff>

  at CSharp.Bar/Destructor.Finalize () <0x0002b>

  at (wrapper runtime-invoke) object.runtime_invoke_virtual_void__this__ (object,intptr,intptr,intptr) <0xffffffff>

.NET sometimes happens to be tolerant when it shouldn't and this could be the reason it works for you locally - that is, if you have run all tests before committing. Having said that, please make sure all test projects are regenerated before you run the tests - the pre-build events we have do not always work. For this particular case, open CSharp.cs and check if you have your finalisers - if you don't, the code has not been regenerated after your changes.

genuinelucifer:
@ddobrev I tried a few things but i am unable to get any error on my machine. And I am unable to figure out why there would be such error on the build bot.
By the way, my last commit did not build because of some error in ```CppParser``` I don't think it could happen because of my code. Please look into it.

ddobrev:
@genuinelucifer do you mean it didn't build locally?

tritao:
Probably because you are trying to build with an old version of Clang. We recently updated the Clang version that we depend on.

You can download `https://dl.dropboxusercontent.com/u/194502/cppsharp/llvm/llvm-f79c5c-windows-RelWithDebInfo.7z`, extract it inside `build/scripts` and re-generate your VS project files by invoking Premake.

genuinelucifer:
@joao Thanks. I'll try to build with that.
.
@ddobrev I meant that it is building fine locally. But, on the build-bot it is showing some error which, i don't think, is related to my commits. Please check that. Part of the error:
```
Unhandled Exception:
System.InvalidOperationException: Managed instances owned by native code cannot be disposed of.
   at CppSharp.Parser.ParserResult.Dispose (Boolean disposing) in /Users/travis/build/mono/CppSharp/src/CppParser/Bindings/CSharp/i686-apple-darwin12.4.0/CppParser.cs:line 953
   at CppSharp.Parser.ParserResult.Dispose () in /Users/travis/build/mono/CppSharp/src/CppParser/Bindings/CSharp/i686-apple-darwin12.4.0/CppParser.cs:line 941
   at CppSharp.ClangParser.ParseSourceFile (CppSharp.SourceFile file) in /Users/travis/build/mono/CppSharp/src/Core/Parser/Parser.cs:line 59
   at CppSharp.ClangParser.ParseProject (CppSharp.Project project) in /Users/travis/build/mono/CppSharp/src/Core/Parser/Parser.cs:line 71
   at CppSharp.Driver.ParseCode () in /Users/travis/build/mono/CppSharp/src/Generator/Driver.cs:line 204
   at CppSharp.ConsoleDriver.Run (ILibrary library) in /Users/travis/build/mono/CppSharp/src/Generator/Driver.cs:line 472
   at CppSharp.Tests.CommonTestsGenerator.Main (System.String[] args) in /Users/travis/build/mono/CppSharp/tests/Common/Common.cs:line 83
[ERROR] FATAL UNHANDLED EXCEPTION: System.InvalidOperationException: Managed instances owned by native code cannot be disposed of.
   at CppSharp.Parser.ParserResult.Dispose (Boolean disposing) in /Users/travis/build/mono/CppSharp/src/CppParser/Bindings/CSharp/i686-apple-darwin12.4.0/CppParser.cs:line 953
   at CppSharp.Parser.ParserResult.Dispose () in /Users/travis/build/mono/CppSharp/src/CppParser/Bindings/CSharp/i686-apple-darwin12.4.0/CppParser.cs:line 941
   at CppSharp.ClangParser.ParseSourceFile (CppSharp.SourceFile file) in /Users/travis/build/mono/CppSharp/src/Core/Parser/Parser.cs:line 59
   at CppSharp.ClangParser.ParseProject (CppSharp.Project project) in /Users/travis/build/mono/CppSharp/src/Core/Parser/Parser.cs:line 71
   at CppSharp.Driver.ParseCode () in /Users/travis/build/mono/CppSharp/src/Generator/Driver.cs:line 204
   at CppSharp.ConsoleDriver.Run (ILibrary library) in /Users/travis/build/mono/CppSharp/src/Generator/Driver.cs:line 472
   at CppSharp.Tests.CommonTestsGenerator.Main (System.String[] args) in /Users/travis/build/mono/CppSharp/tests/Common/Common.cs:line 83
make[1]: *** [prebuild] Error 1
make: *** [Common.CSharp] Error 2
The command "config=$BUILD_CONF make -C build/gmake/" exited with 2.
```

ddobrev:
@genuinelucifer you are right, it isn't. I've already fixed that problem.
About the problem, what I meant was that you don't get the error because .NET is sometimes tolerant when it shouldn't. So what I ask of you is to debug CSharp.Bar and track if it is allocated before being freed at the end. Since it doesn't crash, you'll have to manually check that.

genuinelucifer:
Yes i did debug each call. And yes SetupVTables was called and the pointer
was allocated with AllocHGlobal and deallocated with FreeHGlobal exactly
once. I'll rebase the patch onto updated master branch to check if the
error persists on the build bot.

PR #597, opened 2015-12-16
Fixed missing namespace problem.

### LINKED ISSUE #590: Missed namespace
Minimal steps to reproduce is specified below.
C++ code:
```
namespace NamespaceA
{
	CS_VALUE_TYPE class DLL_API A
	{
	};
}
namespace NamespaceB
{
	class DLL_API B
	{
	public:
		void Function(CS_OUT NamespaceA::A &a);
	};
}
```
Fragment of generated C# code:
```
        public void Function(out NamespaceA.A a)
        {
            a = new NamespaceA.A();
            var __arg0 = new A.Internal();
            var arg0 = new global::System.IntPtr(&__arg0);
            Internal.Function_0((__Instance + __PointerAdjustment), arg0);
            NamespaceA.A __result0;
            if (arg0 == IntPtr.Zero) __result0 = new NamespaceA.A();
            else __result0 = NamespaceA.A.__CreateInstance(arg0);
            a = __result0;
        }
```
Second function line should be:
var __arg0 = new **NamespaceA.**A.Internal();
Generated code is not compilable due to missed namespace.
This problem can be reproduced only if **A** is struct in C#.

genuinelucifer:
closes #590

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpMarshal.cs
+++ b/src/Generator/Generators/CSharp/CSharpMarshal.cs
@@ -501,8 +501,9 @@ namespace CppSharp.Generators.CSharp
             {
                 if (Context.Parameter.Usage == ParameterUsage.Out)
                 {
+                    var qualifiedIdentifier = (@class.OriginalClass ?? @class).Visit(typePrinter);
                     Context.SupportBefore.WriteLine("var {0} = new {1}.Internal();",
-                        Generator.GeneratedIdentifier(Context.ArgName), @class.Name);
+                        Generator.GeneratedIdentifier(Context.ArgName), qualifiedIdentifier);
                 }
                 else
                 {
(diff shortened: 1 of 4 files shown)


src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -501,8 +501,10 @@ public override bool VisitPointerType(PointerType pointer, TypeQualifiers quals)
             {
                 if (Context.Parameter.Usage == ParameterUsage.Out)
                 {
+                    var TypePrinter = new CSharpTypePrinter(Context.Driver);
```
ddobrev: The marshaller already has a type printer contained in a field so please use that instead of creating a new one.

genuinelucifer: @ddobrev Done.

tritao:
Thanks a lot @genuinelucifer!

PR #600, opened 2015-12-30
Fix for fn call when def argument is interface.

genuinelucifer:
This is my initial thought as I found that ```desugered``` is always the class required. I couldn't find a way to figure out if the parameter type is interface. So, probably it will change as currently it generates the cast for every argument whether required or not.

### DIFF
--- a/tests/CSharp/CSharp.h
+++ b/tests/CSharp/CSharp.h
@@ -63,6 +63,7 @@ public:
     };
     Bar();
     Bar(Qux qux);
+    Bar(Items item);
     int method();
     const Foo& operator[](int i) const;
     Foo& operator[](int i);
@@ -364,6 +365,7 @@ public:
     // in this case the arg is a MaterializeTemporaryExpr, in the other not
     // I cannot see the difference but it's there so we need both tests
     void defaultImplicitCtorEnum(Baz arg = Bar::Item1);
+    void defaultImplicitCtorEnumTwo(Bar arg = Bar::Items::Item1);
     void defaultIntWithLongExpression(unsigned int i = DEFAULT_INT);
     void defaultRefTypeEnumImplicitCtor(const QColor &fillColor = Qt::white);
     void rotate4x4Matrix(float angle, float x, float y, float z = 0.0f);
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -2202,9 +2202,11 @@ namespace CppSharp.Generators.CSharp
 
         private string OverloadParamNameWithDefValue(Parameter p, ref int index)
         {
+            Class @class;
             return p.Type.IsPointerToPrimitiveType() && p.Usage == ParameterUsage.InOut && p.HasDefaultValue
                 ? "ref param" + index++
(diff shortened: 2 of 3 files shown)


src/Generator/Passes/HandleDefaultParamValuesPass.cs
```diff
@@ -243,7 +243,7 @@ private static bool CheckFloatSyntax(Type desugared, Statement statement, ref st
             if (enumItem != null)
             {
                 result = string.Format("{0}{1}.{2}",
-                    desugared.IsPrimitiveType() ? "(int) " : string.Empty,
+                    desugared.IsPrimitiveType() ? "(int) " : "(" + desugared.CSharpType(new CSharpTypePrinter(Driver)) + ") ",
```
ddobrev: Please break the line after the colon and use string.Format instead.

tests/CSharp/CSharp.h
```diff
@@ -63,6 +63,7 @@ class DLL_API Bar : public Qux
     };
     Bar();
     Bar(Qux qux);
+	Bar(Items item);
```
ddobrev: Please use 4 spaces instead of a tab.

tests/CSharp/CSharp.h
```diff
@@ -364,6 +365,7 @@ class DLL_API MethodsWithDefaultValues : public Quux
     // in this case the arg is a MaterializeTemporaryExpr, in the other not
     // I cannot see the difference but it's there so we need both tests
     void defaultImplicitCtorEnum(Baz arg = Bar::Item1);
+	void defaultImplicitCtorEnumTwo(Bar arg = Bar::Items::Item1);
```
ddobrev: Please use 4 spaces instead of a tab.

ddobrev:
@genuinelucifer you should be able to call GetFinalPointee() on param.Type or desugared (not sure which one, you'll have to check) and then use TryGetClass (or a similar name, you'll find it easily, it's widely used). Then you can check Class.IsInterface. Just a note: GetFinalPointee() returns null if the type is not a pointer so you have to use type.GetFinalPointee() ?? type. A really stupid bug we need to fix but for the time being this is the way to ensure you don't get a null crash.

genuinelucifer:
Actually I already tried that and a few other things but couldn't get it to work.
I think that it should be checked while parsing the parameters only. I am trying that now.

genuinelucifer:
@ddobrev I guess it's good to go now. Please review.

ddobrev:
@genuinelucifer Happy New Year! I wish you health, happiness and many more commits to C++#. :)

genuinelucifer:
@ddobrev Thanks and same to you :)

ddobrev:
@genuinelucifer thank you. Let me know when you are ready for more work.

genuinelucifer:
I actually don't have much work for the next 1 month or so. So, i am free.
Pls provide the next issue for me to work on.

ddobrev:
Great. :) You've been doing such excellent work that I think you are ready for the next level. There's a relatively simple issue which is very suitable to get you acquainted with the C++ part. https://github.com/mono/CppSharp/issues/601 requires you to add the new field to our AST.h and then read it in the parser and in ASTConverter. Any questions are welcome.

genuinelucifer:
Ohh good. Thanks. I'll try it and ask you if any questions arise.  :)

Issue #508, opened 2015-06-30
Marshal (non-const) char* as StringBuilder 


ddobrev:
There may be some gotchas such as ANSI/Unicode so some reading is needed but this can and should be done.

genuinelucifer:
@ddobrev char* is NOT always meant to work as placeholder for a string. So how can it be differentiated if it is meant to be pointer (hence better a 'ref char') or a string (so 'StringBuilder').
In my opinion, it would however be better to marshal std::string as StringBuilder...

ddobrev:
It's impossible to determine. So we have to use the solution which works in both cases and that is StringBuilder because in the case of a pointer to a char it would simply be a StringBuilder with a single character.
std::string represented with a StringBuilder might or might not be a good idea but in any case it's unrelated to the current issue. Also it's quite difficult because std::string is a typedef to a template type and we don't have proper support for templates yet. So if you want to develop your idea, please file another issue.

PR #604, opened 2016-01-11
First try at moving isimplicit to declaration from method.

genuinelucifer:
I have yet to add a test.
.
Please see if I am going in right direction or not.
Also, i was a bit confused by the 2 fields ```IsImplicit``` and ```IsExplicit```. I think they should be opposite of each other but they are implemented separately. Also I could not find a WalkRecord which fills the IsExplicit field.


ddobrev:
@genuinelucifer I am discussing about IsExplicit/IsImplicit with @tritao . In the mean time, please convert your tabs to 4 spaces and add () to D->isImplicit in Parser.cpp. The latter is not a field but a function, I am pretty sure your code does not compile at the moment. You can also start with the test. We have a special test project for checking the obtained AST, CppSharp.Generator.Tests. You can see how it works in the existing tests.

src/Generator.Tests/AST/TestAST.cs
```diff
@@ -281,5 +281,12 @@ public void TestMacroLineNumber()
         {
             Assert.AreEqual(98, AstContext.FindClass("HasAmbiguousFunctions").First().Specifiers.Last().LineNumberStart);
         }
+
+        [Test]
+        public void TestImplicitDeclaration()
+        {
+            Assert.AreEqual(false, AstContext.FindFunction("testExplicitFunc").Single().IsImplicit);
+            Assert.AreEqual(true, AstContext.FindFunction("testImplicitFunc").Single().IsImplicit);
```
genuinelucifer: This test is failing. Am I missing something about Implicit and Explicit declarations?

src/CppParser/AST.h
```diff
@@ -642,9 +643,8 @@ class CS_API Method : public Function
     bool IsVirtual;
     bool IsStatic;
     bool IsConst;
-    bool IsImplicit;
-    bool IsExplicit;
     bool IsOverride;
+    bool IsExplicit;
```
tritao: These changes can be reverted.

src/Generator.Tests/AST/TestAST.cs
```diff
@@ -281,5 +281,12 @@ public void TestMacroLineNumber()
         {
             Assert.AreEqual(98, AstContext.FindClass("HasAmbiguousFunctions").First().Specifiers.Last().LineNumberStart);
         }
+
+        [Test]
+        public void TestImplicitDeclaration()
+        {
+            Assert.AreEqual(false, AstContext.FindFunction("testExplicitFunc").Single().IsImplicit);
```
ddobrev: Please just ignore IsExplicit. It works correctly at the moment and does not need any related changes.

tests/Native/AST.h
```diff
@@ -106,3 +106,13 @@ class Atomics
 # endif
 #endif
 };
+
```
ddobrev: I should explain what implicit declarations are. These are declarations which are not written at all but are instead generated by the compiler. The most common examples are the empty constructor, copy constructor, assigment operator and destructor in C++ classes. So in order to test implicity, just select any class without any of these members explicitly defined in code, and assert that IsImplicit is true. For example, let's say you have this class:

class ImplicitCtor
{
};

Then you can do Assert.IsTrue(classes.Find("ImplicitCtor").Constructors.First(c => c.Parameter.Count ==0).IsImplicit);

genuinelucifer: Thanks. I don't know why but I remembered from somewhere that functions have implicit and explicit declarations (not necessarily in class).
Well, I have fixed it now as you told. Please review.

src/Generator.Tests/AST/TestAST.cs
```diff
@@ -281,5 +281,11 @@ public void TestMacroLineNumber()
         {
             Assert.AreEqual(98, AstContext.FindClass("HasAmbiguousFunctions").First().Specifiers.Last().LineNumberStart);
         }
+
+        [Test]
+        public void TestImplicitDeclaration()
+        {
+            Assert.IsTrue(AstContext.FindClass("ImplicitCtor").First().Constructors.First(c => c.Parameters.Count == 0).IsImplicit);
```
ddobrev: Please break this line - after "First(" is a good location - and I'll merge.

ddobrev:
@genuinelucifer thank you for your work. Let me know when you are ready for a new task.

genuinelucifer:
@ddobrev Pls provide me a new task to work on.

ddobrev:
@genuinelucifer since you are more experienced now compared to 3 or 4 months ago, would you liek to try the one for StringBuilder (https://github.com/mono/CppSharp/issues/508) again?

genuinelucifer:
@ddobrev I'll definitely try that one again.  I guess their will be a few
doubts in that issue. I'll try first and then ask you about those.

PR #605, opened 2016-01-12
Allow ResolveSymbol to find the library under the system Path

grbd:
Due to an issue with QtSharp and QMetaObject I've made a change to ResolveSymbol to allow it to search custom paths other than the current directory or the system path

Resharper decided to reformat the code a bit so thats the reason for the change in spacing on some of the lines.


grbd:
Closing this for further testing

PR #608, opened 2016-01-12
Added the ability to search the machine Path for ResolveSymbol

grbd:
This change allows ResolveSymbol to find a library by
1. the current directory or a specified directory
2. the directories within the system path

QtSharp uses ResolveSymbol when getting the QMetaObject of a class, however without this change it's unable to find the Qt dll's unless they're in the current directory. With this change as long as the Qt libs are in the system path ResolveSymbol is able to find them.

### DIFF
--- a/src/Runtime/SymbolResolver.cs
+++ b/src/Runtime/SymbolResolver.cs
@@ -57,9 +57,33 @@ namespace CppSharp
 
         public static IntPtr LoadImage (ref string name)
         {
+            var pathvalues = Environment.GetEnvironmentVariable("PATH");
+
             foreach (var format in formats)
             {
-                var attempted = System.IO.Path.Combine(Environment.CurrentDirectory, string.Format (format, name));
+                // Search the Current or specified directory for the library
+                string filename = string.Format(format, name);
+                string attempted = System.IO.Path.Combine(Environment.CurrentDirectory, filename);
+                if (!System.IO.File.Exists(attempted))
+                {
+                    // Search the Path directories for the library
+                    if (pathvalues == null)
+                        continue;
+
+                    foreach (var path in pathvalues.Split(System.IO.Path.PathSeparator))
+                    {
+                        var fullPath = System.IO.Path.Combine(path, filename);
+                        if (System.IO.File.Exists(fullPath))
+                        {
+                            attempted = fullPath;
+                            break;
+                        }
+
+                    }
+                }
+                if (!System.IO.File.Exists(attempted))
+                    continue;
+
                 var ptr = loadImage (attempted);
 
                 if (ptr == IntPtr.Zero)


tritao:
Looks mostly OK, but I'd like to think more about it tomorrow and discuss it with Dimitar. It might make sense to add the concept of search paths directly to the resolver.

ddobrev:
@grbd our discussion is over. We like your pull and we would only like you to fix some minor issues before merging. I have added you a review.

src/Runtime/SymbolResolver.cs
```diff
@@ -72,9 +97,9 @@ public static IntPtr LoadImage (ref string name)
             return IntPtr.Zero;
         }
 
-        public static IntPtr ResolveSymbol (string name, string symbol)
+        public static IntPtr ResolveSymbol (string name, string symbol, string searchdir = null)
```
ddobrev: Why do we need this parameter given that this method is only called by generated code?

src/Runtime/SymbolResolver.cs
```diff
@@ -55,11 +55,36 @@ static SymbolResolver ()
...
+            if (searchdir == null)
+                searchdir = Environment.CurrentDirectory;
+
             foreach (var format in formats)
             {
-                var attempted = System.IO.Path.Combine(Environment.CurrentDirectory, string.Format (format, name));
+                // Search the Current or specified directory for the library
+                string filename = string.Format(format, name);
+                string attempted = System.IO.Path.Combine(searchdir, filename);
+                if (System.IO.File.Exists(attempted) == false)
```
ddobrev: Please use ! instead of comparing to false.

src/Runtime/SymbolResolver.cs
```diff
@@ -55,11 +55,36 @@ static SymbolResolver ()
...
+                {
+                    // Search the Path directories for the library
+                    var pathvalues = Environment.GetEnvironmentVariable("PATH");
```
ddobrev: Please move this outside the loop since it's always the same - this way there are many redundant calls.

src/Runtime/SymbolResolver.cs
```diff
@@ -55,11 +55,36 @@ static SymbolResolver ()
             }
         }
 
-        public static IntPtr LoadImage (ref string name)
+        public static IntPtr LoadImage (ref string name, string searchdir = null)
         {
+            if (searchdir == null)
```
ddobrev: As with my remark about ResolveSymbol, please convert this to a local variable in case we don't need the parameter.

grbd:
My thoughts were that you might want to manually specify a directory not in the system path, but I'll look into making the changes mentioned above

ddobrev:
It makes sense, just not in the context of generated code. This parameter would only serve a purpose if somebody would manually call ResolveSymbol which is not the point of automation. Even if we were to later on add the option of additional paths, they would not be passed through a parameter but rather through a static property.

grbd:
Okay the commit should be amended now

src/Runtime/SymbolResolver.cs
```diff
@@ -57,9 +57,32 @@ static SymbolResolver ()
...
         {
+            var pathvalues = Environment.GetEnvironmentVariable("PATH");
+
             foreach (var format in formats)
             {
-                var attempted = System.IO.Path.Combine(Environment.CurrentDirectory, string.Format (format, name));
+                // Search the Current or specified directory for the library
+                string filename = string.Format(format, name);
+                string attempted = System.IO.Path.Combine(Environment.CurrentDirectory, filename);
+                if (System.IO.File.Exists(attempted) != true)
```
tritao: This could be simplified to `if (!System.IO.File.Exists(attempted))`.

src/Runtime/SymbolResolver.cs
```diff
@@ -57,9 +57,32 @@ static SymbolResolver ()
...
+                        var fullPath = System.IO.Path.Combine(path, filename);
+                        if (System.IO.File.Exists(fullPath))
+                        {
+                            attempted = fullPath;
+                            break;
+                        }
+
+                    }
+                }
+                if (System.IO.File.Exists(attempted) == false) continue;
```
tritao: Same here, `if (!System.IO.File.Exists(attempted)) continue;`.

grbd:
more amends

src/Runtime/SymbolResolver.cs
```diff
@@ -57,9 +57,33 @@ static SymbolResolver ()
...
+
+                    foreach (var path in pathvalues.Split(System.IO.Path.PathSeparator))
+                    {
+                        var fullPath = System.IO.Path.Combine(path, filename);
+                        if (System.IO.File.Exists(fullPath))
+                        {
+                            attempted = fullPath;
+                            break;
+                        }
+
```
ddobrev: Please remove this line.

src/Runtime/SymbolResolver.cs
```diff
@@ -57,9 +57,33 @@ static SymbolResolver ()
...
+                if (!System.IO.File.Exists(attempted))
+                {
+                    // Search the Path directories for the library
+                    if (pathvalues == null)
+                        continue;
+
+                    foreach (var path in pathvalues.Split(System.IO.Path.PathSeparator))
+                    {
+                        var fullPath = System.IO.Path.Combine(path, filename);
+                        if (System.IO.File.Exists(fullPath))
```
ddobrev: You check for existence here, then break, and at line 84 you check for existence again. Isn't the second check redundant?

grbd:
first check current directory
second check path directories
third if nothing found at all, call continue to avoid running loadImage

the first one would only be redundant if the current directory was in the path

Issue #435, opened 2015-04-09
Header included more than once breaks method names.


jamesford42:
ClassA.h defines ClassA and includes ClassY.h
ClassB.h defines ClassB and includes ClassY.h
ClassA, ClassB, and ClassY are all marked for export.

The C# class generated for ClassY ends up appending '1' to all its method names. Eg, ClassY::Init() in C++ becomes ClassY.Init1() in C#.

Note that ClassY.h is protected by header guards, since it needs to be included in more than one other header file. I assume this situation is not dealt with well by CppSharp and/or may be related to the problem.

tomspilman:
This is still a problem almost a year later.  This causes all sorts of confusion when using CppSharp.

genuinelucifer:
@ddobrev I created the 3 headers in the same folder as the CSharp test project files. But, I could not generate their C# files. I also tried to add them to premake4.lua in the CSharp folder by changing
```SetupTestCSharp("CSharp")```
to
```SetupTestCSharp("CSharp", {"ClassA", "ClassB", "ClassY"})```
but it didn't work.
How do I add them to any test to generate the output c# files?

**Comment by ddobrev** [comment]
Where exactly are your new headers located? If you place them in CppSharp/tests/CSharp, they should appear as soon as you have regenerated your C++# solution.

**Comment by tritao** [comment]
Hey guys (cc @jamesford42 @tomspilman)

We'd like to get this fixed but @genuinelucifer is having trouble reproducing the issue.

Can you look at https://github.com/mono/CppSharp/pull/620/files and see what we're missing?

**Comment by tomspilman** [comment]
@tritao - Maybe using `#pragma once` instead of header guards?

**Comment by tritao** [comment]
I can't see it making any difference, it should be equivalent, but give it a shot @genuinelucifer.

**Comment by genuinelucifer** [comment]
Tried, didn't work. @tritao @tomspilman

**Comment by tomspilman** [comment]
@genuinelucifer - Maybe it is in how we normally include the files in the generation process:
```
            options.Headers.Add(@"ClassA.h ");
            options.Headers.Add(@"ClassB.h ");
            options.Headers.Add(@"ClassY.h");
```

genuinelucifer:
Umm. In tests the exact same thing is done via ```GeneratorTest.cs```
So, probably won't make much difference. I'll to test anyway.

ddobrev:
@tomspilman could you perhaps gradually decrease your real code to a test case?

[...]

PR #612, opened 2016-01-26
Marshal all char ptrs to StringBuilder

genuinelucifer:
Build is failing currently. I'm trying to fix it.

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
+++ b/src/Generator/Generators/CSharp/CSharpTextTemplate.cs
@@ -571,9 +571,9 @@ namespace CppSharp.Generators.CSharp
         {
             var isNonConstParam = IsNonConstCharPtrParam(param);
             if (isNonConstParam && param.Type.IsPointerToPrimitiveType(PrimitiveType.Char))
-                return string.Format("[MarshalAs(UnmanagedType.LPStr)]{0} {1}", "StringBuilder", param.Name);
+                return string.Format("[MarshalAs(UnmanagedType.AnsiBStr)]{0} {1}", "String", param.Name);
             else if (isNonConstParam && param.Type.IsPointerToPrimitiveType(PrimitiveType.WideChar))
-                return string.Format("[MarshalAs(UnmanagedType.LPWStr)]{0} {1}", "StringBuilder", param.Name);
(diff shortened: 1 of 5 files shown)


ddobrev:
Great to see you working again, I am sure you'll overcome it this time.

genuinelucifer:
Thanks. I hope too that this time i would be able to complete it.

genuinelucifer:
@ddobrev It builds correctly on my windows pc and tests do run too. I don't know what's going wrong on the Mac buildbot!
I'm looking into it.
Meanwhile, I was trying to marshal the constant sized value type arrays too but it seems then we can't take pointer to ```Internal``` struct if it has any managed type and fixing that seemed to be quite large task. I couldn't think of any other solution so I dropped that idea and left my temporary solutions as permanent ones. If it is acceptable to change the types by comparing and checking equality to ```StringBuilder``` at the 2 places then I'll leave it else I will try to do something else.

ddobrev:
Thank you for your work. if you have problems with the arrays, please leave them for now, pointers are fine as a start. About the crash on Unix, searching yields quite a few results (https://www.google.bg/search?q=Structure+field+of+type+Char[]+can%27t+be+marshalled+as+LPArray&ie=utf-8&oe=utf-8&gws_rd=cr&ei=OsmsVuuqJIqasAGJyLTIBA#q=%22Structure+field+of+type%22+%22can%27t+be+marshalled+as+LPArray%22), it might be helpful. I'll discuss it with @tritao a little later.

genuinelucifer:
I tried those things already. Actually those define for Marshalling fields
which i haven't incorporated yet.

ddobrev:
@tritao do you have any clue about the crash at the Unix builds?

genuinelucifer:
@ddobrev Although i'll try to resolve the issues in this pull. But, please give me some other task to work  on untill we find a solution for this.

ddobrev:
@genuinelucifer my apologies, both @tritao and I have been really busy lately. Another issue you could try is https://github.com/mono/CppSharp/issues/435 .

tritao:
After careful deliberation and discussion between me and @ddobrev, I have decided that this approach is not the best design for this feature since it will not work in a lot of cases we would like to support. We'll eventually come up with a better proposal for the new design.

PR #620, opened 2016-02-18
Just checking the issue

genuinelucifer:
I can't reproduce #435 
What am I doing wrong?

### DIFF
--- a/tests/CSharp/ClassA.h
+++ b/tests/CSharp/ClassA.h
@@ -1,5 +1,5 @@
 #include "ClassY.h"
 
-class ClassA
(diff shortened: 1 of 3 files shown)


genuinelucifer:
The function ```init()``` is named ```Init()``` and NO 1 is appended when I generated the files.

ddobrev:
@genuinelucifer the only difference I can see is that the issue says all three classes are exported (DLL_API in our tests), however, I don't think it would affect the generated code.

genuinelucifer:
@ddobrev tried, didn't work.

Issue #615, opened 2016-02-03
fail to generate cs file for function with union as argument


chunseoklee:
```cpp
#include <stdio.h>
typedef union {
  int c;
  int d[10];
} union_t;
struct test_struct2                                                            
  int b;
  union_t u;
};
int func_union(union_t u) {
}

int func(struct test_struct2 *p) {
}
```
The code above failed to generate cs file for ```func_union```.  
I think CppSharp omitted functions that get union structure as an argument.
Could someone help me to resolve this issue ? or this feature is not implemented yet ?

tritao:
I took a quick look and this is because we ignore the typedef declaration, the relevant code is https://github.com/mono/CppSharp/blob/master/src%2FGenerator%2FPasses%2FCleanInvalidDeclNamesPass.cs#L118.

If I remove that whole method, your example code works. @ddobrev do you know if we can safely get rid of this code?

PR #621, opened 2016-02-18
Fixing generation of functions with union params

genuinelucifer:
Fix for #615 as suggested by @tritao

### DIFF
--- a/src/Generator/Passes/CleanInvalidDeclNamesPass.cs
+++ b/src/Generator/Passes/CleanInvalidDeclNamesPass.cs
@@ -108,25 +108,6 @@ namespace CppSharp.Passes
             return ret;
         }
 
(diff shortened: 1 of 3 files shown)


ddobrev:
@genuinelucifer thank you. Could you please try decreasing the test case? The bug is that we used to ignore certain typedefs and much of the test code, while precisely taken from the user, is actually irrelevant. I think that:

    struct test_struct2                                                            
      int b;
      union_t u;
    };

and:

    int func(struct test_struct2 *p) {
    }

are redundant.

tritao:
Additionally, we could probably move this test to Common, since it does not seem C# specific.

genuinelucifer:
Closing since it has merge conflicts.

PR #635, opened 2016-03-06
Fix generation of functions with union params

### LINKED ISSUE #615: fail to generate cs file for function with union as argument

genuinelucifer:
Fixes #615

### DIFF
--- a/tests/Common/Common.h
+++ b/tests/Common/Common.h
@@ -1072,3 +1072,10 @@ public:
     DerivedFromTemplateInstantiationWithVirtual();
 };
 
+typedef DLL_API union {
+    int c;
+} union_t;
+
+int DLL_API func_union(union_t u) {
+    return u.c;
+}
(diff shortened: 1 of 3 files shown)


tests/Common/Common.h
```diff
@@ -1072,3 +1072,10 @@ class DLL_API DerivedFromTemplateInstantiationWithVirtual : public TemplateWithV
     DerivedFromTemplateInstantiationWithVirtual();
 };
 
+typedef DLL_API union {
+	int c;
```
ddobrev: Please just use 4 spaces instead of tabs here and below and I'll merge.

genuinelucifer:
@ddobrev Done.

PR #661, opened 2016-06-22
Added marshalling of std::string in the C# end

ddobrev:


src/CppParser/Parser.cpp
```diff
@@ -63,6 +63,25 @@ using namespace CppSharp::CppParser;
 // We use this as a placeholder for pointer values that should be ignored.
 void* IgnorePtr = (void*) 0x1;
 
+bool IsNamespaceStd(const clang::DeclContext* Namespace)
```
tritao: This can be made `static`.

src/CppParser/Parser.cpp
```diff
@@ -63,6 +63,25 @@ using namespace CppSharp::CppParser;
...
+{
+    if (!Namespace)
```
tritao: Is there any possibility to `Namespace` actually being null? If not I'd rather put an `assert` here.

ddobrev: Yes, if we reach the translation unit and then go up.

src/CppParser/Parser.cpp
ddobrev: I got confused here, I thought it already was because it's not registered in the header, something like a C function. I'll change it.

PR #681, opened 2016-07-26
Changes for VS2015 and Clang 3.8.1

loofbon:
Basic changes to compile under VS2015 (Update 3) and link with Clang 3.8.1

### DIFF
--- a/build/LLVM.lua
+++ b/build/LLVM.lua
@@ -74,6 +74,9 @@ function SetupLLVMLibs()
 
   configuration "macosx"
     links { "c++", "curses", "pthread", "z" }
+    
+  configuration "vs*"
+    links { "version" }
 
   configuration {}
 
--- a/src/CppParser/Parser.cpp
+++ b/src/CppParser/Parser.cpp
@@ -2256,8 +2256,8 @@ Type* Parser::WalkType(clang::QualType QualType, clang::TypeLoc* TL,
             TSTL = &TSpecTL;
         }
 
-        TemplateArgumentList TArgs(TemplateArgumentList::OnStack, TS->getArgs(),
-            TS->getNumArgs());
+        ArrayRef<clang::TemplateArgument> TSArgs(TS->getArgs(), TS->getNumArgs());
+        TemplateArgumentList TArgs(TemplateArgumentList::OnStack, TSArgs);
         TST->Arguments = WalkTemplateArgumentList(&TArgs, TSTL);
 
         Ty = TST;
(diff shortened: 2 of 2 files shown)


src/CppParser/Parser.cpp
```diff
@@ -3638,7 +3638,7 @@ ParserResult* Parser::ParseHeader(const std::vector<std::string>& SourceFiles, P
     std::unique_ptr<llvm::Module> M(new llvm::Module("", Ctx));
 
     M->setTargetTriple(AST->getTargetInfo().getTriple().getTriple());
-    M->setDataLayout(AST->getTargetInfo().getDataLayoutString());
+    M->setDataLayout(AST->getTargetInfo().getDataLayout().getStringRepresentation());
```
ddobrev: getDataLayout() can be directly used so let's go for the shorter code.

src/CppParser/Parser.cpp
```diff
@@ -3904,7 +3904,7 @@ ParserTargetInfo* Parser::GetTargetInfo()
...
```
ddobrev: getDataLayout() can be directly used.

src/CppParser/Parser.cpp
```diff
@@ -3992,4 +3992,4 @@ Declaration* Parser::GetDeclarationFromFriend(clang::NamedDecl* FriendDecl)
         Decl->LineNumberEnd = MinLineNumberEnd;
     }
     return Decl;
-}
\ No newline at end of file
+}
```
ddobrev: Please revert this part. I understand the new line's been silently added by the editor but we use a different editor so if we leave this in, it will keep switching back and forth between commits.

src/CppParser/Parser.cpp
```diff
@@ -3832,7 +3832,7 @@ ParserResult* Parser::ParseLibrary(const std::string& File, ParserResult* res)
     }
 
     auto BinaryOrErr = llvm::object::createBinary(FileEntry);
-    if (BinaryOrErr.getError())
+    if (!BinaryOrErr) // false indicates error
```
ddobrev: Please remove the comment, it's clear enough without it and it makes the line even longer compared to if takeError() was used instead.

ddobrev:
@loofbon excellent work. Just a few minor changes and it's good to go.

build/LLVM.lua
```diff
@@ -74,6 +74,9 @@ function SetupLLVMLibs()
 
   configuration "macosx"
     links { "c++", "curses", "pthread", "z" }
+    
+  configuration "vs2015"
+    links { "version" }
```
ddobrev: This is required by Clang, not by VS, so I think it's not just VS 2015 that needs it. @tritao what do you think?

ddobrev:
@loofbon my apologies about forgetting to tell you in advance. When you fix your pulls, please amend your commits so that the history is not filled with small meaningless ones. In this case the changes to the parser should all go in "Update Parser.cpp" and any changes to the premake should go in the other commit so that you end up with two commits in total again. I know how painful it is to have to go back in history and rebase and all that and I am really sorry. I hope you will accept my apology and understand why this is important to the project.

build/LLVM.lua
tritao: I'm not sure exactly about this. I can think of three cases:

1) "version" is a Windows-specific library needed by Clang, so the filter would be "windows"
2) same as above but VS-specific, in which case maybe "vs*" would be more future-proof
3) or as you suggest it's needed by all OS, but I think option 1) or 2) are more likely

tritao:
Great work, let's merge this as soon as we upload the updated Linux and Windows build.

build/LLVM.lua
ddobrev: I meant we most probably need this block for "vs2013" too.

loofbon:
No worries, will clean up the history.

PR #705, opened 2016-10-25
Use the qualified pointee type when printing/visiting pointer types.

tritao:
@ddobrev please review and see if it makes sense to change all these instances to qualified visitation.

The tests seem to pass but I am not sure about making these changes.


ddobrev:
This does look cleaner to me and I can see it breaks no tests so I am all for merging it.

PR #736, opened 2017-01-24
Added more information when errors occur or invalid options are given

zillemarco:


src/CLI/CLI.cs
```diff
@@ -9,45 +9,41 @@ namespace CppSharp
 {
     class CLI
     {
-        private static Options _options = new Options();
+        private static Mono.Options.OptionSet CLIOptions = new Mono.Options.OptionSet();
```
ddobrev: Please rename to "optionSet". Fields should not start with a capital letter and it's also redundant to use a prefix named after the parent class.

src/CLI/CLI.cs
```diff
@@ -89,7 +95,15 @@ static void ShowHelp(Mono.Options.OptionSet options)
             Console.WriteLine("   contain only the bindings for that header file.");
         }
 
-        static void HandleOutputArg(String arg)
+        static void AddIncludeDirs(String dir, ref List<String> messages)
```
ddobrev: List<> is a class and there's no need to pass classes with "ref" because they are passed by reference anyway.
Please remove "ref" in all places where you pass lists, not just here.

zillemarco:
Done, if the code needs other changes please let me know.

ddobrev:
@zillemarco thank you. Just please merge the commits into one and change "informations" to "information" in the commit message.

zillemarco:
Done, sorry it took me so long but I've had exams at the university :)

ddobrev:
@zillemarco thank you.
I would like to talk to you about the formatting of your code now. I would've opened a new issue but I cannot assign it to you. So I am going to describe it here. Could you please open a new pull request where you have:
1. Removed the underscore prefixes from your fields;
2. Used automatic properties wherever possible (Options.cs is the best example);
3. Made list properties read-only and initialised within the class - please use the property initialisers from C# 6.0;
4. Replaced all usages of String with string;
5. Replaced all comparisons with string.Empty with string.IsNullOrEmpty;
6. Replaced all checks of Boolean properties with "if (Boolean/!Boolean)" instead of "if (Boolean == true/false)";
7. Used StringBuilder instead of string concatenation where you dynamically compose strings (for example at https://github.com/mono/CppSharp/blob/master/src/CLI/Generator.cs#L240);
8. Removed any redundant usings - for example, you seem not to need using System.Threading.Tasks; in Options.cs.
The excellent add-in of Refactoring Essentials (http://vsrefactoringessentials.com/) can help you with some of these tasks, especially the last one.

zillemarco:
Of course, no problem.
One question though: is it possible to implement some test cases to test the CLI? Because I've made the changes but I'm running short on test cases and since I still don't know CppSharp that well I'm not sure if my results are correct or not.

ddobrev:
One possible way of testing, which I am not sure if you have implemented, is printing the compilation errors, if any, when the C# generator in combination with the option for auto-compilation is used. It's not really testing, in a way, but the user can see if anything is wrong with the code generated for the binding.

Issue #745, opened 2017-01-30
Wrong program name on CLI help text under Mono


tritao:
##### Brief Description

When running the new CLI under Mono, the wrong program name `mono-sgen32` shows up.

```mono CppSharp.CLI.exe

Usage: mono-sgen32 [OPTIONS]+ [FILES]+
Generates target language bindings to interop with unmanaged code.
```

OS: OS X

tritao:
@zillemarco ˆ

zillemarco:
Yes sure, do I name it CppSharp.CLI like on Windows?

ddobrev:
Fixed by https://github.com/zillemarco/CppSharp/commit/6faa5918844c8a9560fc3a4679416066a4edaddf.

PR #756, opened 2017-02-11
Changed code to fit repository style and fixed wrong CLI application name

zillemarco:
Changed code to fit repository style and fixed wrong CLI application name (see #745)

### DIFF
--- a/src/CLI/Generator.cs
+++ b/src/CLI/Generator.cs
@@ -6,108 +6,111 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using CppAbi = CppSharp.Parser.AST.CppAbi;
 
 namespace CppSharp
 {
     class Generator : ILibrary
     {
-        private Options _options;
-
-        private String _triple = "";
-        private CppAbi _abi = CppAbi.Microsoft;
+        private Options options = null;
+        private string triple = "";
+        private CppAbi abi = CppAbi.Microsoft;
 
         public Generator(Options options)
         {
             if (options == null)
-                throw new ArgumentNullException("options");
+                throw new ArgumentNullException(nameof(options));
 
-            _options = options;
+            this.options = options;
         }
 
-        public bool ValidateOptions(List<String> messages)
+        public bool ValidateOptions(List<string> messages)
         {
-            if (Platform.IsWindows && _options.Platform != TargetPlatform.Windows)
+            if (Platform.IsWindows && options.Platform != TargetPlatform.Windows)
             {
                 messages.Add("Cannot create bindings for a platform other that Windows from a Windows running machine");
                 return false;
             }
-            else if (Platform.IsMacOS && _options.Platform != TargetPlatform.MacOS)
+            else if (Platform.IsMacOS && options.Platform != TargetPlatform.MacOS)
             {
                 messages.Add("Cannot create bindings for a platform other that MacOS from a MacOS running machine");
                 return false;
             }
-            else if (Platform.IsLinux && _options.Platform != TargetPlatform.Linux)
+            else if (Platform.IsLinux && options.Platform != TargetPlatform.Linux)
             {
                 messages.Add("Cannot create bindings for a platform other that Linux from a Linux running machine");
                 return false;
             }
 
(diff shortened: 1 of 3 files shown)


src/CLI/Generator.cs
```diff
@@ -6,155 +6,158 @@
...
 
 namespace CppSharp
 {
     class Generator : ILibrary
     {
-        private Options _options;
-
-        private String _triple = "";
-        private CppAbi _abi = CppAbi.Microsoft;
+        private Options Options = null;
```
ddobrev: Private fields start with a lowercase letter. To make a distinction between a field and a parameter, use "this":

this.options = options;

src/CLI/Options.cs
```diff
@@ -16,52 +11,36 @@ enum TargetArchitecture
...
-        public List<String> Libraries { get { return _libraries; } set { _libraries = value; } }
+        public List<string> Libraries { get; } = new List<string>();
 
-        private Dictionary<String, String> _defines = new Dictionary<String, String>();
-        public Dictionary<String, String> Defines { get { return _defines; } set { _defines = value; } }
+        public Dictionary<string, string> Defines { get; } = new Dictionary<string, string>();
 
-        private String _outputDir = "";
-        public String OutputDir { get { return _outputDir; } set { _outputDir = value; } }
+        public string OutputDir { get; set; } = "";
```
ddobrev: Now that you use string.IsNullOrEmpty, this initialisation is needless, the default value of "null" will work just the same:

public string OutputDir { get; set; };

src/CLI/Options.cs
```diff
@@ -16,52 +11,36 @@ enum TargetArchitecture
...
-        public TargetArchitecture Architecture { get { return _architecture; } set { _architecture = value; } }
+        public TargetArchitecture Architecture { get; set; } = TargetArchitecture.x86;
 
-        private GeneratorKind _kind = GeneratorKind.CSharp;
-        public GeneratorKind Kind { get { return _kind; } set { _kind = value; } }
+        public GeneratorKind Kind { get; set; } = GeneratorKind.CSharp;
         
-        private bool _checkSymbols = false;
-        public bool CheckSymbols { get { return _checkSymbols; } set { _checkSymbols = value; } }
+        public bool CheckSymbols { get; set; } = false;
```
ddobrev: Boolean fields are false by default, no need to initialise.

src/CLI/Generator.cs
```diff
@@ -13,151 +13,151 @@ namespace CppSharp
...
         {
             var parserOptions = driver.ParserOptions;
-            parserOptions.TargetTriple = Triple;
-            parserOptions.Abi = Abi;
+            parserOptions.TargetTriple = triple;
+            parserOptions.Abi = abi;
 
             var options = driver.Options;
-            options.LibraryName = Options.OutputFileName;
+            options.LibraryName = this.options.OutputFileName;
```
ddobrev: We only use "this" when we have to make a distinction, for example in a constructor.

src/CLI/Generator.cs
```diff
@@ -6,155 +6,158 @@
...
         {
             var parserOptions = driver.ParserOptions;
-            parserOptions.TargetTriple = _triple;
-            parserOptions.Abi = _abi;
+            parserOptions.TargetTriple = triple;
+            parserOptions.Abi = abi;
 
             var options = driver.Options;
-            options.LibraryName = _options.OutputFileName;
+            options.LibraryName = this.options.OutputFileName;
```
ddobrev: Some more "this" to remove. To make sure, please search for all usages of the field.

zillemarco:
I've made the changes and squashed them together.

ddobrev:
@zillemarco I am afraid I have to ask you to send your pull request again. gayhump obliterated our history by adding my comment to the commit message instead of the discussion here as it had always done before. I am really sorry about this but I had no way of expecting such a moronic behaviour.
Now that you have to do it anyway, please separate just the change for renaming the application into its own commit. It has no logical connection to your refactoring so for better history it really should have its own record.

zillemarco:
All right, no problem. So I just need to divide the commit into two again and then redo the pull request if I understood correctly?

ddobrev:
Right.

PR #767, opened 2017-02-19
Add FFmpeg.AutoGen

Ruslan-B:
I recently switched my bindings generator from ClangSharp to CppSharp.
As CppSharp is designed to support C++ I opted out the code generation part and used parser output only.
Despite it took a loot of effort - I managed to improve the quality of FFmpeg.AutoGen so much!
Thank you for your existence!

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,4 +112,6 @@ CppSharp is used by the following projects:
 
 [ChakraSharp](https://github.com/baristalabs/ChakraSharp)
 
+[FFmpeg.AutoGen](https://github.com/Ruslan-B/FFmpeg.AutoGen)
+
 Please feel free to send us a pull request adding your own projects to the list above.


ddobrev:
@Ruslan-B thank you for your testimonial and for your kind words.
We would be interested to know why you skipped the code generation. We also support wrapping C, not just C++. Have you perhaps encountered any problems?

tritao:
Hi Ruslan, thanks for reaching out, interestingly this is now the third project doing FFmpeg bindings with CppSharp over the years. 👍 

@ddobrev Our C support has never been great... it can work in some cases but it also misses a lot of patterns used by C libraries, in fact I think we might have some old issues about it.

I took a quick look at FFmpeg.AutoGen and there's some stuff that could potentially be interested to get back in CppSharp, like the macro expression parser.

@Ruslan-B What kind of features would make your life easier when wrapping C code with CppSharp?

It seems to me that we're definitely missing something when it comes to C, because people tend to write their own generators (the previous FFmpeg binding project also went with its own C backend).

Ruslan-B:
Hi there,

Well, CppSharp was the next logical step as for me. As it was clear that is based on C++ API for clang. Thus it makes possible to get macro expression in one piece. In previous version based on ClangSharp it was possibly as well but it was very hacky. 

@ddobrev  It is not hard to form exact reasoning why I decided to write my own code generator.
Initially, I made quite a few attempts to extend CppSharp infrastructure in a "legal way" but it simply wasn't possible without changing CppSharp itself. 
As I'm already had the working generator on hands with a way lower level of complexity it was obvious to continue with it.

@tritao I don't mind help to contribute useful stuff back to CppSharp just let me know if I can help.

Here is the list of problems I tried to address:
- functions exported from different (physical) libraries;
- cross library structures and enumerations;
- C style fixed size array, they can be nested as well;
- huge structures with cross references;
- bit fields in structures;
- pointers to anonymous delegates in structures;
- anonymous enums and unions in structures;
- pointers to structures in delegates;
- marshaling of delegate pointers; 
- incomplete structures;

As an extra ClangSharp parser allowed me to generate xml documentation and obsolete attributes with description.

ddobrev:
@Ruslan-B thank you for your detailed reply.
I believe we have already covered everything you've listed except a few issues with arrays. We generate XML documentation too. @tritao and I would like to discuss with you the possibility of switching entirely to CppSharp, that is, what we can do and what you can help with. How would you like such a discussion?

Ruslan-B:
@ddobrev I do and I wouldn't mind to switching it entirely to CppSharp. Just let me know how you would like further on. As for example, as a starting point, I can recreate setup I had before just to check what exactly went wrong.

ddobrev:
Thank you, this would be great. Once we know what the problems are, we'll see about either fixing them or helping you do so.

PR #769, opened 2017-02-23
Allow indexer to set non-primitive types (that aren't by reference and aren't const)

u255436:
Imagine this code:
~~~
struct MyIndexerStruct { /*...*/ };
struct Test { MyIndexerStruct operator[](int index); };
~~~

Even though the struct is not returned by reference, it is still valid to assign to the indexer, e.g.:
~~~
Test t;
t[0] = MyIndexerStruct(); // valid code
~~~

The current implementation does not generate a set method, thus preventing this valid case.  Of course, if the indexer has a const return or primitive type, it is not valid, e.g.:
~~~
struct Test
{
   const MyIndexerStruct operator[](int index); // can get but not set
   int operator[](int index); // can get but not set
}
~~~

### DIFF
--- a/src/Generator/Passes/CheckOperatorsOverloads.cs
+++ b/src/Generator/Passes/CheckOperatorsOverloads.cs
@@ -108,6 +108,13 @@ namespace CppSharp.Passes
                 if (!qualifiedPointee.Qualifiers.IsConst)
                     property.SetMethod = @operator;
             }
+            // We can set non-primitive types (even if they are not addresses)
+            // as long as they are not const.
+            else if (!returnType.IsPrimitiveType())
+            {
+                if (!@operator.ReturnType.Qualifiers.IsConst)
+                    property.SetMethod = @operator;
+            }
             
             // If we've a setter use the pointee as the type of the property.
             var pointerType = property.Type as PointerType;


ddobrev:
@u255436 thank you very much for your code. Could you just please:
1. Remove the merge commit from your pull;
2. Add your sample code as a test. Please try Common.h first, if you hit a problem in our C++/CLI generator, use CSharp.h instead. Please also add a unit test that actually uses the code. Finally, when sending the changes, amend your commit instead of sending a separate one for the test.

u255436:
I'll create a different pull request following your instructions.

u255436:
I will just open an issue.  I only cared about the C++/CLI side, but apparently it doesn't work so well on the C# side.

ddobrev:
@u255436 while we would love a fix for C# as well, we understand you don't want to make that effort too. In this case you can use CLI.h instead, it only work with the C++/CLI generator.

u255436:
I opened https://github.com/mono/CppSharp/issues/774.  I would be willing to fix it, but I probably won't have time until next week.

ddobrev:
@u255436 I understand, please take your time. We're grateful you're willing to help out.

Issue #763, opened 2017-02-17
Generate bindings with sequential layout instead of fixed


ha11owed:
I want to use the generated C# bindings with both 32bit and 64bit dlls.

The problems is that the generated structures look like this:
[StructLayout(LayoutKind.Explicit, Size = ....)] and the pointers are all 8 bytes aligned,
making the generated structures 64 bit only.

Is there any option of changing the layout type to Sequencial and remove the FieldOffset attributes from the structure fields? This would be enough to make the points be 4 bytes for 32bit dlls.

ddobrev:
@ha11owed we considered unifying 32 and 64-bit bindings long ago. The problem is that having universal internal structures wouldn't be enough. ABI-s differ between 32 and 64 as well which would force each function to have two internal versions and check the current architecture on each call in order to know which internal function to call. This effectively means significantly increasing the size of resulting bindings. Therefore there would be little gain in switching to sequential layouts which we have considered doing by itself too. It is difficult to get right in all cases because of packing and the risk is just not worth it considering that, as I said, it would be unreasonable to merge 32/64 bindings anyway.
Having said that, in case your binding has some specifics which enforce such a requirement, we'd be ahppy to hear about them.

ha11owed:
@ddobrev In my case i only have C exports. This means there is no need to have two different functions (even the name mangling is the same on both of my dlls).

Actually I just did a manual replace of the LayoutKind in the StructureLayout attribute and removed the FieldOffset attributes and got an version that works with both my 32bit and 64bit dll.

I suppose the only issue is that Marshal.AllocHGlobal still uses the fixed size of the 64bit structures instead of Marshal.SizeOf(typeof(myType.__Internal)). But in my particular case this is not an issue.

ddobrev:
@ha11owed I am afraid this is something which cannot be done with passes. However, I'll discuss with @tritao the possibility of switching to sequential layouts after all, this way C bindings would get what you want for free.

ddobrev:
@ha11owed @tritao and I discussed this and we could add it, possibly optional. However, we are both too busy with much other work, both features and bugs. So I would like to ask you to help with this effort, it would actually cost you less than writing our own custom pass. To begin with, could you please send us a pull request replacing the hard-coded number of bytes used by Marshal.AllocHGlobal with Marshal.SizeOf(typeof(myType.__Internal)), as you suggested?

ha11owed:
@ddobrev Ok, I will do the changes and send a pull request

[...]

PR #770, opened 2017-02-24
Let the marshal determine the size of the structure

ha11owed:
related to issue #763


tritao:
Looks good, thanks for contribution.

ddobrev:
@ha11owed thank you for your work. I am afraid, though, that you had made a fatal mistake which @tritao failed to notice. You had used "Marshal.SizeOf(sizeof(MyType))" which resolves to Marshal.SizeOf(int) and always returns 4 or 8. I have fixed it now but you need to use our latest code in order to get working bindings. In addition, please mind https://github.com/mono/CppSharp/issues/771 which means that those of your structures, if any, which happen to use the primitive type listed there, will crash.

PR #794, opened 2017-03-16
Fixed ArgumentException 

realvictorprm:
Which were due to nullptr arguments in Marshaling native.

### DIFF
--- a/src/Generator/Generators/CLI/CLIMarshal.cs
+++ b/src/Generator/Generators/CLI/CLIMarshal.cs
@@ -249,7 +249,7 @@ namespace CppSharp.Generators.CLI
             FunctionType function;
             if (decl.Type.IsPointerTo(out function))
             {
-                Context.Return.Write("safe_cast<{0}>(", typedef);
+                Context.Return.Write($"{Context.ReturnVarName} == nullptr ? nullptr : safe_cast<{typedef}>(");
                 Context.Return.Write("System::Runtime::InteropServices::Marshal::");
                 Context.Return.Write("GetDelegateForFunctionPointer(");
                 Context.Return.Write("IntPtr({0}), {1}::typeid))",Context.ReturnVarName,
--- a/src/Generator/Generators/CSharp/CSharpMarshal.cs
+++ b/src/Generator/Generators/CSharp/CSharpMarshal.cs
@@ -249,9 +249,9 @@ namespace CppSharp.Generators.CSharp
 
                 Context.SupportBefore.WriteLine("var {0} = {1};", ptrName,
                     Context.ReturnVarName);
-
-                Context.Return.Write("({1})Marshal.GetDelegateForFunctionPointer({0}, typeof({1}))",
-                    ptrName, typedef.ToString());
+                
+                var res = $"{ptrName} == IntPtr.Zero? null : ({typedef})Marshal.GetDelegateForFunctionPointer({ptrName}, typeof({typedef}))";
+                Context.Return.Write(res);
                 return true;
             }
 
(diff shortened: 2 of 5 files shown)


src/Generator/Generators/CLI/CLIMarshal.cs
```diff
@@ -249,7 +249,7 @@ public override bool VisitTypedefType(TypedefType typedef, TypeQualifiers quals)
             FunctionType function;
             if (decl.Type.IsPointerTo(out function))
             {
-                Context.Return.Write("safe_cast<{0}>(", typedef);
+                Context.Return.Write("{0} == nullptr ? nullptr : safe_cast<{1}>(", Context.ReturnVarName, typedef);
```
ddobrev: We are in the process of switching to interpolated strings so please use them here. You could use the excellent add-ins of http://vsrefactoringessentials.com/ and https://marketplace.visualstudio.com/items?itemName=DustinCampbell.CEssentials, the latter has a refactoring to turn string.Format (which won't work in this particular case, though) to an interpolated string.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -250,7 +250,7 @@ public override bool VisitTypedefType(TypedefType typedef, TypeQualifiers quals)
                 Context.SupportBefore.WriteLine("var {0} = {1};", ptrName,
                     Context.ReturnVarName);
 
-                Context.Return.Write("({1})Marshal.GetDelegateForFunctionPointer({0}, typeof({1}))",
+                Context.Return.Write("{0} == IntPtr.Zero ? null : ({1})Marshal.GetDelegateForFunctionPointer({0}, typeof({1}))",
```
ddobrev: Please extract the delegate expression to a separate string and use interpolated strings - we avoid multi-lined ones.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -250,7 +250,7 @@ public override bool VisitTypedefType(TypedefType typedef, TypeQualifiers quals)
...
                     ptrName, typedef.ToString());
```
ddobrev: Please remove ToString(), it's not necessary.

tests/Common/Common.Tests.cs
```diff
@@ -317,6 +317,9 @@ public void TestDelegates()
 
         var cdecl = delegates.CDecl(i => i);
         Assert.AreEqual(1, cdecl);
+
+        var emptydelegeate = delegates.MarshalNullDelegate;
+
```
ddobrev: Assert the returned variable is null, you might as well inline the call.

tests/Common/Common.cpp
```diff
@@ -421,6 +421,11 @@ void TestDelegates::MarshalDelegateInAnotherUnit(DelegateInAnotherUnit del)
 {
 }
 
+DelegateNullCheck TestDelegates::MarshalNullDelegate()
+{
+	return nullptr;
```
ddobrev: We use 4 spaces instead of a tab in C++ too.

tests/Common/Common.h
```diff
@@ -345,6 +346,8 @@ struct DLL_API TestDelegates
 
     void MarshalDelegateInAnotherUnit(DelegateInAnotherUnit del);
 
+	DelegateNullCheck MarshalNullDelegate();
```
ddobrev: 4 spaces instead of a tab.

realvictorprm:
thx for info.

PR #795, opened 2017-03-16
Fix CheckStaticClass pass and add test.

### LINKED ISSUE #743: C++ class with only static methods; not appearing as C# static class
It use to be that if I declared a C++ class, and its publically accessible methods/fields were all statics, then it would be exported to C# as an actual static C# class. However, I can no longer get this to occur (using most recent release CppSharp source code).

I have added the following to the bottom of the CSharp.Native.vxproj file CSharp.h

```
class DLL_API TestClassWithOnlyStatics
{
	// private constructor, indicating this entire type is to appear static in C#
	TestClassWithOnlyStatics()
	{

	}

public:

	static void StaticMethod1()
	{
	}

	static void StaticMethod2()
	{	
	}
};
```

Running CSharp.Gen.csproj results in the following output from that:

```
public unsafe partial class TestClassWithOnlyStatics : IDisposable
    {
        [StructLayout(LayoutKind.Explicit, Size = 0)]
        public partial struct __Internal
        {
            [DllImport("CSharp.Native.dll", CallingConvention = global::System.Runtime.InteropServices.CallingConvention.ThisCall,
                EntryPoint="??0TestClassWithOnlyStatics@@QAE@ABV0@@Z")]
            internal static extern global::System.IntPtr cctor_1(global::System.IntPtr instance, global::System.IntPtr _0);

            [DllImport("CSharp.Native.dll", CallingConvention = global::System.Runtime.InteropServices.CallingConvention.Cdecl,
                EntryPoint="?StaticMethod1@TestClassWithOnlyStatics@@SAXXZ")]
            internal static extern void StaticMethod1_0();

            [DllImport("CSharp.Native.dll", CallingConvention = global::System.Runtime.InteropServices.CallingConvention.Cdecl,
                EntryPoint="?StaticMethod2@TestClassWithOnlyStatics@@SAXXZ")]
            internal static extern void StaticMethod2_0();
        }

        public global::System.IntPtr __Instance { get; protected set; }

        protected int __PointerAdjustment;
        internal static readonly System.Collections.Concurrent.ConcurrentDictionary<IntPtr, TestClassWithOnlyStatics> NativeToManagedMap = new System.Collections.Concurrent.ConcurrentDictionary<IntPtr, TestClassWithOnlyStatics>();
        protected void*[] __OriginalVTables;

        protected bool __ownsNativeInstance;

        internal static TestClassWithOnlyStatics __CreateInstance(global::System.IntPtr native, bool skipVTables = false)
        {
            return new TestClassWithOnlyStatics(native.ToPointer(), skipVTables);
        }

        internal static TestClassWithOnlyStatics __CreateInstance(TestClassWithOnlyStatics.__Internal native, bool skipVTables = false)
        {
            return new TestClassWithOnlyStatics(native, skipVTables);
        }

        private static void* __CopyValue(TestClassWithOnlyStatics.__Internal native)
        {
            var ret = Marshal.AllocHGlobal(0);
            *(TestClassWithOnlyStatics.__Internal*) ret = native;
            return ret.ToPointer();
        }

        private TestClassWithOnlyStatics(TestClassWithOnlyStatics.__Internal native, bool skipVTables = false)
            : this(__CopyValue(native), skipVTables)
        {
            __ownsNativeInstance = true;
            NativeToManagedMap[__Instance] = this;
        }

        protected TestClassWithOnlyStatics(void* native, bool skipVTables = false)
        {
            if (native == null)
                return;
            __Instance = new global::System.IntPtr(native);
        }

        public TestClassWithOnlyStatics(TestClassWithOnlyStatics _0)
        {
            __Instance = Marshal.AllocHGlobal(0);
            __ownsNativeInstance = true;
            NativeToManagedMap[__Instance] = this;
            *((TestClassWithOnlyStatics.__Internal*) __Instance) = *((TestClassWithOnlyStatics.__Internal*) _0.__Instance);
        }

        public void Dispose()
        {
            Dispose(disposing: true);
        }

        public virtual void Dispose(bool disposing)
        {
            TestClassWithOnlyStatics __dummy;
            NativeToManagedMap.TryRemove(__Instance, out __dummy);
            if (__ownsNativeInstance)
                Marshal.FreeHGlobal(__Instance);
        }

        public static void StaticMethod1()
        {
            __Internal.StaticMethod1_0();
        }

        public static void StaticMethod2()
        {
            __Internal.StaticMethod2_0();
        }
    }
```

genuinelucifer:
Fixes #743  
  
I had to change protected members to private as in C# static class cannot have protected members...  
Also, the way to check static is taken from [here](http://stackoverflow.com/questions/1175888/determine-if-a-type-is-static).  
  
Although my NUnit 2.0 could not find any tests, so I I haven't run the test after edit, but most probably it should pass. Do I need NUnit 3.0 to run the tests?

### DIFF
--- a/tests/Common/Common.Tests.cs
+++ b/tests/Common/Common.Tests.cs
@@ -339,6 +339,9 @@ public class CommonTests : GeneratorTestFixture
     [Test]
     public void TestStaticClasses()
     {
+        Type staticClassType = typeof(TestStaticClass);
+        // Only static class can be both abstract and sealed
+        Assert.IsTrue(staticClassType.IsAbstract && staticClassType.IsSealed);
         Assert.That(TestStaticClass.Add(1, 2), Is.EqualTo(3));
         Assert.That(TestStaticClass.OneTwoThree, Is.EqualTo(123));
         Assert.That(TestStaticClassDerived.Foo, Is.EqualTo(0));
--- a/tests/Common/Common.cpp
+++ b/tests/Common/Common.cpp
@@ -668,3 +668,8 @@ void hasPointerParam(const Foo& foo)
 void sMallFollowedByCapital()
 {
 }
+
+TestStaticClass& TestStaticClass::operator=(const TestStaticClass& oth)
+{
+    return *this;
+}
--- a/src/Generator/Passes/CheckStaticClass.cs
+++ b/src/Generator/Passes/CheckStaticClass.cs
@@ -90,8 +90,8 @@ namespace CppSharp.Passes
             // If one exists, we assume it's a factory function and the class is
             // not meant to be static. It's a simple heuristic but it should be
             // good enough for the time being.
-            if (@class.Functions.Any(ReturnsClassInstance) ||
-                @class.Methods.Any(ReturnsClassInstance))
+            if (@class.Functions.Any(m => !m.IsOperator && ReturnsClassInstance(m)) ||
+                @class.Methods.Any(m => !m.IsOperator && ReturnsClassInstance(m)))
                 return false;
 
             // If the class is to be used as an opaque type, then it cannot be
--- a/tests/Common/Common.h
+++ b/tests/Common/Common.h
@@ -385,13 +385,14 @@ struct DLL_API TestStaticClass
 
     static int GetOneTwoThree();
 
-protected:
+    TestStaticClass& operator=(const TestStaticClass& oth);
+
+private:
 
     static int _Mult(int a, int b);
 
     static int GetFourFiveSix();
 
-private:
     TestStaticClass();
 };
 


tests/Common/Common.Tests.cs
```diff
@@ -339,6 +339,9 @@ public void TestPropertyChains()
     [Test]
     public void TestStaticClasses()
     {
+        Type staticClassType = typeof(TestStaticClass).GetType();
```
ddobrev: .GetType() actually returns a .NET system type here, RuntimeType - that is, the type of the type. You only need typeof(TestStaticClass).

tests/Common/Common.cpp
```diff
@@ -668,3 +668,8 @@ void hasPointerParam(const Foo& foo)
 void sMallFollowedByCapital()
 {
 }
+
+TestStaticClass& TestStaticClass::operator=(const TestStaticClass& oth)
+{
+	return *this;
```
ddobrev: Please use 4 spaces instead of a type here.

ddobrev:
Abhinav, first of all, welcome back!
Yes, we switched to NUnit 3.6 but we have updated our reference. Perhaps you only need an updated test runner. I personally use https://marketplace.visualstudio.com/items?itemName=NUnitDevelopers.NUnit3TestAdapter, it enables test runs within VS itself.
Anyway, you definitely should run the tests locally first because I am almost certain your new test will fail - see my top remark in my review.

ddobrev:
@genuinelucifer also, since it's been a while since we worked together, please let me remind you. When fixing remarks, do not send new commits on top, instead amend and push force.

genuinelucifer:
Thanks. I fixed it and ran the test with NUnit 3 and it passed. I also run tests from inside VS only, by installing it from VS Marketplace.  
  
Please review again.

ddobrev:
@genuinelucifer thank you very much.

PR #796, opened 2017-03-17
Add a new pass to fix duplicate xml tags in comments.

### LINKED ISSUE #759: Duplicate <summary> in inline comments while rewriting XML comments in C++
##### Brief Description
Duplicate <summary> while rewriting XML comments in C++

C++:
```
/// <summary>text</summary>
enum_element = 0,
```

C#:
```
/// <summary><summary>text</summary></summary>
enum_element = 0,
```

Need to modify GenerateInlineSummary to something like:
```
if( comment.BriefText.StartsWith( "<summary>" ) && comment.BriefText.EndsWith( "</summary>" ) )
					WriteLine("{0} {1}", Options.CommentPrefix, comment.BriefText);
				else
					WriteLine( "{0} <summary>{1}</summary>", Options.CommentPrefix, comment.BriefText);
```
Ideally it should be able to parse C++ XML comments and rewrite(?) it to C# XML comments

OS: Windows / OS X / Linux (include version and/or distro)

##### Used settings

Target: MSVC/GCC/Clang

Other settings

genuinelucifer:
Fixes #759   
  
I think the comment added in `Common.h` produces desired result in `Common.cs`. I donno how to test this! Is it possible to get comments in the NUnit tests?  
  
Also, I don't think it will pass in the buildbot because I had to add `System.Xml.Linq` to References in the Generator as the new pass requires it. How can I make sure that dependency is automatically added when creating the project files? Sorry, I have no idea about how exactly the project files are generated using premake!

### DIFF
--- a/src/Generator/Passes/FixCommentsPass.cs
+++ b/src/Generator/Passes/FixCommentsPass.cs
@@ -1,0 +1,96 @@
+﻿using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Xml.Linq;
+using CppSharp.AST;
+
+namespace CppSharp.Passes
+{
+    /// <summary>
+    /// Pass that makes Xml Documentation Comments almost Xml Documentation Comments again.
+    /// </summary>
+    /// <remarks>
+    /// But has a bunch of para cruft that doesn't seem to be removable.
+    /// </remarks>
+    public class FixCommentsPass : TranslationUnitPass
+    {
+        private readonly Regex m_rx = new Regex(@"///(?<text>.*)", RegexOptions.ExplicitCapture | RegexOptions.Compiled);
+
+        public override bool VisitDeclaration(Declaration declaration)
+        {
+            if (AlreadyVisited(declaration))
+                return false;
+
+            //if (declaration.TranslationUnit == null || !declaration.TranslationUnit.FileName.StartsWith("Chakra"))
+            //    return false;
+
+            if (declaration.Comment != null)
+            {
+                var xDoc = GetOriginalDocumentationDocument(declaration.Comment.Text);
+                var xRoot = xDoc.Root;
+
+                declaration.Comment.Kind = CommentKind.BCPLSlash;
+                var fullComment = declaration.Comment.FullComment;
+                fullComment.Blocks.Clear();
+
+                var summaryPara = new ParagraphComment();
+                var summaryElement = xRoot.Element("summary");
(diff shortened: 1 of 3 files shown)


src/Generator/Passes/FixCommentPass.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            //if (declaration.TranslationUnit == null || !declaration.TranslationUnit.FileName.StartsWith("Chakra"))
+            //    return false;
+
+            if (declaration.Comment != null)
+            {
+                var xDoc = GetOriginalDocumentationDocument(declaration.Comment.Text);
+                var xRoot = xDoc.Root;
+
+                declaration.Comment.Kind = CommentKind.BCPLSlash;
```
genuinelucifer: I actually left everything as it was in this Pass' original writer's version. Just change something here. He used `RawCommentKind.BPCLSlash` from `CppSharp.Parser.AST` but seeing the type of left hand expression is not that. I changed it to `CommentKind.BPCLSlash` from `CppSharp.AST`. I hope it's okay.  
  
Also, me should mention somewhere that this code was taken from another repository...

src/Generator/Passes/FixCommentsPass.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            //if (declaration.TranslationUnit == null || !declaration.TranslationUnit.FileName.StartsWith("Chakra"))
+            //    return false;
+
+            if (declaration.Comment != null)
+            {
+                var xDoc = GetOriginalDocumentationDocument(declaration.Comment.Text);
+                var xRoot = xDoc.Root;
+
+                declaration.Comment.Kind = CommentKind.BCPLSlash;
```
genuinelucifer: I actually left everything as it was in this Pass' original writer's version. Just change something here. He used `RawCommentKind.BPCLSlash` from `CppSharp.Parser.AST` but seeing the type of left hand expression is not that. I changed it to `CommentKind.BPCLSlash` from `CppSharp.AST`. I hope it's okay.  
  
Also, we should mention somewhere that this code was taken from another repository...

genuinelucifer:
I saw that the name was missing a `s` so I just renamed the file to `FixCommentsPass` to match the pass name exactly. I've reposted my comment on the file.

tritao:
Should we just add this pass as a default pass instead of adding it exclusively to Common tests?

ddobrev:
@tritao as I was trying to tell @genuinelucifer , I don't think we need a pass at all. Just a check not to add a repetitive summary when printing the comment.

ddobrev:
Already fixed by https://github.com/mono/CppSharp/commit/d7c8a61ecc2d29e78a21f6605f39877cf093d004 .

PR #797, opened 2017-03-18
Incomplete classes and structs are now generated for C#.

### LINKED ISSUE #779: Partially defined structs not being treated as IntPtr
Imagine the following simple c api.

```c
enum medx_dicom_response
{
  ok,
  error
}
struct medx_dicom_server;

medx_dicom_response test1(
    medx_dicom_server** server
);

medx_dicom_response test2(
    medx_dicom_server* server
);

medx_dicom_response test3(
    medx_dicom_server* server
);

medx_dicom_response test4(
    void*
);

medx_dicom_response test5(
    void**
);

medx_dicom_response test6(
);
```

With this API, only ```test4```, ```test4``` and ```test6``` get generated.

I think the other methods should be generated with an IntPtr (or maybe something more strongly typed? does't matter).

realvictorprm:
They're stil W.I.P and template classes don't work yet.

This should fix #779

### DIFF
--- a/src/AST/Class.cs
+++ b/src/AST/Class.cs
@@ -96,8 +96,20 @@ namespace CppSharp.AST
         // True if the class is final / sealed.
         public bool IsFinal { get; set; }
 
+        private bool? isOpaque = null;
+
         // True if the type is to be treated as opaque.
-        public bool IsOpaque;
+        public bool IsOpaque
+        {
+            get
+            {
+                return isOpaque == null ? IsIncomplete && CompleteDeclaration == null : isOpaque.Value;
+            }
+            set
+            {
+                isOpaque = value;
+            }
+        }
 
         // True if the class is dynamic.
         public bool IsDynamic;
@@ -127,7 +139,6 @@ namespace CppSharp.AST
             IsAbstract = false;
             IsUnion = false;
             IsFinal = false;
-            IsOpaque = false;
             IsPOD = false;
             Type = ClassType.RefType;
             Layout = new ClassLayout();
--- a/src/Generator/Generators/CSharp/CSharpSources.cs
+++ b/src/Generator/Generators/CSharp/CSharpSources.cs
@@ -272,7 +272,7 @@ namespace CppSharp.Generators.CSharp
 
         public override bool VisitClassDecl(Class @class)
         {
-            if (@class.IsIncomplete)
+            if (@class.IsIncomplete && !@class.IsOpaque)
                 return false;
 
             if (@class.IsInterface)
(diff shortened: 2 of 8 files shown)


realvictorprm:
The commit breakes CLI generation. I'm trying to fix it.

src/AST/Class.cs
```diff
@@ -96,8 +96,20 @@ public class Class : DeclarationContext
         // True if the class is final / sealed.
         public bool IsFinal { get; set; }
 
+        public bool? CustomOpaque = null;
```
ddobrev: Make private and rename to "isCustom".

src/Generator/Generators/CLI/CLIHeaders.cs
```diff
@@ -164,10 +164,7 @@ public void GenerateDeclContext(DeclarationContext decl)
             // Generate all the struct/class declarations for the module.
             foreach (var @class in decl.Classes)
             {
-                if (!@class.IsGenerated || @class.IsIncomplete || @class.IsDependent)
-                    continue;
-
-                if (@class.IsOpaque)
+                if (!@class.IsGenerated || (@class.IsIncomplete && !@class.IsOpaque) || @class.IsDependent)
```
ddobrev: You only need to check IsOpaque, not IsIncomplete as well. Please fix here and anywhere else you do this.

tests/Common/Common.Tests.cs
```diff
@@ -723,4 +723,22 @@ public void TestFuncWithUnionParam()
         var v = Common.FuncUnion(ut);
         Assert.AreEqual(20, v);
     }
+
+    [Test]
+    public void TestGenerationOfIncompleteClasses()
+    {
+        Type[] types = System.Reflection.Assembly.GetExecutingAssembly().GetTypes();
```
ddobrev: Don't use reflection here, just call the generated binding.

tests/Common/Common.cpp
```diff
@@ -674,6 +674,18 @@ void sMallFollowedByCapital()
...
+struct IncompleteStruct {};
+
+IncompleteStruct* createIncompleteStruct()
+{
+	return new IncompleteStruct();
+}
+
+void UseIncompleteStruct(IncompleteStruct* a)
+{
+	return;
```
ddobrev: We use 4 spaces in C++, not a tab.

tests/Common/Common.h
```diff
@@ -1039,20 +1039,20 @@ void NonTrivialDtor::setDtorCalled(bool value)
...
-{
-    ForwardedTemplate<T> functionInForwardedTemplate() const;
-};
-
-template <class T>
-ForwardedTemplate<T> ForwardedTemplate<T>::functionInForwardedTemplate() const
-{
-    return ForwardedTemplate<T>();
-}
+//template <class T> class ForwardedTemplate;
```
ddobrev: Restore these and just skip opaque classes if they are dependent.

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -364,6 +363,55 @@ public override bool VisitClassDecl(Class @class)
 
                 if (@class.IsDynamic)
                     GenerateVTable(@class);
+            } else
```
ddobrev: All of this code looks identical to the one in the not opaque case so why duplicate it instead of just removing the check?

src/Generator/Passes/CheckIgnoredDecls.cs
```diff
@@ -367,6 +367,16 @@ private bool HasInvalidDecl(Declaration decl, out string msg)
             {
                 msg = "null";
                 return true;
+            }            
+
+            if (decl is Class)
```
ddobrev: In general, do not use "is" + "as" because these are two type casts which is a little slower. Instead, use "as" + a null check. Please fix here and anywhere else you've used it this way.

src/Generator/Generators/CSharp/CSharpSources.cs
realvictorprm: If we make adjustments. That's why. But If you're fine without any adjustments to the generations, I'll remove it.

ddobrev: If we do, they will be to specific pieces in the large block. So we'd better off just with one block for the time being.

PR #804, opened 2017-03-30
Fixing class specifiers in namespaces

realvictorprm:
The specifiers for classes which own any attribute which isn't public have to be handled different in namespaces.
The old code generated not compileable code.


tritao:
Not sure I understand this change, a test would be nice.

Why do we need this?

tritao:
@realvictorprm Should we still merge this? Can you add a test?

realvictorprm:
LeT me see...

Issue #806, opened 2017-03-31
Duplicate generation of incomplete structs


realvictorprm:
...if they're defined in multiple headers.

Thanks to @ddobrev we found another bug which has to be fixed.
### Brief Description:
Forward declarations in a header and the implementation in a different header lead to multiple generation of the same class. This can break several code and most be noticed as a fatal bug. 
#### Header 1:
```
typedef struct Incomplete Incomplete;
```
#### Header 2:
```
#include "Header1.h"
struct Incomplete{

}
```

This will generate two time's the Incomplete struct.

### Status:
I already discussed with @ddobrev what's to do.
Because I introduced the feature I'll take care of the bugs I'm responsible for.

- [x] Fix the bug

realvictorprm:
I've made a PR which fixes the Bug so fellow devs out there, the bug fixing update is near. I have to make only small normal adjustments, which will be done tomorrow.

realvictorprm:
We're having some issues, stand by we're still trying to fix it.

realvictorprm:
This if fixed, right @ddobrev

PR #809, opened 2017-04-04
Fixing critical duplicate generation bug.

realvictorprm:
This should fix bug #806
The test consists simply of that the generated code is compileable.

### DIFF
--- a/docs/GeneratingBindings.md
+++ b/docs/GeneratingBindings.md
@@ -1,7 +1,7 @@
 # Generating bindings
 
-Suppose we have the following declarations in a file named `Sample.h` and we
-want to bind it to .NET.
+Suppose we have the following declarations in a file named `Sample.h` located
+in `C:\Sample\include` and we want to bind it to .NET.
 
 ```csharp
 class Foo
@@ -77,7 +77,9 @@ void Setup(Driver driver)
     var options = driver.Options;
     options.GeneratorKind = GeneratorKind.CSharp;
     var module = options.AddModule("Sample");
+    module.IncludeDirs.Add("C:\Sample\include");
     module.Headers.Add("Sample.h");
+    module.LibraryDirs.Add("C:\Sample\lib");
     module.Libraries.Add("Sample.lib");
 }
 ```


src/AST/Declaration.cs
```diff
@@ -315,6 +315,8 @@ public virtual bool Ignore
         // another translation unit. 
         public string USR;
 
+        public List<Declaration> Redeclarations = new List<Declaration>();
```
ddobrev: Do not use public fields. We do still have them in many places but we gradually replace them with properties. In this case it's really convenient to use C# 6.0 property initialisers:

    public List<Declaration> Redeclarations { get; } = new List<Declaration>();

src/Parser/ASTConverter.cs
```diff
@@ -940,6 +940,27 @@ void VisitDeclaration(Declaration decl, AST.Declaration _decl)
             _decl.IsIncomplete = decl.IsIncomplete;
             _decl.IsDependent = decl.IsDependent;
             _decl.DefinitionOrder = decl.DefinitionOrder;
+            if (decl.RedeclarationsCount > 0) Diagnostics.Debug($"There are redeclarations of decl {decl.Name}");
+            for (uint i = 0; i < decl.RedeclarationsCount; i++)
+            {
+                if(_decl.Name.Contains("HasGetterAndOverriddenSetter"))System.Diagnostics.Debugger.Break();
```
ddobrev: Remove any debugging code.

src/CppParser/Parser.cpp
```diff
@@ -903,6 +903,14 @@ void Parser::WalkRecord(const clang::RecordDecl* Record, Class* RC)
         ReadClassLayout(RC, Record, CharUnits(), true);
     }
 
+    if (Record->redecls_begin() != Record->redecls_end()) {
```
ddobrev: Some formatting here: opening braces must go to their own line, local variables should start with a capital letter (Clang style)

src/CppParser/Parser.cpp
```diff
@@ -903,6 +903,14 @@ void Parser::WalkRecord(const clang::RecordDecl* Record, Class* RC)
...
+        for (auto redecl : Record->redecls()) {
+            auto dec = WalkDeclaration(redecl);
```
ddobrev: Rename "dec" to "Decl".

src/CppParser/Parser.cpp
```diff
@@ -903,6 +903,14 @@ void Parser::WalkRecord(const clang::RecordDecl* Record, Class* RC)
...
+            if (dec != nullptr)
```
ddobrev: I don't think it's possible for the returned result to be null so remove the check unless it really is necessary.

src/Generator/Passes/ResolveIncompleteDeclsPass.cs
```diff
@@ -68,7 +68,17 @@ private void EnsureCompleteDeclaration(Declaration declaration)
                 return;
 
             var @class = declaration as Class;
-            if (@class != null && @class.IsOpaque) return;
+            if (@class != null && @class.IsOpaque)
+            {
+                if (@class.Redeclarations.Count > 0)
+                {
+                    if (@class.Redeclarations.ElementAt(0) == @class &&
```
ddobrev: Use [] instead of ElementAt(), it's simpler.

src/Generator/Passes/ResolveIncompleteDeclsPass.cs
```diff
@@ -68,7 +68,17 @@ private void EnsureCompleteDeclaration(Declaration declaration)
...
+                                           !@class.Redeclarations.Exists((Declaration c) => { return c.CompleteDeclaration != null; }))
```
ddobrev: Align to 4 spaces after the "if" above.

src/Generator/Passes/ResolveIncompleteDeclsPass.cs
```diff
@@ -68,7 +68,17 @@ private void EnsureCompleteDeclaration(Declaration declaration)
...
+                        return;
+                }
+                else
```
ddobrev: To reduce nesting, invert the "if" and exit early.

tests/Common/AnotherUnit.h
```diff
@@ -3,6 +3,8 @@
 // Verifies the header is included when the delegate is defined in a different file
 typedef void (*DelegateInAnotherUnit)();
 
+struct DuplicateStruct;
```
ddobrev: Add a function which uses this struct and call that function in the test code. We must ensure such functions are generated.

tests/Common/Common.h
```diff
@@ -1227,3 +1227,7 @@ enum EmptyEnum { };
 enum __enum_with_underscores { lOWER_BEFORE_CAPITAL, CAPITALS_More, underscore_at_end_, usesDigits1_0 };
 
 void DLL_API sMallFollowedByCapital();
+
```
ddobrev: Add a function which uses this struct and call that function in the test code. We must ensure such functions are generated. In fact, you don't need to use a new incomplete structure. The one we already have - I believe it's called "IncompleteStruct" - should do just fine, simply forward it in AnotherUnit.h as well.

src/Parser/ASTConverter.cs
tritao: Some debug code here.

realvictorprm: Oh I'm very sorry. I thought I've removed all.

src/CppParser/Parser.cpp
realvictorprm: It is needed for safety. I encountered that it could happen that the iterator for Redeclarations iterates over a list with empty entries.

realvictorprm:
Okay I've read everything and I'll apply the changes tomorrow after my final English exam. Till then, cross the fingers for me :smiley: !

src/CppParser/Parser.cpp
ddobrev: Remove this check, the C++ "for" does it for you.

src/Generator/Passes/ResolveIncompleteDeclsPass.cs
realvictorprm: Sorry, bad habit from my java times

src/Parser/ASTConverter.cs
realvictorprm: done

src/AST/Declaration.cs
realvictorprm: done

src/Parser/ASTConverter.cs
realvictorprm: done

src/CppParser/Parser.cpp
realvictorprm: done

src/Generator/Passes/ResolveIncompleteDeclsPass.cs
realvictorprm: done

tests/Common/Common.h
realvictorprm: done

src/CppParser/Parser.cpp
```diff
@@ -903,6 +903,12 @@ void Parser::WalkRecord(const clang::RecordDecl* Record, Class* RC)
         ReadClassLayout(RC, Record, CharUnits(), true);
     }
 
+    for (auto redecl : Record->redecls()) {
```
ddobrev: Move '{' to its own line.

src/Generator/Passes/ResolveIncompleteDeclsPass.cs
```diff
@@ -68,7 +68,13 @@ private void EnsureCompleteDeclaration(Declaration declaration)
                 return;
 
             var @class = declaration as Class;
-            if (@class != null && @class.IsOpaque) return;
+            if (@class != null && @class.IsOpaque)
```
ddobrev: Combine all conditions for an early exit to a single conditional operator.

src/Parser/ASTConverter.cs
```diff
@@ -940,6 +940,24 @@ void VisitDeclaration(Declaration decl, AST.Declaration _decl)
             _decl.IsIncomplete = decl.IsIncomplete;
             _decl.IsDependent = decl.IsDependent;
             _decl.DefinitionOrder = decl.DefinitionOrder;
+            for (uint i = 0; i < decl.RedeclarationsCount; i++)
+            {
+                var redecl = decl.GetRedeclarations(i);
+                if (Declarations.ContainsKey(redecl.OriginalPtr))
```
ddobrev: Do not do this yourself but rather call redecl.Visit. It checks the dictionary for you.

src/Generator/Passes/ResolveIncompleteDeclsPass.cs
```diff
@@ -68,7 +68,13 @@ private void EnsureCompleteDeclaration(Declaration declaration)
...
+            {
+                if (@class.Redeclarations.Count == 0) return;
+                else if (@class.Redeclarations[0] == @class &&
+                    !@class.Redeclarations.Exists((Declaration c) => { return c.CompleteDeclaration != null; }))
```
ddobrev: I think this lambda can be shortened to 'c => c.CompleteDeclaration != null'.

PR #812, opened 2017-04-08
Preparing fix for duplicate generated structs.

### LINKED ISSUE #806: Duplicate generation of incomplete structs

realvictorprm:
W.I.P
should fix #806

### DIFF
--- a/src/CppParser/Parser.cpp
+++ b/src/CppParser/Parser.cpp
@@ -3489,7 +3489,7 @@ Declaration* Parser::WalkDeclarationDef(clang::Decl* D)
 }
 
 Declaration* Parser::WalkDeclaration(const clang::Decl* D,
-                                           bool CanBeDefinition)
+                                           bool CanBeDefinition, bool WalkRedecls)
 {
     using namespace clang;
 
@@ -3540,6 +3540,10 @@ Declaration* Parser::WalkDeclaration(const clang::Decl* D,
         auto RD = cast<CXXRecordDecl>(D);
 
         auto Class = WalkRecordCXX(RD);
+
+        if (WalkRedecls)
+            for (auto redecl : RD->redecls())
+                Class->Redeclarations.push_back(WalkDeclaration(redecl, CanBeDefinition, false));
 
         // We store a definition order index into the declarations.
         // This is needed because declarations are added to their contexts as
--- a/src/CppParser/Parser.h
+++ b/src/CppParser/Parser.h
@@ -61,7 +61,8 @@ public:
 private:
     // AST traversers
     void WalkAST();
-    Declaration* WalkDeclaration(const clang::Decl* D, bool CanBeDefinition = false);
+    Declaration* WalkDeclaration(const clang::Decl* D, bool CanBeDefinition = false,
+        bool WalkDeclarations = true);
     Declaration* WalkDeclarationDef(clang::Decl* D);
     Enumeration* WalkEnum(const clang::EnumDecl* ED);
 	Enumeration::Item* WalkEnumItem(clang::EnumConstantDecl* ECD);
--- a/src/Generator/Passes/ResolveIncompleteDeclsPass.cs
+++ b/src/Generator/Passes/ResolveIncompleteDeclsPass.cs
@@ -1,5 +1,6 @@
 ﻿using System.Linq;
 using CppSharp.AST;
+using System.Collections.Generic;
 
 namespace CppSharp.Passes
 {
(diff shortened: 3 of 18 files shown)


src/CppParser/Parser.cpp
```diff
@@ -795,6 +795,9 @@ Class* Parser::GetRecord(const clang::RecordDecl* Record, bool& Process)
     using namespace clang;
     Process = false;
 
+    if (Record->isInjectedClassName())
```
ddobrev: You restored old code in this file - it's most probably a rebasing error.

src/CppParser/Parser.cpp
```diff
@@ -3541,6 +3543,18 @@ Declaration* Parser::WalkDeclaration(const clang::Decl* D,
 
         auto Class = WalkRecordCXX(RD);
 
+        if (WalkRedecls && Class != nullptr)
```
ddobrev: Is this check for null, as well as the one below, really necessary now that we handle injected classes?

src/Generator/Passes/ResolveIncompleteDeclsPass.cs
```diff
@@ -68,7 +68,16 @@ private void EnsureCompleteDeclaration(Declaration declaration)
                 return;
 
             var @class = declaration as Class;
-            if (@class != null && @class.IsOpaque) return;
+            if (@class != null && @class.IsOpaque)
+                if (@class.Redeclarations.Count == 0)
+                    return;
+                else if (@class.Redeclarations[0] == @class && !@class.Redeclarations.Exists((Declaration c) => c.CompleteDeclaration != null))
+                {
+                    Diagnostics.Message($"There are redeclarations of class {@class.Name} at line {@class.LineNumberStart}");
```
ddobrev: Remove this debugging code. This will also allow you to merge all conditions into one.

src/Parser/ASTConverter.cs
```diff
@@ -955,6 +955,16 @@ void VisitDeclaration(Declaration decl, AST.Declaration _decl)
             _decl.OriginalPtr = originalPtr;
 
             NativeObjects.Add(decl);
+
+            for (uint i = 0; i < decl.RedeclarationsCount; i++)
+            {
+                var redecl = decl.GetRedeclarations(i);
+                if (redecl.OriginalPtr != _decl.OriginalPtr)
```
ddobrev: This isn't necessary because the current declaration is already added to the cache.

src/CppParser/Parser.cpp
```diff
@@ -3541,6 +3541,10 @@ Declaration* Parser::WalkDeclaration(const clang::Decl* D,
 
         auto Class = WalkRecordCXX(RD);
 
+        if (WalkRedecls)
+            for (auto redecl : RD->redecls())
+                Class->Redeclarations.push_back(WalkDeclaration(redecl, true, false));
```
ddobrev: Pass CanBeDefinition on instead of hard-coding "true".

src/CppParser/Parser.h
```diff
@@ -61,7 +61,7 @@ class Parser
 private:
     // AST traversers
     void WalkAST();
-    Declaration* WalkDeclaration(const clang::Decl* D, bool CanBeDefinition = false);
+    Declaration* WalkDeclaration(const clang::Decl* D, bool CanBeDefinition = false, bool WalkDeclarations = true);
```
ddobrev: We avoid long lines, move WalkDeclarations to a new one.

src/Generator/Passes/ResolveIncompleteDeclsPass.cs
```diff
@@ -68,8 +69,14 @@ private void EnsureCompleteDeclaration(Declaration declaration)
                 return;
 
             var @class = declaration as Class;
-            if (@class != null && @class.IsOpaque) return;
-
+            if (@class != null && @class.IsOpaque)
+                if (@class.Redeclarations.Count == 0)
+                    return;
+                else if (@class.Redeclarations.Last() == @class && !@class.Redeclarations.Exists((Declaration c) => !c.IsIncomplete))
```
ddobrev: Shorten the lambda and merge with the condition above.

src/Generator/Passes/ResolveIncompleteDeclsPass.cs
```diff
@@ -68,8 +69,14 @@ private void EnsureCompleteDeclaration(Declaration declaration)
...
+                    return;
+                else
```
ddobrev: Remove the redundant else.

ddobrev:
This has been manually merged at https://github.com/mono/CppSharp/commit/10c86a2a10e0e3337ba5d6075a380ed9e2c8bc4b.

PR #816, opened 2017-04-15
Fixed wrong recording of declaration order

realvictorprm:
This is a bugfix for critical bug. With the fix the generation of the parser bindings should be nearly correct.


realvictorprm:
👍

PR #828, opened 2017-05-08
Deleted Obsolete Options and replaced them with Modules

### LINKED ISSUE #766: Remove the general options related to native libraries
CppSharp has had settings such as Options.Headers, Options.Libraries and others from the very beginning. However, they reflect its past limitation of only being able to produce one binding per run. With the Module class which we've had for quite some time now, this is no longer true. Therefore it should remain the only way of specifying settings of libraries to wrap.
We should also update any parts of our documentation which talk about these options to remove.

mohtamohit:
Removed the Obsolete Options.Headers and Options.Libraries and replaced
it with module. Generator.cs and CLI.cs still need to be updated, though.
Fixes #766

### DIFF
--- a/src/CppParser/Bootstrap/Bootstrap.cs
+++ b/src/CppParser/Bootstrap/Bootstrap.cs
@@ -31,27 +31,36 @@ namespace CppSharp
         public void Setup(Driver driver)
         {
             var options = driver.Options;
-            options.LibraryName = "CppSharp";
+            options.GeneratorKind = GeneratorKind.CSharp;
             options.DryRun = true;
-            options.Headers.AddRange(new string[]
-            {
-                "clang/AST/Expr.h",
-            });
-
             options.SetupXcode();
             options.MicrosoftMode = false;
             options.TargetTriple = "i686-apple-darwin12.4.0";
 
-            options.addDefines ("__STDC_LIMIT_MACROS");
-            options.addDefines ("__STDC_CONSTANT_MACROS");
+            var module = options.AddModule("CppSharp");
+
+            module.Defines.Add("__STDC_LIMIT_MACROS");
+            module.Defines.Add("__STDC_CONSTANT_MACROS");
 
             var llvmPath = Path.Combine (GetSourceDirectory ("deps"), "llvm");
             var clangPath = Path.Combine(llvmPath, "tools", "clang");
 
-            options.addIncludeDirs(Path.Combine(llvmPath, "include"));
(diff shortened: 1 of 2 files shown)


tritao:
Don't merge yet as I think it will break Embeddinator, I need to check some things first.

src/CppParser/Bootstrap/Bootstrap.cs
```diff
@@ -31,27 +31,30 @@ static string GetSourceDirectory(string dir)
         public void Setup(Driver driver)
         {
             var options = driver.Options;
-            options.LibraryName = "CppSharp";
+            options.GeneratorKind = GeneratorKind.Csharp;
             options.DryRun = true;
-            options.Headers.AddRange(new string[]
```
ddobrev: You've deleted this but you haven't added it to the module.

src/CppParser/Bootstrap/Bootstrap.cs
```diff
@@ -31,27 +31,30 @@ static string GetSourceDirectory(string dir)
...
-
             options.SetupXcode();
             options.MicrosoftMode = false;
             options.TargetTriple = "i686-apple-darwin12.4.0";
 
-            options.addDefines ("__STDC_LIMIT_MACROS");
-            options.addDefines ("__STDC_CONSTANT_MACROS");
+            var module = options.AddModule("CppSharp");
+
+            module.Defines.add("__STDC_LIMIT_MACROS");
```
ddobrev: How does the lower case work when List<>.Add is upper-cased? Does your code compile at all?

src/CppParser/Bootstrap/Bootstrap.cs
mohtamohit: Sorry, I forgot to do that. I made the the necessary changes. Actually, since I rebased the code to the latest one, this one is again appearing as a separate commit only (sorry for that).

realvictorprm:
@tritao, @ddobrev yes I tested it.
Embeddinator need to change following files (and the corresponding location):
```
Embeddinator-4000\binder\Generators\AstGenerator.cs(59,32,59,43): 
Embeddinator-4000\binder\Compilation.cs(456,66,456,77): 
Embeddinator-4000\binder\Driver.cs(31,21,31,31): 
Embeddinator-4000\binder\Tests\TestsGenerator.cs(35,21,35,32): 
Embeddinator-4000\binder\Tests\TestsGenerator.cs(39,25,39,36):
``

PR #837, opened 2017-05-24
Stack mismatch (bool param) + Implement default c'tors (without DllImports)

giladlevi:
This pull request includes fixes for both issues 831 and 834.


ddobrev:
@giladlevi I appreciate your effort but I am extremely sorry to say these fixes are workarounds at best. First of all, I would ask you to send separate pull requests, starting with the stack mismatch. Second, if you need any help fixing it properly, I am here to help so ask away.

Issue #834, opened 2017-05-21
Stack mismatch when bool param passed from C++ to C# 


giladlevi:
I have a c++ class that I inherits from and overrides some virtual functions.

The C++ virtual function looks like this:
```c++
virtual void OnSomething(bool iAmABool) {};
```

C++ stores booleans (at least on VC++) as 1-byte.
Therefore, when C++ calls the generated delegate it will push 1 byte onto the stack (for example, in the case of x86 thiscall)
The generated delegate on the c# side will looks like this:
```c#
[SuppressUnmanagedCodeSecurity, UnmanagedFunctionPointer(global::System.Runtime.InteropServices.CallingConvention.ThisCall)]
internal unsafe delegate void Action_bool(bool instance);
```
This signature uses the c# bool type for the delegate definition which apparently consumes 4 bytes off the stack.

This leads to a most unfortunate mismatch, where 3 bytes of the c# bool are basically random garbage read from the stack.

ddobrev:
Thank you for your report. We know where the bug is so since we're satisfied with the code you sent the last time, I would like to let you know I'll be happy to provide any assistance you might need should you decide to try fixing this one as well.

giladlevi:
@ddobrev I did it, I'll open a pull request.
Please take a look on issue https://github.com/mono/CppSharp/issues/831 that I wrote also few days ago. I already fixed it also, maybe can I open 2 pull requests?

ddobrev:
@giladlevi excellent, please do.
About #831, I am a little surprised you managed. It's not trivial. I would be curious indeed to see your solution.

Issue #511, opened 2015-07-02
Bind delegates with higher-level managed types


tritao:
Imagine a callback in native code:
```void Callback(Class& c);```

When bound in managed it ends up as:
```delegate void Callback(IntPtr c);```

And we need to do later:
```var c = Class.__CreateInstance(c)```

Not pretty.
I think we could improve this on the generator to generate a lower-level (same as what we already generate) and higher-level proxy delegate:
```delegate void Callback(Class c);```

genuinelucifer:
@tritao How to generate 'delegate' from the function 'Callback'.. I tried to recreate it but the function ended up as another function in managed code.

realvictorprm:
Sorry for posting duplicates. Blame me!

I'm presenting an update to the current situation:

> The new generated code should look like this:
> 
> ```
> public delegate void HighLevelDelegate (SomeClass class, int someparam);
> internal delegate void LowLevelDelegate(IntPtr class, int someparam);
> 
> public void SetCallback(HighLevelDelegate del);
> internal void SetCallback(LowLevelDelegate del);
> ```
> However our current approach includes that each callback will be saved until the life-time of a specific class or the application itself ends.

PR #838, opened 2017-05-24
[Fix issue #834] Stack mismatch when bool param passed from C++ to C#…

giladlevi:
…, cast to sbyte instead. keeping the interface

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpSources.cs
+++ b/src/Generator/Generators/CSharp/CSharpSources.cs
@@ -2647,7 +2647,7 @@ namespace CppSharp.Generators.CSharp
                     && !string.IsNullOrWhiteSpace(param.Context.ArgumentPrefix))
                     name.Append(param.Context.ArgumentPrefix);
 
-                var paramString = param.Param.Type.IsPrimitiveType() && ((BuiltinType)param.Param.Type).Type == PrimitiveType.Bool? $"{param.Name} ? (sbyte)1: (sbyte)0" : param.Name;
+                var paramString = param.Param.Type.IsPrimitiveType() && (param.Param.Type as BuiltinType)?.Type == PrimitiveType.Bool? $"{param.Name} ? (sbyte)1: (sbyte)0" : param.Name;
                 name.Append(paramString);
                 names.Add(name.ToString());
             }
--- a/src/Generator/Generators/CSharp/CSharpSources.cs
+++ b/src/Generator/Generators/CSharp/CSharpSources.cs
@@ -537,7 +537,16 @@ namespace CppSharp.Generators.CSharp
 
             retType = function.ReturnType.CSharpType(TypePrinter);
 
-            var @params = function.GatherInternalParams(Context.ParserOptions.IsItaniumLikeAbi).Select(p =>
+            var paramsCopy = function
+              .GatherInternalParams(Context.ParserOptions.IsItaniumLikeAbi, false)
+              .Select(p => new Parameter(p)).ToArray();
+
+            foreach (var p in paramsCopy.Where(paramCopy => paramCopy.Type.IsPrimitiveType() && (paramCopy.Type as BuiltinType)?.Type == PrimitiveType.Bool))
+            {
+                p.QualifiedType = new QualifiedType(new BuiltinType(PrimitiveType.Char));
+            }
+
+            var @params = paramsCopy.Select(p =>
                 string.Format("{0} {1}", p.CSharpType(TypePrinter), p.Name)).ToList();
 
             TypePrinter.PopContext();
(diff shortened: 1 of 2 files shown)


tritao:
I think the only "right" way to fix this is as part of https://github.com/mono/CppSharp/issues/511, we need to generate a native and higher-level delegate and marshal types in the native delegate so we can call the managed delegate. So in this case, the native delegate would have the `byte` parameter, but the higher level delegate would still have the `bool` signature as expected.

@ddobrev What do you think?

ddobrev:
@tritao it's actually simpler. We just need to only print delegates with the native context. That we'll have the managed delegates later on is actually irrelevant to the problem.

ddobrev:
@giladlevi thank you for separating the pull. I would now like to help you fix it properly. Basically, we need to print parameters of delegates as if they are in a native context rather than managed. In the native context a bool is a byte, as you correctly said. So you need to update our code so that Boolean parameters of delegates are printed as bytes. You can use MarshalKind to see what context you are in. Any questions, please ask.

tritao:
To add on to what @ddobrev has said, you should also be looking at ParserTargetInfo.BoolWidth member, to check the size of booleans in the parsed target ABI. We should only map them to `byte` if the current target has byte-sized booleans.

giladlevi:
@ddobrev I think this is actually what i'm doing https://github.com/giladlevi/CppSharp/blob/stackMismatch/src/Generator/Generators/CSharp/CSharpSources.cs#L1560
By default MarshalKind is Unknown.

ddobrev:
@giladlevi yes but my point is that if it was fixed in the right place, all of your remaining code would've been unnecessary.

giladlevi:
@ddobrev 
I have 3 places that require change in my opinion to fix the issue:

1.  At the declaration of the delegate type used for native => managed transition:
I must change the arg1 param to sbyte (it is currently generated as bool in c#).
This is critical, otherwise, a bool param in c++ (1 byte) will be read as 4 bytes from the stack by the generated c# code due to the wrong signature and completely mess up the stack (potential stack overflow exception / or reading bytes from the next param bytes in to `bool`)
This is done in [this area](https://github.com/giladlevi/CppSharp/blob/stackMismatch/src/Generator/Generators/CSharp/CSharpTypePrinter.cs#L632)
```c#
internal unsafe delegate void Action_IntPtr_bool(global::System.IntPtr instance, sbyte arg1);
```

2. At the actual delegate "instantiation" for a specific hooked virtual function I must perform the same change as (1).
private static void _OnEvent_0DelegateHook(global::System.IntPtr instance, **sbyte** bIsLast)
This is done in [this area](https://github.com/giladlevi/CppSharp/blob/stackMismatch/src/Generator/Generators/CSharp/CSharpSources.cs#L540)
```c#
private static void _OnEvent_0DelegateHook(global::System.IntPtr instance, sbyte bIsLast)
```

3. The existing code can generate sbyte => bool casts in the form of `someSbyte != 0` .
This is why I change the MarshalKind to MarshalKind.NativeField instead of MarshalKind.Unknown
This is already implemented by you [in this area](https://github.com/giladlevi/CppSharp/blob/stackMismatch/src/Generator/Generators/CSharp/CSharpMarshal.cs#L202) and I fixed the flow to go there from [here](https://github.com/giladlevi/CppSharp/blob/stackMismatch/src/Generator/Generators/CSharp/CSharpSources.cs#L1560)
```c#
private static void _OnEvent_0DelegateHook(global::System.IntPtr instance, sbyte bIsLast)
        {
            if (!NativeToManagedMap.ContainsKey(instance))
                throw new global::System.Exception("No managed instance was found");

            var __target = (global::LTSApi.CSecurityFtdcQuerySpi) NativeToManagedMap[instance];
            if (__target.__ownsNativeInstance)
                __target.SetupVTables();
            __target.OnEvent(bIsLast != 0);
        }
```

4. at the implementation of the base virtual function (which performs a managed => native transition) I want to keep the signature with a c# `bool` while casting to `sbyte` when calling the native delegate so I need to perform a cast like `someBool ? (sbyte)1: (sbyte)0`. 
This come from [here](https://github.com/giladlevi/CppSharp/blob/stackMismatch/src/Generator/Generators/CSharp/CSharpSources.cs#L2650)
```c#
public virtual void OnEvent(bool bIsLast)
        {
            var __slot = *(void**) ((IntPtr) __OriginalVTables[0] + 3 * 8);
            var ___OnEvent_0Delegate = (global::LTSApi.Delegates.Action_IntPtr_bool) Marshal.GetDelegateForFunctionPointer(new IntPtr(__slot), typeof(global::LTSApi.Delegates.Action_IntPtr_bool));
            ___OnEvent_0Delegate((__Instance + __PointerAdjustment), bIsLast ? (sbyte)1: (sbyte)0);
        }
```

Sometimes I need param's QualifiedType to be PrimitiveType.Char and sometimes PrimitiveType.Bool.
How all this changes can be performed only with changing to MarshalKind.NativeField?

ddobrev:
@giladlevi I understand what you mean, I should've explained myself better. What I am talking about is changing the type printer instead of the generator. We need to extend the action of the context now called NativeField to also apply to delegates. You see, delegates already use the Native context, this is why we get IntPtr and .__Internal instead of the real classes. We need to use NativeField instead because that context, since it directly concerns marshalling, also converts bools to bytes. So, since NativeField would no longer be used just for delegates, let's rename it to, say, NativeType. Then whenever we need a delegate, for v-tables or other, we push NativeType instead of just Native. Makes sense?

tritao:
Closing, issue has been fixed.

PR #849, opened 2017-06-02
Duplication of inline comments

ktopouzi:


src/Generator.Tests/AST/TestAST.cs
```diff
@@ -405,6 +405,10 @@ public void TestComments()
 // <para>feeding the input stream for the debugger, or NULL if there is</para>
 // <para>no string for this control key.</para>
 // </returns>".Replace("\r", string.Empty), commentMethod.Replace("\r", string.Empty));
+
+            var enumClass = AstContext.FindCompleteClass("TestInlineComments");
```
ktopouzi: This is where i got the nullPointerException.

PR #851, opened 2017-06-02
Fixed the generated C# for const ref parameter of Indexer

### LINKED ISSUE #628: Error generating code for const ref subscript operator
Add to `Common.h` TestIndexedProperties:

``` cpp
    int operator[](const int& key);
```

Error:

```
../gen/Common/Common.cs(4932,24): error CS0213: You cannot use the fixed statement to take the address of an already fixed expression
../gen/Common/Common.cs(4932,46): error CS0266: Cannot implicitly convert type `int**' to `int*'. An explicit conversion exists (are you missing a cast?)
```

C# code:

``` csharp
        public int this[int* key]
        {
            get
            {
                fixed (int* __refParamPtr0 = &key)
                {
                    var arg0 = __refParamPtr0;
                    var __ret = Internal.OperatorSubscript_9((__Instance + __PointerAdjustment), arg0);
                    return __ret;
                }
            }
        }
```

mohtamohit:
Fixes #628

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpMarshal.cs
+++ b/src/Generator/Generators/CSharp/CSharpMarshal.cs
@@ -245,7 +245,7 @@ namespace CppSharp.Generators.CSharp
 
                 Context.Before.WriteLine("var {0} = {1};", ptrName,
                     Context.ReturnVarName);
-                
+
                 var res = $"{ptrName} == IntPtr.Zero? null : ({typedef})Marshal.GetDelegateForFunctionPointer({ptrName}, typeof({typedef}))";
                 Context.Return.Write(res);
                 return true;
@@ -492,7 +492,8 @@ namespace CppSharp.Generators.CSharp
 
             var pointee = pointer.Pointee.Desugar();
             if (Context.Function != null && pointer.IsPrimitiveTypeConvertibleToRef() &&
-                Context.MarshalKind != MarshalKind.VTableReturnValue)
+                Context.MarshalKind != MarshalKind.VTableReturnValue &&
+                Context.Function.OperatorKind != CXXOperatorKind.Subscript)
             {
                 var refParamPtr = string.Format("__refParamPtr{0}", Context.ParameterIndex);
                 var templateSubstitution = pointer.Pointee as TemplateParameterSubstitutionType;
@@ -580,13 +581,17 @@ namespace CppSharp.Generators.CSharp
                 if (isRefParam)
                 {
                     var typeName = Type.TypePrinterDelegate(finalPointee);
-
-                    if (param.IsInOut)
-                        Context.Before.WriteLine("{0} _{1} = {1};", typeName, param.Name);
+                    if (Context.Function.OperatorKind == CXXOperatorKind.Subscript)
+                        Context.Return.Write(param.Name);
                     else
-                        Context.Before.WriteLine("{0} _{1};", typeName, param.Name);
+                    {
+                        if (param.IsInOut)
+                            Context.Before.WriteLine($"{typeName} _{param.Name} = {param.Name};");
+                        else
+                            Context.Before.WriteLine($"{typeName} _{param.Name};");
 
-                    Context.Return.Write("&_{0}", param.Name);
+                        Context.Return.Write($"&_{param.Name}");
+                    }
                 }
                 else
                 {
--- a/tests/CSharp/CSharp.Tests.cs
+++ b/tests/CSharp/CSharp.Tests.cs
@@ -834,4 +834,14 @@ public unsafe class CSharpTests : GeneratorTestFixture
             complexArrayElement.Dispose();
         }
     }
+
+    [Test]
+    public void TestConstRefIndexer()
+    {
+        using (var indexproperty = new TestIndexedProperties())
+        {
+            int a = 2;
+            Assert.That(indexproperty[&a], Is.EqualTo(2));
+        }
+    }
 }
--- a/tests/CSharp/CSharp.h
+++ b/tests/CSharp/CSharp.h
@@ -1193,3 +1193,14 @@ struct HasComplexArray
 {
     ComplexArrayElement complexArray[ARRAY_LENGTH_MACRO];
 };
+
+class DLL_API TestIndexedProperties
+{
+public:
+    int operator[](const int& key);
+};
+
+int TestIndexedProperties::operator[](const int& key)
+{
+    return key;
+}


src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -492,7 +492,7 @@ public override bool VisitPointerType(PointerType pointer, TypeQualifiers quals)
 
             var pointee = pointer.Pointee.Desugar();
             if (Context.Function != null && pointer.IsPrimitiveTypeConvertibleToRef() &&
-                Context.MarshalKind != MarshalKind.VTableReturnValue)
+                Context.MarshalKind != MarshalKind.VTableReturnValue && Context.Function.OperatorKind != CXXOperatorKind.Subscript)
```
ddobrev: Move the new condition to a new line - notice that this line is too long compared to the others.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -507,7 +507,7 @@ public override bool VisitPointerType(PointerType pointer, TypeQualifiers quals)
                 else
                 {
                     Context.Before.WriteLine(
-                        $"fixed ({pointer} {refParamPtr} = &{Context.Parameter.Name})");
+                        $"fixed ({pointer} {refParamPtr} = &{Context.Parameter.Name})");
```
ddobrev: Remove the space at the end.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -534,7 +534,7 @@ public override bool VisitPointerType(PointerType pointer, TypeQualifiers quals)
                 {
                     Context.Return.Write(MarshalStringToUnmanaged(Context.Parameter.Name));
                     Context.Cleanup.WriteLine("Marshal.FreeHGlobal({0});", Context.ArgName);
-                }
+                }
```
ddobrev: Remove the space at the end.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -580,13 +580,16 @@ public override bool VisitPointerType(PointerType pointer, TypeQualifiers quals)
                 if (isRefParam)
                 {
                     var typeName = Type.TypePrinterDelegate(finalPointee);
-
-                    if (param.IsInOut)
-                        Context.Before.WriteLine("{0} _{1} = {1};", typeName, param.Name);
+                    if (Context.Function.OperatorKind != CXXOperatorKind.Subscript)
```
ddobrev: Positive conditions are a little faster to understand when reading code so invert the conditional.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -580,13 +580,16 @@ public override bool VisitPointerType(PointerType pointer, TypeQualifiers quals)
...
+                    {
+                        if (param.IsInOut)
+                            Context.Before.WriteLine("{0} _{1} = {1};", typeName, param.Name);
```
ddobrev: Now that you're changing this anyway, convert to interpolated strings this line and the other 3 in the changed block.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -580,13 +580,16 @@ public override bool VisitPointerType(PointerType pointer, TypeQualifiers quals)
...
+                        else
+                            Context.Before.WriteLine("{0} _{1};", typeName, param.Name);
+                        Context.Return.Write("&_{0}", param.Name);
+                    }
                     else
-                        Context.Before.WriteLine("{0} _{1};", typeName, param.Name);
-
-                    Context.Return.Write("&_{0}", param.Name);
+                        Context.Return.Write("{0}", param.Name);
```
ddobrev: Just write the name, no need for either formatting or interpolation.

tests/CSharp/CSharp.Tests.cs
```diff
@@ -834,4 +834,16 @@ public void TestGenerationOfArraySetter()
...
\ No newline at end of file
+
+    [Test]
+    public void TestConstRefIndexer()
+    {
+        using (var indexproperty = new TestIndexedProperties())
+        {
+            int a = 2;
+            int* b = &a;
+            int* c = &a;
```
ddobrev: No need for a second variable with the same value.

tests/CSharp/CSharp.Tests.cs
```diff
@@ -834,4 +834,16 @@ public void TestGenerationOfArraySetter()
...
-}
\ No newline at end of file
+
+    [Test]
+    public void TestConstRefIndexer()
+    {
+        using (var indexproperty = new TestIndexedProperties())
+        {
+            int a = 2;
+            int* b = &a;
```
ddobrev: No need for a local variable, you can just write &a in the indexer.

tests/CSharp/CSharp.Tests.cs
```diff
@@ -834,4 +834,16 @@ public void TestGenerationOfArraySetter()
...
+            int* c = &a;
+            Assert.AreEqual(indexproperty[b], indexproperty[c]);
```
ddobrev: Since b and c have the same value, you compare two identical values which is pointless. Compare to a hard-coded value instead.
In addition, please use the Assert.That(... Is.EqualTo(...)) syntax because we have long decided to use it over the old one.

tests/CSharp/CSharp.h
```diff
@@ -1193,3 +1193,14 @@ struct HasComplexArray
 {
     ComplexArrayElement complexArray[ARRAY_LENGTH_MACRO];
 };
+
+class DLL_API TestIndexedProperties
+{
+public:
+	int operator[](const int& key);
```
ddobrev: You have used a tab here while we use 4 spaces in C++ too. You could change your VS settings in order not to fix it manually each time.

tests/CSharp/CSharp.h
```diff
@@ -1193,3 +1193,14 @@ struct HasComplexArray
...
+};
+
+int TestIndexedProperties::operator[](const int& key)
+{
+	return key;
```
ddobrev: Tab -> 4 spaces.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -580,13 +581,18 @@ public override bool VisitPointerType(PointerType pointer, TypeQualifiers quals)
                 if (isRefParam)
                 {
                     var typeName = Type.TypePrinterDelegate(finalPointee);
-
-                    if (param.IsInOut)
-                        Context.Before.WriteLine("{0} _{1} = {1};", typeName, param.Name);
+                    if (Context.Function.OperatorKind == CXXOperatorKind.Subscript)
+                    {
```
ddobrev: We do not use braces with single statements.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -580,13 +581,18 @@ public override bool VisitPointerType(PointerType pointer, TypeQualifiers quals)
...
+                        Context.Return.Write("key");
```
ddobrev: You have forgotten to replace your test code with the real code.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -580,13 +581,18 @@ public override bool VisitPointerType(PointerType pointer, TypeQualifiers quals)
...
                     else
-                        Context.Before.WriteLine("{0} _{1};", typeName, param.Name);
-
-                    Context.Return.Write("&_{0}", param.Name);
+                    {
+                        if (param.IsInOut)
+                            Context.Before.WriteLine($"{typeName} _{param.Name} = {param.Name};");
+                        else
+                            Context.Before.WriteLine($"{typeName} _{param.Name};");
+                        Context.Return.Write($"&_{param.Name}");
```
ddobrev: Restore the empty line before this statement.

PR #858, opened 2017-06-07
Added int128 and uint128 to the marshalling structures in the C# generator. 

### LINKED ISSUE #771: Handle all primitive types in internal layouts in the generated C#
We have the following piece in the C# generator:

    // we do not support the primitives below yet because their representation in C# is problematic
    if (coreType.IsPrimitiveType(PrimitiveType.LongDouble) ||
        coreType.IsPrimitiveType(PrimitiveType.Int128) ||
        coreType.IsPrimitiveType(PrimitiveType.UInt128) ||
        coreType.IsPrimitiveType(PrimitiveType.Half))
        return;

This is troublesome now that we have an option for generating sequential layouts. We need to find a way to represent such fields and include them in our internal layouts.

ktopouzi:


src/Generator/Generators/CSharp/CSharpTypePrinter.cs
```diff
@@ -512,8 +512,10 @@ static string GetIntString(PrimitiveType primitive, ParserTargetInfo targetInfo)
                 case PrimitiveType.LongLong:
                 case PrimitiveType.ULongLong:
                     return GetIntString(primitive, Context.TargetInfo);
-                case PrimitiveType.Int128: return "__int128";
-                case PrimitiveType.UInt128: return "__uint128_t";
+                case PrimitiveType.Int128: return new TypePrinterResult { Type = "fixed byte",
+                    NameSuffix = $"[{16}]"}; // The type is always 128 bits wide
```
ddobrev: No need for an interpolated string here, just "16".

src/Generator/Generators/CSharp/CSharpTypePrinter.cs
```diff
@@ -512,8 +512,10 @@ static string GetIntString(PrimitiveType primitive, ParserTargetInfo targetInfo)
...
+                case PrimitiveType.UInt128: return new TypePrinterResult { Type = "fixed byte",
+                    NameSuffix = $"[{16}]"}; // The type is always 128 bits wide
```
ddobrev: See above.

PR #864, opened 2017-06-14
Enabled dtor generation in basic_string

mohtamohit:


src/CppParser/Bindings/CSharp/i686-apple-darwin12.4.0/Std.cs
```diff
@@ -1227,6 +1227,24 @@ namespace Std
 {
     namespace __1
     {
+        namespace Map
```
ddobrev: Revert this entire piece, it's generated because of an unrelated bug.

src/CppParser/Bindings/CSharp/x86_64-apple-darwin12.4.0/Std.cs
```diff
@@ -1227,6 +1227,24 @@ namespace Std
 {
     namespace __1
     {
+        namespace Map
```
ddobrev: Revert this entire piece, it's generated because of an unrelated bug.

PR #876, opened 2017-06-27
Manipulated comments to avoid duplication of summary block.

ktopouzi:


tritao:
This should not be done in the AST converter layer, since it modifies the contents of the original code and might not be suitable for people just using the AST layer to parse native code.

@ddobrev any idea where we should move this to? I was thinking a pass but I don't see a suitable existing pass for it and don't think it deserves a new one.

ddobrev:
@tritao I was actually imagining precisely a pass. It's basically a hack because of a Clang bug, the pieces of the comment come garbled from Clang itself.

tritao:
@ktopouzi Like @ddobrev said, please create a new pass, you can name it `CleanCommentsPass` and move the clean up code to be done inside.

src/Parser/ASTConverter.cs
```diff
@@ -2063,6 +2063,13 @@ protected override AST.Comment VisitHTMLEndTagComment(HTMLEndTagComment comment)
 
         protected override AST.Comment VisitTextComment(TextComment comment)
         {
+            if (comment.Text.Contains(">") || comment.Text.Contains("<"))
```
tritao: Move to pass.

src/Generator/Generators/CodeGenerator.cs
```diff
@@ -32,7 +32,7 @@ public virtual CommentKind CommentKind
             get
             {
                 if (!Options.CommentKind.HasValue)
-                    return CommentKind.BCPL;
```
ddobrev: Revert this, instead, pass DocumentationCommentKind to GenerateComment.

src/Generator.Tests/Passes/TestPasses.cs
```diff
@@ -63,6 +63,25 @@ public void TestFunctionToStaticPass()
...
+
+            passBuilder.AddPass(new CleanCommentsPass());
+            passBuilder.RunPasses(pass => pass.VisitDeclaration(c));
+
+            var para = (ParagraphComment)c.Comment.FullComment.Blocks[0];
+            System.Text.StringBuilder comment = new System.Text.StringBuilder();
+
+            foreach (CppSharp.AST.TextComment item in para.Content)
+            {
+                comment.Append(item.Text);
```
ddobrev: Do not construct the comment like this. Instead, use our CommentPrinter.

src/Generator.Tests/Passes/TestPasses.cs
```diff
@@ -63,6 +63,25 @@ public void TestFunctionToStaticPass()
...
+            }
+
+            Assert.That(comment.ToString(), Is.EqualTo(" A simple test. "));
```
ddobrev: You have a leading space in the expected value but not in the actual one.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,31 @@
...
+                {
+                    if (item.Text.Contains("<") || item.Text.Contains(">"))
+                        item.Text = item.Text.TrimStart('<', '>');
+                    if (item.Text.Contains("summary"))
+                        item.Text = "";
+                        
+                }
+                decl.Comment.FullComment.Blocks[0] = summaryPara;
+            }
+            return base.VisitDeclaration(decl);
```
ddobrev: As you can see in other passes, they usually start with:

    if (!base_call)
        return false;

This ensures already visited declarations are skipped. Do it here too.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace CppSharp.Passes
+{
+    public class CleanCommentsPass : TranslationUnitPass
+    { 
+        public override bool VisitDeclaration(Declaration decl)
+        {
+            if (decl.Comment != null)
+            {
+                var summaryPara = (ParagraphComment)decl.Comment.FullComment.Blocks[0];
```
tritao: @ddobrev Are we guaranteed this will always be true? Shouldn't we be using an `as` here and testing for true? Or better yet, use a comment visitor?

ddobrev: @tritao this needs rewriting rather than fixing. It had become too late so I advised @ktopouzi to just continue tomorrow.

ktopouzi: @tritao As i said to @ddobrev  it was my attempt to make this work. Tomorrow i will ensure it will work by changing it from top to bottom.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,98 @@
...
+{
+    public class CleanCommentsPass : TranslationUnitPass,ICommentVisitor<bool>
+    {
+        public bool VisitBlockCommand(BlockCommandComment comment)
+        {
+            return true;
+        }
+
+        public override bool VisitDeclaration(Declaration decl)
+        {
```
ktopouzi: @ddobrev While adding if (base.VisitDeclaration(decl)) return false; i wasn't able to enter the Visit. So i just skipped. If you want to add something else please let me know.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,101 @@
...
+            return true;
+        }
+
+        public bool VisitParagraphCommand(ParagraphComment comment)
+        {
+            foreach (TextComment item in comment.Content)
+            {
+                if (item.Text.StartsWith("<", StringComparison.Ordinal) 
+                    && !item.Text.EndsWith(">", StringComparison.Ordinal))
+                    if (item.Text.Contains("summary"))
```
ddobrev: We've already discussed that this should not be specific to "summary". It might be some other element. So please replace this check with something more general. For example, a regex which checks for a tag where a tag is a single word i.e. only Latin letters, digits and underscores and starts with a letter.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,128 @@
...
+                            break;
+                        else
+                        {
+                            if (com.Text.StartsWith("<", StringComparison.Ordinal))
+                                com.Text = $"{com.Text}{">"}";
+                            else if (com.Text.StartsWith(">", StringComparison.Ordinal))
+                                com.Text = com.Text.Substring(1);
+                            break;
+                        }
+                    case DocumentationCommentKind.FullComment:
```
ddobrev: No need to enumerate all types just to write a comment. Adding it to the default section suffices.

src/Generator/Utils/HtmlEncoder.cs
```diff
@@ -54,6 +54,11 @@ public static string HtmlEncode (string s)
 
 			if (s.Length == 0)
 				return String.Empty;
+
+            if (s.StartsWith("<", StringComparison.Ordinal)
```
ddobrev: Reuse the regex for tags here.

ddobrev: Also, move this to the comment printer. This piece is its job.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,128 @@
...
+        public bool VisitParagraphCommand(ParagraphComment comment)
+        {
+            Regex regex = new Regex(@"^(<|>).{0,}(|_|-)");
+            foreach (var item in comment.Content)
+            {
+                switch (item.Kind)
+                {
+                    case DocumentationCommentKind.TextComment:
+                        TextComment com = (TextComment)item;
+                        if (!regex.IsMatch(com.Text))
```
ddobrev: Invert the if, you have a break anyway.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        public bool VisitInlineCommand(InlineCommandComment comment)
+        {
+            return true;
+        }
+
+        public bool VisitParagraphCommand(ParagraphComment comment)
+        {
+            Regex regex = new Regex(@"^(<|>).{0,}(|_|-)");
```
ddobrev: This regex does not do what we want.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,115 @@
...
+                    case DocumentationCommentKind.TextComment:
+                        TextComment com = (TextComment)item;
+                        if (!regex.IsMatch(com.Text))
+                        {
+                            if (tag)
+                                com.Text = com.Text.Substring(1);
+                            break;
+                        }
+                        tag = true;
+                            if (com.Text.StartsWith("<", StringComparison.Ordinal))
```
ddobrev: This code is misaligned.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,115 @@
...
+        public bool VisitParagraphCommand(ParagraphComment comment)
+        {
+            Regex regex = new Regex(@"(^<|<\/)[a-zA-Z][\w\-]*>?");
+            bool tag = false;
+            foreach (var item in comment.Content)
+            {
+                switch (item.Kind)
+                {
+                    case DocumentationCommentKind.TextComment:
+                        TextComment com = (TextComment)item;
```
ddobrev: Add a space after the type cast.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,115 @@
...
+            return true;
+        }
+
+        public bool VisitParagraphCommand(ParagraphComment comment)
+        {
+            Regex regex = new Regex(@"(^<|<\/)[a-zA-Z][\w\-]*>?");
+            bool tag = false;
+            foreach (var item in comment.Content)
+            {
+                switch (item.Kind)
```
ddobrev: While we only handle one case, let's make it simpler with an "if" instead.

ddobrev: In face, just use LINQ Where in the loop.

src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -89,10 +89,17 @@ private static void GetCommentSections(this Comment comment, List<Section> secti
 
         private static string GetText(Comment comment, bool trim = false)
         {
+            System.Text.RegularExpressions.Regex regex =
```
ddobrev: Do not copy the regex like this. Extract it to a public variable somewhere instead and reuse it.

src/Generator/Generators/CodeGenerator.cs
```diff
@@ -393,6 +393,7 @@ public virtual bool VisitTypeAliasTemplateDecl(TypeAliasTemplate typeAliasTempla
 
     public static class Helpers
     {
+        public static System.Text.RegularExpressions.Regex Regex = new System.Text.RegularExpressions.Regex(@"^(<|</)[a-zA-Z][\w\-]*?>?$");
```
ddobrev: Remove the qualified name, use a using instead.

ddobrev: Rename the field to describe what this regex is for.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        public bool VisitInlineCommand(InlineCommandComment comment)
+        {
+            return true;
+        }
+
+        public bool VisitParagraphCommand(ParagraphComment comment)
+        {
+            Regex regex = new Regex(@"^(<|</)[a-zA-Z][\w\-]*?>?$");
```
ddobrev: You have already declared an accessible regex but you still have a copy here, why? The whole point was not to have a repetition.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,107 @@
...
+            bool tag = false;
+            foreach (var item in comment.Content.Where(c=> c.Kind == DocumentationCommentKind.TextComment))
+            {
+                        TextComment com = (TextComment) item;
```
ddobrev: Fix your indentation.

ddobrev: Add a space before =>.

ddobrev: Add a space before =>.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,107 @@
...
+                        if (!regex.IsMatch(com.Text))
+                        {
+                            if (tag)
```
ddobrev: Merge the if-s.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,105 @@
...
+            return true;
+        }
+
+        public bool VisitParagraphCommand(ParagraphComment comment)
+        {
+            bool tag = false;
+            foreach (var item in comment.Content.Where(c=> c.Kind == DocumentationCommentKind.TextComment))
+            {
+                TextComment com = (TextComment) item;
+                if (!Generators.Helpers.RegexTag.IsMatch(com.Text) && tag)
```
ddobrev: Now that you use an else, is && the same as a nested if? Think about it and fix if necessary.

ddobrev: Since I am pretty sure the correct path now that you have an else if the old one - if (regex) { if (tag) } invert the if so that you have a positive condition.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,105 @@
...
+                    com.Text = com.Text.Substring(1);
+                else
+                tag = true;
```
ddobrev: Indentation. Since you seem to often have problems with it, why don't you just auto-format?

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,106 @@
...
+        public bool VisitParagraphCommand(ParagraphComment comment)
+        {
+            bool tag = false;
+            foreach (var item in comment.Content.Where(c => c.Kind == DocumentationCommentKind.TextComment))
+            {
+                TextComment com = (TextComment)item;
+                if (Generators.Helpers.RegexTag.IsMatch(com.Text))
+                    tag = true;
+                else
+                    if (tag)
```
ddobrev: Move to the line above i.e. "else if (tag)"

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            return true;
+        }
+
+        public bool VisitParagraphCommand(ParagraphComment comment)
+        {
+            bool tag = false;
+            foreach (var item in comment.Content.Where(c => c.Kind == DocumentationCommentKind.TextComment))
+            {
+                TextComment com = (TextComment)item;
```
ddobrev: Missing a space after the cast. If you used auto-formatting, change your C# settings.

PR #881, opened 2017-07-02
Included the native AST.h to the solution

### LINKED ISSUE #850: Include the AST.h to non-compiled files.
When trying to add a new test in AST.h, I realized it is not the header that Test.AST.cs uses. but an external dependency that CppSharp.Parse.CLI has, while the correct one is under Tests/Native. So include the file to non-complied in AST test project may avoid confusion.

ktopouzi:
Fixes #850

### DIFF
--- a/src/Generator.Tests/premake5.lua
+++ b/src/Generator.Tests/premake5.lua
@@ -6,12 +6,17 @@ project "CppSharp.Generator.Tests"
   files { "**.cs" }
   vpaths { ["*"] = "*" }
 
-  libdirs 
+  libdirs
   {
     depsdir .. "/NUnit",
     depsdir .. "/NSubstitute"
   }
-  
+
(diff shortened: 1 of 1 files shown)


src/Generator.Tests/premake5.lua
```diff
@@ -6,12 +6,17 @@ project "CppSharp.Generator.Tests"
   files { "**.cs" }
   vpaths { ["*"] = "*" }
 
-  libdirs 
+  libdirs
```
ktopouzi: I don't know why that happened but i can restore the empty space if you want. I really don't know if it is necessary or not.

src/Generator.Tests/premake5.lua
```diff
@@ -6,12 +6,17 @@ project "CppSharp.Generator.Tests"
...
   {
     depsdir .. "/NUnit",
     depsdir .. "/NSubstitute"
   }
-  
+
+  files { testsdir .. "/Native/AST.h" }
```
ddobrev: Now that you're doing this, you might as well add Passed.h too.

PR #882, opened 2017-07-03
Fixing printing of comments for C#

realvictorprm:
Signed-off-by: realvictorprm <mueller.vpr@gmail.com>


realvictorprm:
This is much better in my opinion. I don't understand why we used `<param>` always before.

tritao:
I think the `<param>` stuff is to force VS to break new lines when showing Intellisense tooltips. @ddobrev will know better.

realvictorprm:
@tritao Need to check this again, however we finally have no remarks anymore where summaries should be.

realvictorprm:
@tritao yes para forces intellisense to use a newline ... hm let's remove at least this change.

realvictorprm:
I need to go over this again

PR #883, opened 2017-07-07
Fixes Stack mismatch when bool param passed from C++ to C#

### LINKED ISSUE #834: Stack mismatch when bool param passed from C++ to C#

mohtamohit:
Fixes #834

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpSources.cs
+++ b/src/Generator/Generators/CSharp/CSharpSources.cs
@@ -1509,8 +1509,10 @@ namespace CppSharp.Generators.CSharp
                     ParameterIndex = i
                 };
 
+                ctx.PushMarshalKind(MarshalKind.GenericDelegate);
                 var marshal = new CSharpMarshalNativeToManagedPrinter(ctx) { MarshalsParameter = true };
                 param.Visit(marshal);
+                ctx.PopMarshalKind();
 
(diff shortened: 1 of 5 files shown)


tritao:
Hey Mohit,

Can you also extend the managed test code to call the new native method/class you added?

mohtamohit:
Yes. I updated it accordingly.

ddobrev:
Merged as 1cbddb76eaf72b2071ee45a48f698d6c43453233.

PR #886, opened 2017-07-14
Added passage of VS Version selected at configuration time to the build

### LINKED ISSUE #814: Building for Windows x86 fails, .Gen projects use VS2017 includes when built with VS2015
##### Brief Description
I've followed the build instructions for Windows from the documentation, but when trying to compile I get a lot of "constexpr variable '_Memcpy_move_offset' must be initialized by a constant expression" errors in xstring. I've noticed that the file in question is located within the Visual Studio 2017 folder, even though I want to use (and have generated a 2015 project).
Is there a way to choose the MSVC version manually or is the problem I'm having completely unrelated to that?

OS: Windows 10.14393.969

Target: MSVC

### LINKED ISSUE #815: The VS version selected at configuration time is not passed to the build
Our code always uses the latest version of Visual Studio installed on the computer regardless of the selection at configuration time. We must somehow pass the correct version and then pass the correct value to SetupMSVC.

mohtamohit:
Fixes #815
Fixes #814

### DIFF
--- a/src/Parser/ParserOptions.cs
+++ b/src/Parser/ParserOptions.cs
@@ -14,8 +14,25 @@ namespace CppSharp.Parser
         public bool IsMicrosoftAbi { get { return Abi == CppAbi.Microsoft; } }
 
         /// Sets up the parser options to work with the given Visual Studio toolchain.
-        public void SetupMSVC(VisualStudioVersion vsVersion = VisualStudioVersion.Latest)
+        public void SetupMSVC()
         {
+            VisualStudioVersion vsVersion = VisualStudioVersion.Latest;
+            switch (BuildConfig.Choice)
+            {
+                case "vs2012":
+                    vsVersion = VisualStudioVersion.VS2012;
+                    break;
+                case "vs2013":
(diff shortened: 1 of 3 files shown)


src/Parser/ParserOptions.cs
```diff
@@ -16,6 +16,24 @@ public ParserOptions()
         /// Sets up the parser options to work with the given Visual Studio toolchain.
         public void SetupMSVC(VisualStudioVersion vsVersion = VisualStudioVersion.Latest)
         {
+            switch(BuildConfig.VisualStudioVersionSelected)
```
ddobrev: Add a space.

src/Parser/ParserOptions.cs
```diff
@@ -16,6 +16,24 @@ public ParserOptions()
         /// Sets up the parser options to work with the given Visual Studio toolchain.
         public void SetupMSVC(VisualStudioVersion vsVersion = VisualStudioVersion.Latest)
```
ddobrev: Now that we override the VS version in the switch below, we don't need a parameter. Turn it into a local.

PR #887, opened 2017-07-14
Fixed the generated #includes when renaming output files.

### LINKED ISSUE #868: Bad includes from CLI Generator when renaming outputs with GenerateName
OS: Windows
Target: MSVC

##### Brief Description
I was using a the GenerateName property of Driver.Options to provide a function that modified the names of my output files (adding "Wpr" to the end of the name in this case) to make it easier to identify the automatically generated wrappers. This worked until I tried to generate a wrapper for a class that made use of another wrapped class. The cpp file for the dependent wrapper tried to #include the other wrapper on it's original rather than modified name.

Please let me know if you need more information to reproduce this.

ktopouzi:
Fixes #868

### DIFF
--- a/build/Tests.lua
+++ b/build/Tests.lua
@@ -18,11 +18,11 @@ function SetupExampleProject()
   SetupParser()
 end
 
-function SetupTestProject(name, extraFiles)
+function SetupTestProject(name, extraFiles, suffix)
   SetupTestGeneratorProject(name)
   SetupTestNativeProject(name)
-  SetupTestProjectsCSharp(name, nil, extraFiles)
-  SetupTestProjectsCLI(name, extraFiles)
+  SetupTestProjectsCSharp(name, nil, extraFiles, suffix)
+  SetupTestProjectsCLI(name, extraFiles, suffix)
 end
 
 function SetupTestCSharp(name)
@@ -127,7 +127,14 @@ function LinkNUnit()
   }
 end
 
-function SetupTestProjectsCSharp(name, depends)
+function SetupTestProjectsCSharp(name, depends, extraFiles, suffix)
+    if suffix ~= nil then
+      nm = name .. suffix 
+      str = "Std" .. suffix
+    else
+      nm = name
+      str = "Std"
+    end
   project(name .. ".CSharp")
     SetupManagedTestProject()
 
@@ -136,8 +143,8 @@ function SetupTestProjectsCSharp(name, depends)
 
     files
     {
-      path.join(gendir, name, name .. ".cs"),
-      path.join(gendir, name, "Std.cs")
+      path.join(gendir, name, nm .. ".cs"),
+      path.join(gendir, name, str .. ".cs")
     }
     vpaths { ["*"] = "*" }
 
@@ -162,7 +169,7 @@ function SetupTestProjectsCSharp(name, depends)
     links { "CppSharp.Runtime" }
 end
 
-function SetupTestProjectsCLI(name, extraFiles)
+function SetupTestProjectsCLI(name, extraFiles, suffix)
   if not os.ishost("windows") then
     return
   end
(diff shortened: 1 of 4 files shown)


src/Generator/Generators/CLI/CLITypeReferences.cs
```diff
@@ -142,7 +142,7 @@ private void GenerateInclude(ASTRecord<Declaration> record)
         private string GetIncludePath(TranslationUnit translationUnit)
         {
             if (!DriverOptions.UseHeaderDirectories)
-                return translationUnit.FileName;
+                return $"{DriverOptions.GenerateName(translationUnit)}.h";
```
ddobrev: This must also work if GenerateName is not set i.e. is null.

tests/CLI/CLI.cs
```diff
@@ -13,6 +13,11 @@ public CLITestsGenerator(GeneratorKind kind)
 
         public override void Setup(Driver driver)
         {
+            driver.Options.GenerateName = file =>
```
ddobrev: I guess this is set because you changed it in such a way that it crashes if GenerateName is not set. But as indicated above, this is not correct. Remove this piece, it must work without it.

tests/Common/Common.cs
```diff
@@ -54,6 +54,11 @@ public override void Setup(Driver driver)
         {
             base.Setup(driver);
 
+            driver.Options.GenerateName = file =>
```
ddobrev: Remove.

tests/NamespacesDerived/NamespacesDerived.cs
```diff
@@ -14,6 +14,11 @@ public NamespacesDerivedTests(GeneratorKind kind)
 
         public override void Setup(Driver driver)
         {
+            driver.Options.GenerateName = file =>
```
ddobrev: Remove.

tests/StandardLib/StandardLib.cs
```diff
@@ -13,6 +13,11 @@ public StandardLibTestsGenerator(GeneratorKind kind)
 
         public override void Preprocess(Driver driver, ASTContext ctx)
         {
+            driver.Options.GenerateName = file =>
```
ddobrev: Remove.

tests/VTables/VTables.cs
```diff
@@ -19,7 +19,10 @@ public override void SetupPasses(Driver driver)
 
         public override void Preprocess(Driver driver, ASTContext ctx)
         {
-
+            driver.Options.GenerateName = file =>
```
ddobrev: Remove.

build/Tests.lua
```diff
@@ -18,11 +18,11 @@ function SetupExampleProject()
   SetupParser()
 end
 
-function SetupTestProject(name, extraFiles)
+function SetupTestProject(name, extraFiles, extension)
```
ddobrev: Rename "extension" to "suffix" because of the precise meaning "extension" has for files. Rename it everywhere.

build/Tests.lua
```diff
@@ -127,7 +127,14 @@ function LinkNUnit()
   }
 end
 
-function SetupTestProjectsCSharp(name, depends)
+function SetupTestProjectsCSharp(name, depends, extraFiles, extension)
+    if extension ~= nil then
+      nm = name .. extension 
+      str = "Std" .. extension
+    else
+      str = "Std"
```
ddobrev: Set the name first for consistency with the first branch of the conditional.

src/Generator/Generators/CLI/CLITypeReferences.cs
ddobrev: There are cases when headers have an extension other than ".h", even no extension at all. Handle this case in your logic.

src/Generator/Generators/CLI/CLITypeReferences.cs
```diff
@@ -142,7 +142,13 @@ private void GenerateInclude(ASTRecord<Declaration> record)
         private string GetIncludePath(TranslationUnit translationUnit)
         {
             if (!DriverOptions.UseHeaderDirectories)
-                return translationUnit.FileName;
+                if(DriverOptions.GenerateName != null)
```
ddobrev: Fix your indentation - if the indentation is correct, merge the if-s with &&. Add a space after the if.

PR #888, opened 2017-07-17
Added clang headers' path to system includes

### LINKED ISSUE #885: Error parsing with VS 2017
Using VS 2017 causes this error during parsing:
xstring(1905,26): error: constexpr variable '_Memcpy_move_offset' must be initialized by a constant expression

It's non-standard C++ which Clang cannot cope with. It might be a Clang bug. I have sent a report at https://bugs.llvm.org/show_bug.cgi?id=33735 . We'd best wait for their reply before deciding what to do.

mohtamohit:
Fixes #885

### DIFF
--- a/src/CppParser/Parser.cpp
+++ b/src/CppParser/Parser.cpp
@@ -340,6 +340,16 @@ void Parser::SetupHeader()
     if (opts->verbose)
         HSOpts.Verbose = true;
 
+#ifndef __APPLE__
+    // Initialize the default platform headers.
+    HSOpts.ResourceDir = GetClangResourceDir();
+
+    llvm::SmallString<128> ResourceDir(HSOpts.ResourceDir);
+    llvm::sys::path::append(ResourceDir, "include");
+    HSOpts.AddPath(ResourceDir.str(), clang::frontend::System, /*IsFramework=*/false,
+        /*IgnoreSysRoot=*/false);
+#endif
+
     for (unsigned I = 0, E = opts->IncludeDirs.size(); I != E; ++I)
     {
         const auto& s = opts->IncludeDirs[I];
@@ -363,14 +373,6 @@ void Parser::SetupHeader()
         const auto& undefine = opts->Undefines[I];
         PPOpts.addMacroUndef(undefine);
     }
-
-    // Initialize the default platform headers.
-    HSOpts.ResourceDir = GetClangResourceDir();
-
-    llvm::SmallString<128> ResourceDir(HSOpts.ResourceDir);
-    llvm::sys::path::append(ResourceDir, "include");
-    HSOpts.AddPath(ResourceDir.str(), clang::frontend::System, /*IsFramework=*/false,
-        /*IgnoreSysRoot=*/false);
 
 #ifdef _MSC_VER
     if (opts->microsoftMode)


src/Parser/ParserOptions.cs
```diff
@@ -68,6 +71,20 @@ public void SetupXcode()
 
         public void SetupIncludes()
         {
+            // Search for the available Clang versions
+            var location = System.Reflection.Assembly.GetExecutingAssembly().Location;
+            var a = Path.Combine(location, "..", "clang");
```
ddobrev: Make "a" descriptive.

src/Parser/ParserOptions.cs
```diff
@@ -68,6 +71,20 @@ public void SetupXcode()
...
+            var versions = Directory.EnumerateDirectories(a);
+            if (versions.Count() == 0)
+                throw new Exception("No valid Clang version found on system include paths");
+
+            string ClangVersionPath = versions.First();
```
ddobrev: Make the local variables start with a small letter.

src/Parser/ParserOptions.cs
```diff
@@ -68,6 +71,20 @@ public void SetupXcode()
 
         public void SetupIncludes()
         {
+            // Search for the available Clang versions
+            var location = System.Reflection.Assembly.GetExecutingAssembly().Location;
+            var a = Path.Combine(location, "..", "clang");
+            var versions = Directory.EnumerateDirectories(a);
```
ddobrev: You check for any directories, not just ones matching a version number.

src/Parser/ParserOptions.cs
```diff
@@ -68,6 +71,20 @@ public void SetupXcode()
 
         public void SetupIncludes()
         {
+            // Search for the available Clang versions
+            var location = System.Reflection.Assembly.GetExecutingAssembly().Location;
```
ddobrev: Import System.Reflection instead, we only use qualified names in the generated code or to avoid conflicts.

src/Parser/ParserOptions.cs
tritao: This needs to be guarded with an `Directory.Exists` call.

src/Parser/ParserOptions.cs
```diff
@@ -68,6 +71,20 @@ public void SetupXcode()
...
+            var a = Path.Combine(location, "..", "clang");
+            var versions = Directory.EnumerateDirectories(a);
+            if (versions.Count() == 0)
+                throw new Exception("No valid Clang version found on system include paths");
```
tritao: This is too agressive, needs to be at most a warning, not a fatal error.

tritao:
Clang provides a `ResourceDir` concept for these, check https://github.com/mono/CppSharp/blob/master/src/CppParser/Parser.cpp#L368.

Please check if passing the correct path to Clang's resource dir works. If it does, we should prefer that to adding it as a regular system include directory.

ddobrev:
@tritao has an excellent point, the resource dir should've done the job. It would be useful to know why it doesn't.

tritao:
I think it's not added because of https://github.com/mono/CppSharp/blob/master/src/CppParser/Parser.cpp#L396.

Here's what I suggest we do:

We add a new `ResourceDir` option to the native parser layer and we always pass the right path from managed code. This allows to get rid of `GetClangResourceDir` and `GetCurrentLibraryDir`. Then the parser can just add the resource directory with the already existing code.

So the parser layer will just be something like: `HSOpts.ResourceDir = opts->clangResourceDir;`.

PR #889, opened 2017-07-18
Added support for the Doxygen tag for making bold text.

ktopouzi:


src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -81,6 +81,17 @@ private static void GetCommentSections(this Comment comment, List<Section> secti
...
                     break;
                 case DocumentationCommentKind.InlineCommandComment:
+                    var lastSection = sections.Last();
+                    var inlineCommand = (InlineCommandComment) comment;
+
+                    if (inlineCommand.CommandKind == CommentCommandKind.B)
+                        if (lastSection.Lines.Count > 0)
+                            lastSection.Lines[lastSection.Lines.Count - 1] +=$" <c>{inlineCommand.Arguments[0].Text}</c>";
+                        else
+                        {
```
ddobrev: Remove {}.

src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -81,6 +81,17 @@ private static void GetCommentSections(this Comment comment, List<Section> secti
                 case DocumentationCommentKind.InlineContentComment:
                     break;
                 case DocumentationCommentKind.InlineCommandComment:
+                    var lastSection = sections.Last();
+                    var inlineCommand = (InlineCommandComment) comment;
+
+                    if (inlineCommand.CommandKind == CommentCommandKind.B)
+                        if (lastSection.Lines.Count > 0)
+                            lastSection.Lines[lastSection.Lines.Count - 1] +=$" <c>{inlineCommand.Arguments[0].Text}</c>";
```
ddobrev: Extract the string to a local variable and reuse it in both conditional branches.

src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -81,6 +81,15 @@ private static void GetCommentSections(this Comment comment, List<Section> secti
                 case DocumentationCommentKind.InlineContentComment:
                     break;
                 case DocumentationCommentKind.InlineCommandComment:
+                    var lastSection = sections.Last();
+                    var inlineCommand = (InlineCommandComment) comment;
+                    var argText = inlineCommand.Arguments[0].Text;
```
ddobrev: Move within the external if and include <c/> too because it's the same.

PR #890, opened 2017-07-18
 Added support for class templates which do not specialise types external to them

ddobrev:


src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -641,10 +642,16 @@ public override void GenerateClassSpecifier(Class @class)
...
-                    where @base.IsClass
-                    select @base.Class.Visit(TypePrinter).Type);
+                foreach (var @base in @class.Bases.Where(b => b.IsClass))
+                {
+                    var typeMaps = new List<System.Type>();
+                    var keys = new List<string>();
+                    this.DisableTypeMap(@base.Class, typeMaps, keys);
+                    var printedBase = @base.Type.Desugar().Visit(TypePrinter);
+                    bases.Add($"{printedBase}{printedBase.NameSuffix}");
+                    for (int i = 0; i < typeMaps.Count; i++)
```
tritao: Let's add a new property, `TypeMap.Enabled`, or `Active` so we do not have to mutate the lists while generating code. It's wasteful and not very elegant.

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -275,7 +275,8 @@ private void GenerateClassTemplateSpecializationInternal(Class classTemplate)
             if (generateNested)
             {
                 foreach (var nestedClass in classTemplate.Classes.Where(c => !c.IsDependent).Union(
-                    generated.First().Classes.Where(c => !c.IsDependent)))
+                    (generated.FirstOrDefault(s => !s.Ignore) ?? generated.First()).Classes.Where(
```
tritao: This nested usage of LINQ is hard to read. Can you maybe split up these nested LINQ into locals with a comment explaining what we're doing here?

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -932,26 +939,40 @@ private void GenerateIndexerSetter(Function function)
             function.Type.IsPointerTo(out type);
             PrimitiveType primitiveType;
             var internalFunction = GetFunctionNativeIdentifier(function);
-            var @internal = $@"{Helpers.InternalStruct}{
-                Helpers.GetSuffixForInternal(function.Namespace)}";
+            TypePrinter.PushContext(TypePrinterContextKind.Native);
```
tritao: Ever heard of new lines? :)

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -932,26 +939,40 @@ private void GenerateIndexerSetter(Function function)
...
             {
                 WriteLine($@"*{@internal}.{internalFunction}({
-                    GetInstanceParam(function)}, {function.Parameters[0].Name}) = value;");
+                    GetInstanceParam(function)}, {function.Parameters[0].Name}) = {
+                    marshal.Context.Return};");
             }
             else
             {
-                var typeString = type.ToString();
+                TypePrinter.PushContext(TypePrinterContextKind.Native);
```
tritao: I see this pattern a couple times.

Let's abstract it into a static method, `PrintNativeTypeName`, or something.

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -2401,14 +2424,15 @@ private void GenerateEqualsAndGetHashCode(Function method, Class @class)
...
                 if (@class.IsRefType)
                 {
-                    WriteLine("return this == obj as {0};", @class.Name);
+                    WriteLine($"return this == obj as {printedClass}{printedClass.NameSuffix};");
                 }
                 else
                 {
-                    WriteLine("if (!(obj is {0})) return false;", @class.Name);
-                    WriteLine("return this == ({0}) obj;", @class.Name);
+                    WriteLine($"if (!(obj is {printedClass}{printedClass.NameSuffix})) return false;");
```
tritao: I don't get why we need `NameSuffix` at all for these cases.

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -2417,22 +2441,23 @@ private void GenerateEqualsAndGetHashCode(Function method, Class @class)
...
+                    WriteLine($"return {Helpers.InstanceIdentifier}.GetHashCode();");
                 WriteCloseBraceIndent();
             }
         }
 
+        private void GenerateGetHashCode(Class @class)
+        {
+            WriteLine($"if ({Helpers.InstanceIdentifier} == global::System.IntPtr.Zero)");
+            WriteLineIndent("return global::System.IntPtr.Zero.GetHashCode();");
+            TypePrinter.PushContext(TypePrinterContextKind.Native);
```
tritao: Same here, use `PrintNativeTypeName`.

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -2768,9 +2794,17 @@ into context
                 // Special case for indexer - needs to dereference if the internal
                 // function is a pointer type and the property is not.
                 if (retType.Type.IsAddress() &&
-                    retType.Type.GetPointee().Equals(returnType) &&
-                    returnType.IsPrimitiveType())
-                    WriteLine("return *{0};", marshal.Context.Return);
```
tritao: I know this is existing code, but this is doing the job of the marshaler. We should move it to marshal if possible.

src/Generator/Generators/CSharp/CSharpSources.cs
ddobrev: This cannot work. TypeMaps is a list of types, not of TypeMap-s. We cannot control this through such a property. We cannot use a static one either because it would enable or disable all type maps rather than just the one in question.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -358,6 +358,13 @@ private string HandleReturnedPointer(Class @class, string qualifiedClass)
                 Context.Before.WriteLine("else {0} = {1}.{2}({3});", ret, qualifiedClass,
                     Helpers.CreateInstanceIdentifier, Context.ReturnVarName);
             }
+            Class returnedClass;
+            var pointee = Context.ReturnType.Type.Desugar().GetFinalPointee().Desugar();
```
ddobrev: This cannot work either because of typedef-ed function pointers. For those we use the name of the typedef for the name of the delegate and such built-in desugaring would always eliminate that name.

src/Generator/Generators/CSharp/CSharpSources.cs
ddobrev: Done.

src/Generator/Generators/CSharp/CSharpSources.cs
ddobrev: Done.

src/Generator/AST/Utils.cs
```diff
@@ -90,8 +90,16 @@ public static bool CheckIgnoreProperty(Property prop)
         {
             type = type.Desugar();
             type = (type.GetFinalPointee() ?? type).Desugar();
-            ClassTemplateSpecialization specialization;
-            type.TryGetDeclaration(out specialization);
+            ClassTemplateSpecialization specialization = null;
+            Declaration declaration;
+            if (type.TryGetDeclaration(out declaration))
+            {
+                do
```
ddobrev: Done.

src/Generator/Generators/CSharp/CSharpSources.cs
ddobrev: Done, I was actually able to simplify quite some code.

src/Generator/Generators/CSharp/CSharpSources.cs
ddobrev: Done.

src/Generator/Generators/CSharp/CSharpMarshal.cs
```diff
@@ -732,7 +748,7 @@ private void MarshalRefClass(Class @class)
             }
 
             string param = Context.Parameter.Name;
-            Type type = Context.Parameter.Type.Desugar();
+            Type type = Context.Parameter.Type.Desugar(false);
```
tritao: Add a new overload for this template Desugar case, or add a suitable prefix.

src/Generator/Generators/CSharp/CSharpTypePrinter.cs
```diff
@@ -339,13 +339,25 @@ public static bool IsConstCharString(QualifiedType qualType)
...
             if (!TypeMapDatabase.FindTypeMap(template, out typeMap))
             {
                 if (ContextKind == TypePrinterContextKind.Managed &&
                     decl == template.Template.TemplatedDecl)
-                    return $@"{decl.Visit(this)}<{string.Join(", ",
-                        template.Arguments.Select(VisitTemplateArgument))}>";
-                return decl.Visit(this);
+                {
+                    // HACK: TemplateParameterType.Parameter is null in some corner cases, see the parser
+                    if (template.Arguments.All(a =>
```
tritao: Please factor the lambda into an helper.

src/Generator/Passes/CheckIgnoredDecls.cs
```diff
@@ -39,9 +39,27 @@ public override bool VisitClassDecl(Class @class)
             if (!@class.IsDependent)
                 return false;
 
-            // templates are not supported yet
-            foreach (var specialization in @class.Specializations.Where(s => !s.IsExplicitlyGenerated))
-                specialization.ExplicitlyIgnore();
+            if (Options.GenerateClassTemplates)
```
tritao: Please make this one a little bit nicer.

src/Generator/Generators/CSharp/CSharpSources.cs
tritao: As we discussed, add a `PrintTemplateName` or similar for this.

ddobrev: Done, much nicer indeed.

src/Generator/Generators/CSharp/CSharpTypePrinter.cs
ddobrev: Done.

src/Generator/Passes/CheckIgnoredDecls.cs
ddobrev: Done.

src/Generator/Generators/CSharp/CSharpMarshal.cs
ddobrev: Done.

PR #891, opened 2017-07-18
Making the code run properly from a different working directory

mohtamohit:


src/CppParser/Parser.cpp
```diff
@@ -340,15 +326,13 @@ void Parser::SetupHeader()
     if (opts->verbose)
         HSOpts.Verbose = true;
 
-#ifndef __APPLE__
```
ddobrev: Restore the guard, macOS would still break otherwise.

PR #897, opened 2017-07-20
Moved the arg to enable clang compiler frontend options from native to managed code

mohtamohit:


tritao:
I don't think it's worth it to add a new option for this.

What do you think @ddobrev?

ddobrev:
@tritao since I didn't know what the option did, I asked @mohtamohit to check and add the respective property. He did. Now that I can see it's for enabling essentially Clang itself which we always need anyway, I tend to agree we'd better just leave it hard-coded.

PR #900, opened 2017-07-24
Reused the walking of declarations when getting name-spaces in the parser.

ktopouzi:


src/CppParser/Parser.cpp
```diff
@@ -1792,7 +1792,8 @@ DeclarationContext* Parser::GetNamespace(const clang::Decl* D,
         case Decl::Record:
         {
             auto RD = cast<RecordDecl>(Ctx);
-            DC = WalkRecord(RD);
+            auto W = WalkDeclaration(RD);
```
ddobrev: This whole structure saves nothing. The point is to get less code compared to before i.e:

case 1:
case 2:
case 3:
case 4:
    DC = WalkDecl();
    break;

src/CppParser/Parser.cpp
```diff
@@ -1804,19 +1805,22 @@ DeclarationContext* Parser::GetNamespace(const clang::Decl* D,
         case Decl::ClassTemplateSpecialization:
         {
             auto CTSpec = cast<ClassTemplateSpecializationDecl>(Ctx);
-            DC = WalkClassTemplateSpecialization(CTSpec);
+            auto W = WalkDeclaration(CTSpec);
```
ddobrev: "W" is not a good name, use, for example, "Decl".

PR #901, opened 2017-07-25
Fixed code generation for indexers returning a void pointer

mohtamohit:


tests/CSharp/CSharp.Tests.cs
```diff
@@ -844,4 +844,14 @@ public void TestConstRefIndexer()
...
         }
     }
+
+    [Test]
+    public void TestVoidPtrReturningIndexer()
+    {
+        using (var indexproperty = new TestIndexedProperties())
+        {
+            uint n = 21;
+            Assert.That(*((int *)indexproperty[n]), Is.EqualTo(21));
```
ddobrev: Remove the space before * and add one after the cast.

tests/CSharp/CSharp.h
```diff
@@ -1198,14 +1198,22 @@ struct HasComplexArray
 class DLL_API TestIndexedProperties
 {
 public:
+    mutable int a;
     int operator[](const int& key);
+    void *operator[](size_t n) const;
```
ddobrev: Move the space from before to after *.

tests/CSharp/CSharp.h
```diff
@@ -1198,14 +1198,22 @@ struct HasComplexArray
 class DLL_API TestIndexedProperties
 {
 public:
+    mutable int a;
```
ddobrev: Rename to "field".

PR #902, opened 2017-07-26
Expanded comments to convert some of the Doxygen tags to .Net compatible

ktopouzi:


tritao:
This needs tests.

src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -25,15 +25,49 @@ private static void GetCommentSections(this Comment comment, List<Section> secti
...
+                                blockCommandComment.ParagraphComment.GetCommentSections(sections);
+                                break;
+                            case CommentCommandKind.Sa:
+                                break;
+                            case CommentCommandKind.Since:
+                                foreach (var inlineContentComment in blockCommandComment.ParagraphComment.Content)
+                                {
+                                    inlineContentComment.GetCommentSections(sections);
+                                    if (inlineContentComment.HasTrailingNewline)
+                                        sections.Last().NewLine();
```
ddobrev: Move sections.Last() to a local outside the loop, this way you calculate it each time.

src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -25,15 +25,49 @@ private static void GetCommentSections(this Comment comment, List<Section> secti
...
+
+                            case CommentCommandKind.Return:
+                                sections.Add(new Section(CommentElement.Returns));
+                                blockCommandComment.ParagraphComment.GetCommentSections(sections);
+                                break;
+                            case CommentCommandKind.Note:
+                                sections.Add(new Section(CommentElement.Remarks));
+                                blockCommandComment.ParagraphComment.GetCommentSections(sections);
+                                break;
+                            case CommentCommandKind.Sa:
```
ddobrev: If this case is not supported at present, remove it, we can add it later on.

src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -25,15 +25,49 @@ private static void GetCommentSections(this Comment comment, List<Section> secti
...
-                    if (blockCommandComment.CommandKind == CommentCommandKind.Return &&
-                        blockCommandComment.ParagraphComment != null)
+                    if (blockCommandComment.ParagraphComment != null)
                     {
-                        sections.Add(new Section(CommentElement.Returns));
-                        blockCommandComment.ParagraphComment.GetCommentSections(sections);
+                        switch (blockCommandComment.CommandKind)
+                        {
+                            case CommentCommandKind.Brief:
+                                if (sections.Count == 1)
```
ddobrev: This whole block can be simplified because comment sections are obtained in both cases.

src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -29,11 +29,36 @@ private static void GetCommentSections(this Comment comment, List<Section> secti
                     break;
                 case DocumentationCommentKind.BlockCommandComment:
                     var blockCommandComment = (BlockCommandComment) comment;
-                    if (blockCommandComment.CommandKind == CommentCommandKind.Return &&
-                        blockCommandComment.ParagraphComment != null)
+                    var lastBlockSection = sections.Last();
```
ddobrev: Nice but move within the conditional because it's not used elsewhere.

src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -29,11 +29,36 @@ private static void GetCommentSections(this Comment comment, List<Section> secti
...
+                                if (sections.Count > 1)
+                                    sections.Add(new Section(CommentElement.Summary));
+                                
+                                blockCommandComment.ParagraphComment.GetCommentSections(sections);
+                                break;
+                            case CommentCommandKind.Return:
+                                sections.Add(new Section(CommentElement.Returns));
+                                blockCommandComment.ParagraphComment.GetCommentSections(sections);
+                                break;
+                            case CommentCommandKind.Note:
```
ddobrev: Notes are not the same as remarks. Remarks is a poor name ms gave to what is actually a description, there's a single such element per comment. So unless there's some other element we can map notes too, just take their text.

PR #904, opened 2017-07-28
Ensured a single element for remarks in the generated XML documentation comments.

ktopouzi:


src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -138,6 +138,13 @@ private static string FormatComment(List<Section> sections, CommentKind kind)
         {
             var commentPrefix = Comment.GetMultiLineCommentPrologue(kind);
             var commentBuilder = new StringBuilder();
+
+            sections = sections.OrderBy(s => s.Type).ToList();
```
ddobrev: Use Sort instead so that you don't assign.

src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -138,6 +138,13 @@ private static string FormatComment(List<Section> sections, CommentKind kind)
...
+            var remarks = sections.Where(s => s.Type == CommentElement.Remarks).ToList();
+            if (remarks.Any())
+                remarks.First().GetLines().AddRange(remarks.Skip(1).SelectMany(s => s.GetLines()));
+            sections = sections.Except(remarks.Skip(1)).ToList();
```
ddobrev: Again, do not assign. Instead, just remove the unnecessary ones. You could use RemoveRange.

src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -138,6 +138,14 @@ private static string FormatComment(List<Section> sections, CommentKind kind)
         {
             var commentPrefix = Comment.GetMultiLineCommentPrologue(kind);
             var commentBuilder = new StringBuilder();
+
+            sections.Sort((x, y) => x.Type.CompareTo(y.Type));
+            var remarks = sections.Where(s => s.Type == CommentElement.Remarks).ToList();
+            if (remarks.Any())
+                remarks.First().GetLines().AddRange(remarks.Skip(1).SelectMany(s => s.GetLines()));
+            if (remarks.Count > 1)
+                sections.RemoveRange(sections.FindIndex(p => p.Type == CommentElement.Remarks) + 1, remarks.Count - 1);
```
ddobrev: IndexOf(remarks.First()) is simpler.

PR #908, opened 2017-08-03
Set the name-space of a parameter to its function.

ktopouzi:


src/CppParser/Parser.cpp
```diff
@@ -1799,6 +1799,12 @@ DeclarationContext* Parser::GetNamespace(const clang::Decl* D,
             DC = WalkRecordCXX(RD);
             continue;
         }
+        case Decl::Function:
```
ddobrev: Remove this piece, it does nothing within this commit.

PR #911, opened 2017-08-09
Printed the text from unsupported comment tags.

ktopouzi:


src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -29,11 +29,32 @@ private static void GetCommentSections(this Comment comment, List<Section> secti
                     break;
                 case DocumentationCommentKind.BlockCommandComment:
                     var blockCommandComment = (BlockCommandComment) comment;
-                    if (blockCommandComment.CommandKind == CommentCommandKind.Return &&
-                        blockCommandComment.ParagraphComment != null)
+                    if (blockCommandComment.ParagraphComment != null)
```
ddobrev: Exit early to reduce nesting.

src/Generator/Generators/CSharp/CSharpCommentPrinter.cs
```diff
@@ -29,11 +29,32 @@ private static void GetCommentSections(this Comment comment, List<Section> secti
...
                     {
-                        sections.Add(new Section(CommentElement.Returns));
-                        blockCommandComment.ParagraphComment.GetCommentSections(sections);
+                        var lastBlockSection = sections.Last();
```
ddobrev: Move this to the only case it's actually used in.

PR #912, opened 2017-08-10
Removed leftovers in the comments from unsupported custom xml tags.

ktopouzi:


src/Generator/Generators/CodeGenerator.cs
```diff
@@ -395,6 +395,7 @@ public virtual bool VisitTypeAliasTemplateDecl(TypeAliasTemplate typeAliasTempla
     public static class Helpers
     {
         public static Regex RegexTag = new Regex(@"^(<|</)[a-zA-Z][\w\-]*?>?$");
+        public static Regex RegexCustomTag = new Regex(@"^_\S*");
```
ddobrev: Do we really need to include the underscore? What if a custom element is named "thread-safety" i.e. with a dash? I think just \S should do.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -54,6 +54,16 @@ public bool VisitInlineCommand(InlineCommandComment comment)
 
         public bool VisitParagraphCommand(ParagraphComment comment)
         {
+            if (comment.Content.Any(k => k.Kind == DocumentationCommentKind.InlineCommandComment))
+            {
+                foreach (var item in comment.Content.Where(c => c.Kind == DocumentationCommentKind.TextComment))
```
ddobrev: I think this code checks all text comments rather than only those following a command. Shouldn't we only change the latter?

src/Generator/Generators/CodeGenerator.cs
ddobrev: This regex doesn't really match a tag but rather a part of a comment command. Please rename accordingly.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -54,6 +54,23 @@ public bool VisitInlineCommand(InlineCommandComment comment)
 
         public bool VisitParagraphCommand(ParagraphComment comment)
         {
+            for (int i = 0; i < comment.Content.Count; i++)
+            {
+                if (comment.Content[i].Kind == DocumentationCommentKind.InlineCommandComment)
+                {
+                    TextComment com = (TextComment) comment.Content[i + 1];
```
ddobrev: I think this code is safe but still, as we discussed, it would be really pathetic if we crashed because of a comment. So add guards for the bounds and the type.

src/Generator/Passes/CleanCommentsPass.cs
```diff
@@ -54,6 +54,17 @@ public bool VisitInlineCommand(InlineCommandComment comment)
 
         public bool VisitParagraphCommand(ParagraphComment comment)
         {
+            for (int i = 0; i < comment.Content.Count; i++)
+            {
+                if (comment.Content[i].Kind == DocumentationCommentKind.InlineCommandComment)
+                {
+                    if (comment.Content[i + 1] != null && comment.Content[i + 1].Kind == DocumentationCommentKind.TextComment)
```
ddobrev: The comment cannot be null, I meant check the bounds.
Also, shorten the line.

PR #913, opened 2017-08-15
Merged Delegate Pass and pass for Anonymous Delegates and added the f…

mohtamohit:
…ollowing improvements alongwith.

1. Got rid of Anonymous Names.
2. Merged the common code for Delegate Generation.
3. The delegate pass also works for C++/CLI now.
4. Fixed the calling conventions of delegates for both, C++/CLI and C#.
5. Ensures that if a delegate is used for a virtual as well as something else, it finally ends up as public.
6. Fixed the code generation when the return type of a method is a function pointer that has been used somewhere else as well.
7. Added Access and Calling convention to the delegate definition.

The only thing left is to get rid of the hack used, i.e move the code in VisitFunctionDecl to VisitParametersDecl. Somehow, it's not working right now.


src/Generator/Driver.cs
```diff
@@ -298,7 +298,7 @@ public void SetupPasses(ILibrary library)
             TranslationUnitPasses.AddPass(new CheckStaticClass());
             TranslationUnitPasses.AddPass(new MoveOperatorToClassPass());
             TranslationUnitPasses.AddPass(new MoveFunctionToClassPass());
-            TranslationUnitPasses.AddPass(new GenerateAnonymousDelegatesPass());
+            //TranslationUnitPasses.AddPass(new GenerateAnonymousDelegatesPass());
```
ddobrev: Remove commented code.

src/Generator/Generators/CLI/CLIHeaders.cs
```diff
@@ -745,31 +745,27 @@ public bool GenerateTypedef(TypedefNameDecl typedef)
...
-            if (typedef.Type.IsPointerTo(out function))
+            var functionType = typedef.Type as FunctionType;
+            if (typedef.Type.IsPointerTo(out functionType) || functionType != null)
             {
                 PushBlock(BlockKind.Typedef, typedef);
                 GenerateDeclarationCommon(typedef);
 
                 var insideClass = typedef.Namespace is Class;
 
                 var attributedType = typedef.Type.GetPointee() as AttributedType;
```
ddobrev: I know this is old code but I think we now check for attributed types when desugaring so just desugar instead.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -14,15 +15,21 @@ public class DelegatesPass : TranslationUnitPass
 
         public class DelegateDefinition
         {
-            public DelegateDefinition(string @namespace, string signature)
+            public DelegateDefinition(string @namespace, string signature, AccessSpecifier access, CallingConvention callingConvention)
```
ddobrev: Split the line.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -119,61 +241,114 @@ private void AddDelegatesToDictionary(Declaration decl, Module module)
...
                 return false;
-            
-            var module = method.TranslationUnit.Module;
-            AddDelegatesToDictionary(method, module);
+
+            if (Options.IsCSharpGenerator)
+            {
+                var module = method.TranslationUnit.Module;
+                method.FunctionType = AddDelegatesToDictionary(method.Namespace, method);
+                method.ReturnType = (method.ReturnType.Type.GetFinalPointee().Desugar() is FunctionType) ?
```
ddobrev: Better use a conditional here.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -35,33 +42,135 @@ public DelegatesPass()
...
-                namespacesDelegates[module].Namespace.Declarations.Add(namespacesDelegates[module]);
+                foreach (var module in Options.Modules.Where(m => namespacesDelegates.ContainsKey(m)))
+                    namespacesDelegates[module].Namespace.Declarations.Add(namespacesDelegates[module]);
+            }
 
             return result;
         }
 
-        private void AddDelegatesToDictionary(Declaration decl, Module module)
+        private QualifiedType AddDelegatesToDictionary(DeclarationContext @namespace, Declaration decl)
```
ddobrev: This method adds at most one delegate so why the plural? Wouldn't AddDelegateToDictionary make more sense?

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -35,33 +43,135 @@ public DelegatesPass()
             VisitOptions.VisitTemplateArguments = false;
         }
 
+        private struct Typedef
```
ddobrev: I can see the name-space of the stored typedef is the same as this context here. So having the typedef only is enough, remove this struct.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -35,33 +43,135 @@ public DelegatesPass()
...
+                typedefs = new List<Typedef>();
+                allTypedefs.Add(@namespace.QualifiedName, typedefs);
+            }
+
+            var typedef = FindMatchingTypedef(typedefs, functionType);
+            if (typedef == null)
+            {
+                for (int i = 0; i < functionType.Parameters.Count; i++)
+                {
+                    functionType.Parameters[i].Name = string.Format("_{0}", i);
```
ddobrev: Use an interpolated string.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -35,33 +43,135 @@ public DelegatesPass()
...
+            {
+                typedefs = new List<Typedef>();
+                allTypedefs.Add(@namespace.QualifiedName, typedefs);
+            }
+
+            var typedef = FindMatchingTypedef(typedefs, functionType);
+            if (typedef == null)
+            {
+                for (int i = 0; i < functionType.Parameters.Count; i++)
+                {
```
ddobrev: Remove the braces.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -35,33 +43,135 @@ public DelegatesPass()
...
+        /// iterating over it, so we collect all the typedefs and add them at the end.
+        /// </summary>
+        private readonly Dictionary<string, List<Typedef>> allTypedefs = new Dictionary<string, List<Typedef>>();
+
+        private QualifiedType CheckType(DeclarationContext @namespace, QualifiedType type, Declaration decl)
+        {
+            if (type.Type.IsDependent)
+                return type;
+
+            var pointerType = type.Type as PointerType;
```
ddobrev: Don't we need to desugar here?

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -35,33 +43,135 @@ public DelegatesPass()
...
+            if (pointerType == null)
+                return type;
+
+            var functionType = pointerType.Pointee as FunctionType;
```
ddobrev: Don't we need to desugar here?

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -35,33 +43,135 @@ public DelegatesPass()
...
 
-        private void AddDelegatesToDictionary(Declaration decl, Module module)
+        private QualifiedType AddDelegateToDictionary(DeclarationContext @namespace, Declaration decl)
         {
             CallingConvention callingConvention;
             bool isDependent = false;
             QualifiedType returnType;
             List<Parameter> @params;
-            if (decl is Method)
+            bool isVirtualMethod = false;
```
ddobrev: Not sure what isVirtualMethod is for. If it's a method, we have a delegate for a virtual function, if not, it's for a parameter so no method at all.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -119,61 +242,113 @@ private void AddDelegatesToDictionary(Declaration decl, Module module)
...
+                    existingDelegate.Access = AccessSpecifier.Public;
+                    if (libsDelegates.ContainsKey(module))
+                    {
+                        if (libsDelegates[module].ContainsKey(delegateString))
+                            libsDelegates[module].Remove(delegateString);
+                        libsDelegates[module].Add(delegateString, existingDelegate);
+                    }
+                    foreach (var dec in namespaceDelegates.Declarations.Where(m => m.Name == @delegate.Name))
+                    {
+                        namespaceDelegates.Declarations.Remove((TypedefDecl)dec);
```
ddobrev: This removal looks wrong. If you find a match, you just need to make it public if not already and change the existing one, why remove anything?

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -119,61 +242,113 @@ private void AddDelegatesToDictionary(Declaration decl, Module module)
...
+                    }
+
+                    namespaceDelegates.Declarations.Add(@delegate);
+                }
+                // Code to check if there is an existing delegate with a different calling convention
+                // Add the new delegate with calling convention appended to it's name
+                if (existingDelegate.CallingConvention != callingConvention)
+                {
+                    delegateString += callingConvention.ToString();
+                    existingDelegate = new DelegateDefinition(module.OutputNamespace, delegateString, access, callingConvention);
```
ddobrev: Split the line.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -119,61 +240,112 @@ private void AddDelegatesToDictionary(Declaration decl, Module module)
...
-                return;
+                // Code to ensure that if a delegate is used for a virtual as well as something else, it finally ends up as public
+                if (existingDelegate.Access == AccessSpecifier.Private && access == AccessSpecifier.Public)
+                {
+                    existingDelegate.Access = AccessSpecifier.Public;
+                    if (libsDelegates.ContainsKey(module))
+                    {
+                        if (libsDelegates[module].ContainsKey(delegateString))
+                            libsDelegates[module].Remove(delegateString);
+                        libsDelegates[module].Add(delegateString, existingDelegate);
```
ddobrev: Why remove the delegate and add it again?

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -119,61 +240,112 @@ private void AddDelegatesToDictionary(Declaration decl, Module module)
...
+                    }
+                    foreach (var dec in namespaceDelegates.Declarations.Where(m => m.Name == @delegate.Name))
+                    {
+                        ((TypedefDecl)dec).Access = AccessSpecifier.Public;
```
ddobrev: Why change the access if it's changed in the code above?

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -119,61 +240,112 @@ private void AddDelegatesToDictionary(Declaration decl, Module module)
...
+                        break;
+                    }
+                }
+                // Code to check if there is an existing delegate with a different calling convention
+                // Add the new delegate with calling convention appended to it's name
+                if (existingDelegate.CallingConvention != callingConvention)
+                {
+                    delegateString += callingConvention.ToString();
+                    existingDelegate = new DelegateDefinition(module.OutputNamespace,
```
ddobrev: Use a new local variable to avoid confusion.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -119,61 +240,112 @@ private void AddDelegatesToDictionary(Declaration decl, Module module)
...
+                        delegateString, access, callingConvention);
+                    @delegate.Name += callingConvention.ToString();
```
ddobrev: I assume this changes the name of the existing delegate. It's not necessary, they just need to be different. If one is called MyDelegate, and the other MyDelegateStd, it's enough, no need for the first one to be renamed to MyDelegateC.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -119,61 +213,111 @@ private void AddDelegatesToDictionary(Declaration decl, Module module)
...
             if (!base.VisitMethodDecl(method) || !method.IsVirtual || method.Ignore)
                 return false;
-            
-            var module = method.TranslationUnit.Module;
-            AddDelegatesToDictionary(method, module);
+
+            if (Options.IsCSharpGenerator)
+            {
+                var module = method.TranslationUnit.Module;
+                method.FunctionType = AddDelegateToDictionary(method.Namespace, method);
```
ddobrev: Why set this?

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -119,61 +213,111 @@ private void AddDelegatesToDictionary(Declaration decl, Module module)
...
 
-            // HACK : We should shift this call to VisitFunctionTypeParameters in VisitParameter. We can't do it now
-            // because we need to use the module and a since a parameter's namespace is null, we can't reach the
-            // translation unit and thus the module
-            VisitFunctionTypeParameters(function);
-
+            function.ReturnType = CheckType(function.Namespace, function.ReturnType, function);
+            foreach (var parameter in function.Parameters)
+            {
+                parameter.QualifiedType = CheckType(function.Namespace, parameter.QualifiedType, parameter);
```
ddobrev: You set this once and then set it another way. Use the same method if possible. If not, use an else.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -119,61 +213,111 @@ private void AddDelegatesToDictionary(Declaration decl, Module module)
...
+                    Declaration = @delegate
+                };
+                
+                if(!Context.Delegates.ContainsKey(decl))
+                    Context.Delegates.Add(decl, existingDelegate);
+                return new QualifiedType(typeDefType1);
             }
 
-            existingDelegate = new DelegateDefinition(module.OutputNamespace, delegateString);
+            existingDelegate = new DelegateDefinition(module.OutputNamespace, delegateString, access, callingConvention);
```
ddobrev: Shorten the line.

tritao:
Closing in favor of https://github.com/mono/CppSharp/pull/913

PR #916, opened 2017-08-20
Improved getter/setter property to recognize setters that return bools.

tritao:


tests/Common/Common.Tests.cs
```diff
@@ -473,6 +473,8 @@ public void TestProperties()
 
         prop.GetterAndSetterWithTheSameName = 25;
         Assert.That(prop.GetterAndSetterWithTheSameName, Is.EqualTo(25));
+
+        prop.SetterReturnsBoolean(10);
```
ddobrev: What is this? Shouldn't it have been turned into a property?

src/Generator/Passes/GetterSetterToPropertyPass.cs
```diff
@@ -247,7 +249,9 @@ private static string GetPropertyNameFromSetter(Method setter)
 
             private void DistributeMethod(Method method)
             {
-                if (method.OriginalReturnType.Type.IsPrimitiveType(PrimitiveType.Void))
+                if (method.OriginalReturnType.Type.IsPrimitiveType(PrimitiveType.Void) ||
```
ddobrev: Extract method.OriginalReturnType.Type to a local and desugar it.

src/Generator/Passes/GetterSetterToPropertyPass.cs
```diff
@@ -177,6 +177,8 @@ private static void GenerateProperty(string name, DeclarationContext context, Me
                 }
                 property.GetMethod = getter;
                 property.SetMethod = setter;
+                if (setter != null && !setter.ReturnType.Type.IsPrimitiveType(PrimitiveType.Void))
+                    setter.ReturnType = new QualifiedType(new BuiltinType(PrimitiveType.Void));
```
ddobrev: Wouldn't this negatively affect the internal return type and therefore the P/Invocation? Does it work with an indirect return param?

src/Generator/Passes/GetterSetterToPropertyPass.cs
```diff
@@ -247,7 +249,9 @@ private static string GetPropertyNameFromSetter(Method setter)
 
             private void DistributeMethod(Method method)
             {
-                if (method.OriginalReturnType.Type.IsPrimitiveType(PrimitiveType.Void))
+                if (method.OriginalReturnType.Type.IsPrimitiveType(PrimitiveType.Void) ||
+                    (method.OriginalReturnType.Type.IsPrimitiveType(PrimitiveType.Bool) &&
+                     method.Parameters.Count > 1))
```
ddobrev: More than 1 parameter? How is this supposed to work with setters only having one?

tritao:
Closing, fixed by https://github.com/mono/CppSharp/pull/917.

PR #921, opened 2017-08-23
Merged Delegate Pass and pass for Anonymous Delegates

ddobrev:
1. Got rid of Anonymous Names.
2. Merged the common code for Delegate Generation.
3. The delegate pass also works for C++/CLI now.
4. Fixed the calling conventions of delegates for both, C++/CLI and C#.
5. Ensures that if a delegate is used for a virtual as well as something else, it finally ends up as public.
6. Fixed the code generation when the return type of a method is a function pointer that has been used somewhere else as well.
7. Added Access and Calling convention to the delegate definition.

The only thing left is to get rid of the hack used, i.e move the code in VisitFunctionDecl to VisitParametersDecl. Somehow, it's not working right now.

Signed-off-by: Dimitar Dobrev <dpldobrev@protonmail.com>


src/Generator/Passes/DelegatesPass.cs
```diff
@@ -37,177 +20,201 @@ public DelegatesPass()
...
+        {
+            if (Options.IsCLIGenerator)
+                return @namespace is Function ? @namespace.Namespace : @namespace;
+
+            var module = @namespace.TranslationUnit.Module;
+            if (namespacesDelegates.ContainsKey(module))
+                return namespacesDelegates[module];
+
+            Namespace parent = null;
+            if (string.IsNullOrEmpty(module.OutputNamespace))
```
tritao: Add some spaces between the statements

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -37,177 +20,201 @@ public DelegatesPass()
...
+            {
+                var groups = module.Units.SelectMany(u => u.Declarations).OfType<Namespace>(
+                    ).GroupBy(d => d.Name).Where(g => g.Any(d => d.HasDeclarations)).ToList();
+                if (groups.Count == 1)
+                    parent = groups.Last().Last();
+            }
+            if (parent == null)
```
tritao: Add some spaces between the statements

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -37,177 +20,202 @@ public DelegatesPass()
...
-                module, delegateString);
+            if (!base.VisitMethodDecl(method) || !method.IsVirtual || method.Ignore)
+                return false;
 
-            if (existingDelegate != null)
+            if (Options.IsCSharpGenerator)
             {
-                Context.Delegates.Add(decl, existingDelegate);
-                return;
+                var module = method.TranslationUnit.Module;
```
tritao: Delete this variable

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -37,177 +20,199 @@ public DelegatesPass()
...
-            if (new[] { module }.Union(module.Dependencies).Any(
-                m => libsDelegates.ContainsKey(m) && libsDelegates[m].TryGetValue(
-                    delegateString, out @delegate)))
+            var functionType = pointee.Desugar() as FunctionType;
+            if (functionType == null)
+                return type;
+
+            FunctionType newFunctionType = GetNewFunctionType(decl, type);
+
+            var delegateName = GetDelegateName(newFunctionType);
```
tritao: Let's move this checking to `GetDelegate`

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -37,177 +20,199 @@ public DelegatesPass()
...
+
+        private TypedefDecl GetDelegate(Declaration decl, FunctionType newFunctionType, string delegateName)
+        {
+            DeclarationContext namespaceDelegates = GetDeclContextForDelegates(decl.Namespace);
+            var delegateType = new QualifiedType(new PointerType(new QualifiedType(newFunctionType)));
+            var access = decl is Method ? AccessSpecifier.Private : AccessSpecifier.Public;
+            var existingDelegate = allTypedefs.SingleOrDefault(t => t.Name == delegateName);
+            if (existingDelegate != null)
+            {
+                // Code to ensure that if a delegate is used for a virtual as well as something else, it finally ends up as public
```
tritao: Remove the ` Code to ensure that` prefix on the comment, does not say anything relevant and takes up whitespace

Or just wrap the comment into more lines

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -37,177 +20,199 @@ public DelegatesPass()
...
+                if (existingDelegate.Access == AccessSpecifier.Private && access == AccessSpecifier.Public)
+                    existingDelegate.Access = access;
+                // Code to check if there is an existing delegate with a different calling convention
```
tritao: Same here

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -37,177 +20,199 @@ public DelegatesPass()
...
+                // Add the new delegate with calling convention appended to it's name
```
tritao: Typo: it's, should be its

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -37,177 +20,199 @@ public DelegatesPass()
...
+                if (((FunctionType) existingDelegate.Type.GetPointee()).CallingConvention ==
```
tritao: Also add some space between statements

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -37,177 +20,199 @@ public DelegatesPass()
...
+                    Access = access,
+                    Name = delegateName + newFunctionType.CallingConvention,
+                    Namespace = namespaceDelegates,
+                    QualifiedType = delegateType,
+                    IsSynthetized = true
+                };
+                allTypedefs.Add(@delegate);
                 return @delegate;
+            }
+            existingDelegate = new TypedefDecl
```
tritao: This delegate creation is almost the same as the one above, please unify both

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -37,177 +20,199 @@ public DelegatesPass()
...
+            {
+                functionType.Parameters.AddRange(
+                    method.GatherInternalParams(Context.ParserOptions.IsItaniumLikeAbi, true));
+                functionType.CallingConvention = method.CallingConvention;
+                functionType.IsDependent = method.IsDependent;
+                functionType.ReturnType = method.ReturnType;
+            }
+            else
+            {
+                var funcTypeParam = (FunctionType) decl.Type.Desugar().GetFinalPointee().Desugar();
```
tritao: Add a new `FunctionType(FunctionType)` clone constructor like we have for decls and this gets a lot more clean

Issue #923, opened 2017-08-26
error: MS-style inline assembly is not available: Unable to find target for this triple (no targets are registered)


DevSidious:
Fantastic library. However it gets me stuck here:

```
Error parsing 'C:\Folder\header.h'
C:\Program Files (x86)\Windows Kits\10\include\10.0.15063.0\um\winnt.h(949,5): error: MS-style inline assembly is not available: Unable to find target for this triple (no targets are registered)
```

And the culprit:

```
__inline ULONGLONG
NTAPI
Int64ShllMod32 (
    _In_ ULONGLONG Value,
    _In_ DWORD ShiftCount
    )
{
    __asm    {                                         <- line 949
        mov     ecx, ShiftCount
        mov     eax, dword ptr [Value]
        mov     edx, dword ptr [Value+4]
        shld    edx, eax, cl
        shl     eax, cl
    }
}
```
Any ideas on how one could move past this nuisance ?

tritao:
How are you setting up the parser?

It looks like the Clang is not being setup in MS mode.

It should be setup automatically by the driver, but you can try doing it manually by calling `options.SetupMSVC()`.

See https://github.com/mono/CppSharp/blob/27efa2ec871eb79a02c8d3c6ae09287317e8d99a/src/Parser/ParserOptions.cs#L49.

I just noticed we're not actually setting up the Clang target triple in this call, maybe we should, like we do in the CLI driver:

https://github.com/mono/CppSharp/blob/2b5c256514ff8f50b0d90dd8ad1cb1cfb0b32b82/src/CLI/Generator.cs#L42

Try this:

`parserOptions.TargetTriple = "i686-pc-win32-msvc";`

tritao:
Failing that, it might be because of a missing LLVM targets initialization.

@ddobrev I think we might be missing this in our LLVM/Clang initialization code:

```
  // Initialize targets first, so that --version shows registered targets.
  llvm::InitializeAllTargets();
  llvm::InitializeAllTargetMCs();
  llvm::InitializeAllAsmPrinters();
  llvm::InitializeAllAsmParsers();
```

Takem from https://github.com/llvm-mirror/clang/blob/master/tools/driver/cc1_main.cpp#L181

DevSidious:
I'm getting the same result with `parserOptions.TargetTriple = "i686-pc-win32-msvc";` .

[...]

PR #926, opened 2017-08-28
[parser] Added LLVM target initialization and supporting libraries for MS-style inline assembly parsing

DevSidious:
CppSharp fails parsing MS-style inline assembly.

This pull request adds the necessary target initializations and supporting libraries to mitigate this issue.


build/LLVM.lua
```diff
@@ -139,24 +139,31 @@ function SetupLLVMLibs()
       "LLVMVectorize",
       "LLVMBitWriter",
       "LLVMIRReader",
-      "LLVMAsmParser",
       "LLVMOption",
       "LLVMInstrumentation",
       "LLVMProfileData",
       "LLVMX86AsmParser",
+	  "LLVMAsmParser",
```
tritao: Indentation is off here, looks like it's using tabs instead of spaces.

src/CppParser/Parser.cpp
```diff
@@ -49,6 +49,7 @@
 #include <CodeGen/CGCall.h>
 #include <CodeGen/CGCXXABI.h>
 #include <Driver/ToolChains/Linux.h>
+#include <llvm/Support/TargetSelect.h>
```
tritao: Move this include to the top, near the others ` #include <llvm/Support/...>`

src/CppParser/Parser.cpp
```diff
@@ -63,12 +64,23 @@
 using namespace CppSharp::CppParser;
 
 // We use this as a placeholder for pointer values that should be ignored.
-void* IgnorePtr = (void*) 0x1;
+void* IgnorePtr = reinterpret_cast<void*>(0x1);
+
+static bool initialized_llvm = false;
```
tritao: Please rename this to follow our convention, no underscores to separate words in identifiers.

src/CppParser/Parser.cpp
```diff
@@ -63,12 +64,23 @@
...
 
 //-----------------------------------//
 
 Parser::Parser(CppParserOptions* Opts) : lib(Opts->ASTContext), opts(Opts), index(0)
 {
+	if (!initialized_llvm)
```
tritao: Please check that the indentation is correct here.

tritao:
Weirdly this seems to be breaking the CI on Linux, but CppParser seems to be compiled successfully...

ddobrev:
@DevSidious please add a test as well. I tested on my own last night and it's really easy, just add a new function, as simple as possible (free, void, no params) with a single assembler block inside, one instruction is enough. Place it in CLI.h because it's an MSVC-only test project anyway which would simplify your #ifdef guard. You could use the simpler line syntax:

__asm whatever

Just please use an #ifdef guard to only include the function on 32-bit Windows as the __asm keyword is not supported on Windows 64.

ddobrev:
@DevSidious thank you but please also fix the C++ indentation as @tritao 's requested. In addition, when fixing remarks, please do not send new commits but rather amend you existing one. So I have to ask you to merge your commits into one and then push force.
Also, please don't forget the test.

DevSidious:
Alright, added requested changes and merged commits.

src/CppParser/Parser.cpp
```diff
@@ -63,12 +64,23 @@
...
 Parser::Parser(CppParserOptions* Opts) : lib(Opts->ASTContext), opts(Opts), index(0)
 {
+	if (!initializedLLVM)
+	{
+		llvm::InitializeAllTargets();
+		llvm::InitializeAllTargetMCs();
+		llvm::InitializeAllAsmPrinters();
+		llvm::InitializeAllAsmParsers();
+
+		initialized_llvm = true;
```
ddobrev: Please build locally before updating your pull requests. You renamed the field (and manually at that rather than with the refactoring tool of the IDE) but you forgot this one usage. The result is https://ci.appveyor.com/project/tritao/cppsharp/build/1.0.2631#L102 .

src/CppParser/Parser.cpp
```diff
@@ -63,12 +64,23 @@
...
-void* IgnorePtr = (void*) 0x1;
+void* IgnorePtr = reinterpret_cast<void*>(0x1);
+
+static bool initializedLLVM = false;
 
 //-----------------------------------//
 
 Parser::Parser(CppParserOptions* Opts) : lib(Opts->ASTContext), opts(Opts), index(0)
 {
+	if (!initializedLLVM)
```
ddobrev: @tritao asked you to change all indentation, not just in Lua. You left all of your C++ code with tabs rather than 4 spaces. Please fix it.

tests/CLI/CLI.h
```diff
@@ -55,4 +55,13 @@ struct CompleteIncompleteStruct;
 typedef struct IncompleteStruct IncompleteStruct;
 
 DLL_API IncompleteStruct* createIncompleteStruct();
-DLL_API void useIncompleteStruct(IncompleteStruct* a);
\ No newline at end of file
+DLL_API void useIncompleteStruct(IncompleteStruct* a);
+
+#ifdef _WIN32
+void testMSStyleAssembly() {
```
ddobrev: Move { to the next line as per our convention.

tests/CLI/CLI.h
```diff
@@ -55,4 +55,13 @@ struct CompleteIncompleteStruct;
...
+	__asm {
```
ddobrev: We try keeping our tests as simple as possible. Please reduce the test to a single instruction and then use the line syntax (remove {} from the assembly block).

ddobrev:
@DevSidious in addition, please remove "and afferent test" from your commit message, tests are assumed when sending code.

tests/CLI/CLI.h
DevSidious: Do you guys have a document describing your conventions ? It would certainly help.

ddobrev: We don't, ashamed as I am to admit it. However, in 95 % of all cases it's enough to look at already existing code.

ddobrev:
When you've shortened the commit message, please rename your pull after it. The title you've used is useful but its place is in the description (one empty line after the title) of the commit message. Also, please use just "Fixes #923" because I am not sure if the system automatically catches your wording in order to automatically close the issue.

ddobrev:
@DevSidious thank you very much for your work as well as extreme patience. We try keeping our code as clean as possible in order to, so to speak, fix future problems before they actually happen.
I'd like to apologise for failing to explain myself properly about the title of the pull request. The pattern we use is actually like this:

1. Commit message:
1.1. Title (in your case "[parser] Added LLVM target initialization and supporting libraries for MS-style inline assembly parsing.");
<this is an empty line>
1.2. Description (in your case " Fixes #923 ");

2. Title of the pull request - named after the title of the commit message at 1.1 (in your case " Fixes #923").

Please do me a final favour and change the commit message and pull title as instructed above. After this we'll take this over.

ddobrev:
@DevSidious thank you very much for all of your effort. We'll handle the Linux problem now.

DevSidious:
No problem. 
You're fast, changed the title before I had the chance. :)

ddobrev:
@DevSidious yes, I've just discovered I can do it instead and decided not to bother you because I had made a mistake in the description of the pattern.

DevSidious:
It's fine, we are all spread all over the place working on multiple things simultaneously. It's easy for meaning to get lost in all the noise.

Due to the nature of my usage of CppSharp I have the sneaky suspicion this won't be the last time I'll have to submit code. A short explanation on what's expected from contributors on the main project page would **really** streamline things.

tritao:
Merged in https://github.com/mono/CppSharp/pull/927.

Issue #774, opened 2017-02-24
Improve indexer get/set detection


u255436:
##### Brief Description
Currently, when converting a C++ subscript operator to a C++/CLI indexer, the generator incorrectly assumes that only address types (i.e. pointers, references) need a set operator.  Also, it doesn't handle a C++ subscript operator that returns void (though that is a very rare case).  The example code demonstrates:
~~~
struct StructA {};
struct StructB {};
struct StructC {};

struct MyStruct
{
   StructA operator[](StructA a) {}
   const StructB operator[](StructB b) const {}
   void operator[](StructC c) {}
  
   void TestIt()
   {
       MyStruct s;
       auto a = s[StructA()]; // allowed to get
       s[StructA()] = StructA(); // allowed to set (generator doesn't generate set)
     
       auto b = s[StructB()]; // allowed to get
       //s[StructB()] = StructB(); // not allowed to set (the generator shouldn't accidentally allow this)
       
       //auto c = s[StructC()]; // not allowed to get
       //s[StructC()] = StructC(); // not allowed to set
   }
};
~~~

OS: Windows on at least commit 3bfb24228b0eb568165f7bd1f09d5db67a88f154.

##### Used headers
see example above.

##### Used settings
don't think it matters.

##### Stack trace or incompatible generated code
see example above.

tritao:
Hey @u255436, are you still interested in sending a PR fixing this issue?

u255436:
@triato. Yes. I am having some trouble debugging a unit test failure. Is there a guide for how to build and run unit tests in debug mode? (I can only get release mode running)

tritao:
You should only need Debug configuration if you are trying to debug the native Clang parser code.

Are you using Visual Studio for debugging and running tests? If so you should be able to put some breakpoints under a test and debug from there in Release configuration. You only need the VS NUnit test adapter.

tritao:
If you still have the code to fix, we'd appreciate it, even if some test fails it might be useful to get this fixed.

tritao:
Hey @u255436, any news on this?

u255436:
@tritao, I created a new pull request for the changes. (https://github.com/mono/CppSharp/pull/933)

tritao:
Thanks

PR #933, opened 2017-08-30
Improved indexers

Enumer8:
This enables the ability to use the subscript operator from C# in a (contrived) case like:

Original C++ code:
```
int GlobalVariable1 = 10;
int GlobalVariable2 = 20;

struct MyIndexerStruct
{
   MyIndexerStruct(bool b) : _Variable(b ? GlobalVariable1 : GlobalVariable2) {}

   MyIndexerStruct& operator=(const MyIndexerStruct& other)
   {
      _Variable = other._Variable;
      return *this;
   }

   int& _Variable;
};

struct MyStruct
{
   MyIndexerStruct operator[](int) { return MyIndexerStruct(true); }
};

int main()
{
   MyStruct s;
   // Valid C++. It changes GlobalVariable1 to 20.
   s[0] = MyIndexerStruct(false);
   MyIndexerStruct i = s[0];
}
```

User C# code:
```
class Program
{
   static void Main()
   {
      MyStruct s = new MyStruct();
      s[0] = new MyIndexerStruct(false); // shouldn't be an error, but it currently is (set is disabled)
      MyIndexerStruct i = s[0];
   }
}
```

### DIFF
--- a/src/Generator/Passes/CheckOperatorsOverloads.cs
+++ b/src/Generator/Passes/CheckOperatorsOverloads.cs
@@ -110,6 +110,12 @@ namespace CppSharp.Passes
                 if (!qualifiedPointee.Qualifiers.IsConst)
                     property.SetMethod = @operator;
             }
+            // Non-primitive indexers can set if they are not const
+            else if (!returnType.IsPrimitiveType())
+            {
+                if (!@operator.ReturnType.Qualifiers.IsConst)
+                    property.SetMethod = @operator;
+            }
             
             // If we've a setter use the pointee as the type of the property.
             var pointerType = property.Type as PointerType;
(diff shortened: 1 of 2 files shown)


u255436:
@ddobrev, this pull request fixes this issue (https://github.com/mono/CppSharp/issues/774). I tried fixing the unit tests for this about 6 months ago, but I was not able to get the toolchain set up enough to debug the issues. @tritao asked me to still make the pull request, so I'm hoping he can take over.

ddobrev:
@u255436 I didn't know you had an agreement with @tritao. It's fine in such a case.

u255436:
I messed up the forking workflow :) because I forgot this pull request was still open. Oops.

tritao:
Hey @u255436, checking this PR again.

I can't see the point of this change, just based on your example. The return type of `operator[]` is passing things by value, so any change is never reflected on the object.

See https://ideone.com/GPtdy0.

Can you clarify if there was some error on your example or you really want this?

u255436:
@tritao, I guess the main point is that the code generated classes are preventing valid functions from being called. It's possible that I'm the only one that would want to use edge cases like this, so I am okay staying on my fork.

tritao:
@maeteh Feel free to open a new issue with your questions, we can give you some pointers.

maeteh:
Thank you very much

PR #966, opened 2017-09-27
Fixed cause of AccessViolationException

zillemarco:


src/CppParser/Parser.cpp
```diff
@@ -1227,15 +1227,18 @@ Parser::WalkTemplateArgumentList(const clang::TemplateArgumentList* TAL,
 {
     using namespace clang;
 
+	auto LocValid = TSTL && !TSTL->isNull();
+	auto LocTypePtrValid = LocValid && TSTL->getTypePtr() != nullptr;
+
     auto params = std::vector<CppSharp::CppParser::TemplateArgument>();
-    auto typeLocNumArgs = TSTL && TSTL->getTypePtr() ? TSTL->getNumArgs() : 0;
```
ddobrev: The entire fix can be much simpler. Just add && !TSTL->isNull() at this line.

src/CppParser/Parser.cpp
```diff
@@ -1227,15 +1227,18 @@ Parser::WalkTemplateArgumentList(const clang::TemplateArgumentList* TAL,
...
+    auto typeLocNumArgs = LocTypePtrValid ? TSTL->getNumArgs() : 0;
 
     for (size_t i = 0, e = TAL->size(); i < e; i++)
     {
         auto TA = TAL->get(i);
         TemplateArgumentLoc TArgLoc;
         TemplateArgumentLoc *ArgLoc = 0;
-        if (TSTL && TSTL->getTypePtr() && i < typeLocNumArgs && e == typeLocNumArgs)
+        if (LocTypePtrValid && i < typeLocNumArgs && e == typeLocNumArgs)
```
ddobrev: Simply remove the first condition.

src/CppParser/Parser.cpp
zillemarco: Yes I know but I wanted to keep the same syntax used on another point of the code where it did a similar thing. If you prefer I can change like you said.

ddobrev: Fair enough. Just make sure you use 4 spaces rather than tabs.

zillemarco:
@ddobrev not it should be ok

src/CppParser/Parser.cpp
```diff
@@ -1227,15 +1227,18 @@ Parser::WalkTemplateArgumentList(const clang::TemplateArgumentList* TAL,
 {
     using namespace clang;
 
+    auto LocValid = TSTL && !TSTL->isNull();
+    auto LocTypePtrValid = LocValid && TSTL->getTypePtr() != nullptr;
```
ddobrev: Please remove != nullptr for simplicity. This way the line will become shorter which would enable you to merge it with the previous one and therefore use just one variable.

tritao:
I'm more worried about the lack of tests here.

ddobrev:
@zillemarco @tritao 's right, what is the problem that this fixes and can it be tested?

zillemarco:
@tritao @ddobrev I'm working on the tests at the moment. As soon as I have some results I'll tell you but until I'm done I'd leave this PR aside. Strange things happen on my side....

zillemarco:
@ddobrev I removed the != nullptr. Now it should be fine

src/CppParser/Parser.cpp
```diff
@@ -1227,15 +1227,18 @@ Parser::WalkTemplateArgumentList(const clang::TemplateArgumentList* TAL,
 {
     using namespace clang;
 
+    auto LocValid = TSTL && !TSTL->isNull();
```
ddobrev: Merge into a single line/variable.

ddobrev:
@zillemarco the code is great. Thank you for putting up with me.
Please do me one last favour: rebase your code instead of merging master, and combine all of your work into one commit. You can use rebase and amend, and then push force.

zillemarco:
@ddobrev I just fixed it, I was fighting with GitHub Desktop and I'm not used to use git from command line :)

ddobrev:
@zillemarco beautiful. Thank you very much.

Issue #989, opened 2017-10-20
Generating minimal unsafe bindings for pure C lib


raizam:
I have a pure C library with structs and function.

I just want to generate minimal bindings without any encapsulation and handle the higher level myself.
I'm looking for this kind of layout:

public static class ModuleName
{
public static unsafe extern CFunction(ref byte* ptr);

}

Having read the availlable docs, it's not clear to me how to achieve this

realvictorprm:
That's not possible so far I know.

Am 20.10.2017 02:58 schrieb "raizam" <notifications@github.com>:

ddobrev:
@raizam I am not sure what the problem here is. You get more types in your bindings than you need?

raizam:
The problem is that everything gets encapsulated in class Internal. This is a low level C library and I don't want to get high level C# implementation.
Also, cppsharp is not aware about memory ownship of the library and generates a lot of undesired behavior, like Dictionnary mappings between managed/native instances.

Anyway...don't bother with this, I'm writing my own parser dedicated for this library

tritao:
You can easily extend CppSharp to get this functionality. We've had some requests over the years for this and would gladly add it, but so far no one has contributed an actual PR for it.

PR #991, opened 2017-10-23
Implementation of GenerateRawCBinding option

raizam:
This implements GenerateRawCBinding option, which is designed for C libraries, as described here in https://github.com/mono/CppSharp/issues/989.

This option remove the high level classes, and generates only the "internal" classes as structs, with raw pointers.

if enabled with C++ libraries, the behavior undefined and will probably not work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ src/generator/generator
 *DesignTime*
 *TemporaryGeneratedFile*
 *FileListAbsolute*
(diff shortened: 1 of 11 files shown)


raizam:
@ddobrev here is a gist of the generated code for the Nuklear library: https://gist.github.com/raizam/edba07fb4c019b86ca235690a917ce76

PR #1005, opened 2017-11-19
Fixed boolean arrays

### LINKED ISSUE #1004: Cannot implicitly convert type `byte' to `bool' when binding dear imgui
##### Brief Description

When I try to bind [dear imgui](https://github.com/ocornut/imgui/blob/master/imgui.h), the generated C# code is not compilable. Problem is caused by https://github.com/ocornut/imgui/blob/28a31997b76c59dba9d6d2ba03c5c1cf7b3a6ce6/imgui.h#L855

OS: Windows 7

##### Used headers

https://github.com/ocornut/imgui/blob/28a31997b76c59dba9d6d2ba03c5c1cf7b3a6ce6/imgui.h

##### Used settings

Target: MSVC

Other settings 

```cs
        public bool[] MouseDown
        {
            get
            {
                bool[] __value = null;
                if (((global::Engine.ImGuiIO.__Internal*) __Instance)->MouseDown != null)
                {
                    __value = new bool[5];
                    for (int i = 0; i < 5; i++)
                        __value[i] = ((global::Engine.ImGuiIO.__Internal*) __Instance)->MouseDown[i];
                }
                return __value;
            }

            set
            {
                if (value != null)
                {
                    for (int i = 0; i < 5; i++)
                        ((global::Engine.ImGuiIO.__Internal*) __Instance)->MouseDown[i] = value[i];
                }
            }
        }
```

```error CS0029: Cannot implicitly convert type `byte' to `bool'```

nem0:
fixes #1004 

I'm on Windows and there was docs only for testing on Linux and I have no experience with C# tests, so I was not able to actually run the test.

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpMarshal.cs
+++ b/src/Generator/Generators/CSharp/CSharpMarshal.cs
@@ -115,7 +115,10 @@ namespace CppSharp.Generators.CSharp
                         }
                         else
                         {
-                            if (arrayType.IsPrimitiveType(PrimitiveType.Char) &&
+                            if (arrayType.IsPrimitiveType(PrimitiveType.Bool))
+                                supportBefore.WriteLineIndent($@"{value}[i] = {
+                                    Context.ReturnVarName}[i] != 0;");
+                            else if (arrayType.IsPrimitiveType(PrimitiveType.Char) &&
                                 Context.Context.Options.MarshalCharAsManagedChar)
                                 supportBefore.WriteLineIndent($@"{value}[i] = global::System.Convert.ToChar({
                                     Context.ReturnVarName}[i]);");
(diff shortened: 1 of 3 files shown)


ddobrev:
@nem0 to run the tests on Windows and from within VS, you can use https://marketplace.visualstudio.com/items?itemName=NUnitDevelopers.NUnit3TestAdapter . Whichever way you choose, please check your code because our Windows CI fails.
Please note that you should amend your existing commit and push force rather than add a new commit on top. In addition, now that you need to apply changes anyway, please improved your commit message as well to something like "Fixed the generated C# for fixed arrays of Boolean-s.". We support two target languages in C++# with the intention to add even more in the future so our commit messages have to reflect which generator, if any, we've fixed.

nem0:
I did not know about the other languages, the project name is a bit misleading. Sorry about the second commit, I made a typo in command line ```-amend``` instead of ```--amend```. You can always squash it.

ddobrev:
Squashing just removes the merge commit, I'm afraid. What Git client do you use? If it's the excellent TortoiseGit, you can show the log, then select your two commits and right-click to get to combinding them into one.

nem0:
https://github.com/blog/2141-squash-your-commits it squashes everything into one commit. I use git-bash, git's command line on windows. Unfortunately I can not fix it in my fork, since PR is already created.

ddobrev:
@nem0 you were right, I thought squashing only removed the merge commit. Thank you for the tip and your work.

nem0:
Thank you for prompt responses and merge. :+1: 

Note: when you want to get rid just of the merge commit, that's what rebase do.

PR #1011, opened 2017-11-30
Force-set platform type to managed targets.

rokups:
In some environments mono defaults building managed targets in architecture different from the host (like AnyCPU or x86 on x64 host). For example on x64 host (linux) build system would produce i686 architecture managed executables while native libraries are built as amd64. In such case managed executables are unable to load native libraries because they target different native architectures. This change ensures architecture consistency between all built binaries.

### DIFF
--- a/build/Helpers.lua
+++ b/build/Helpers.lua
@@ -100,6 +100,7 @@ end
 function SetupManagedProject()
   language "C#"
   location ("%{wks.location}/projects")
+  buildoptions {"/platform:".._OPTIONS["arch"]}
 
   dotnetframework "4.6"
 


ddobrev:
@rokups thank you for this code. I would just like to wait until our macOS build has passed too since this is a change to the build itself and I'll merge.

rokups:
It fails on Mac for some reason. It is not clear to me why. Maybe i should just have this option active on linux/windows and call it a day?

ddobrev:
@rokups how can you tell it does? Our CI build hasn't finished yet (macOS builds on Travis can sometimes take very long to start).

tritao:
I restarted the build, it seemed a spurious failure to me, all tests were passing.

PR #1013, opened 2017-11-30
Cloning llvm from git replaced with downloading archives 

rokups:
Exact requires commits are downloaded. Entire process is now limited by your internet connection speed + archive unpacking.

One thing to watch out - at least on linux `os.rmdir` fails to remove some symlinks. Must be premake limitation. I have to manually delete that directory.

Edit:
`package_llvm` command creates archive with wrong commit hash. Do not merge until fixed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,7 +35,7 @@ src/generator/generator
 *DesignTime*
 *TemporaryGeneratedFile*
 *FileListAbsolute*
-/build/scripts/llvm*
+/build/scripts/*.tar.gz
 /build/scripts/.vagrant
 /build/vs2012
 /build/vs2013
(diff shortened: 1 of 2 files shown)


tritao:
This seems to break the `package_llvm` step due to this call: https://github.com/mono/CppSharp/blob/master/build/scripts/LLVM.lua#L310

```
➜  scripts git:(master) ✗ git -C "/Users/joao/Dev/CppSharp/deps/llvm" rev-parse HEAD

1d34ccd9ea0b4cc095b85db90b6d066ddfd388fc
```

It now returns the Git commit hash of CppSharp instead of LLVM since there's no `.git` now with archives.

Sent a fix.

rokups:
Yes, i added note in the first post to not merge until i fix this.. Thanks for taking care of it 👍

PR #1015, opened 2017-11-30
Preliminary script for building 32-bit Nuget package

aybe:
This is an attempt for a script to build nuget package,
 - checks for nuget and downloads it if not found
   - the dev **must** install wget first, this is also checked and batch will show download link if not found
     - I didn't include it because it's not a single executable, url points to an installer
  - creates the scratch folder that contains nupkg content, builds the package
  - cleans up

Problems:
  - added dependency to Microsoft.VisualStudio.Setup.Configuration.Interop but it never copies itself to output even on a brand new project
    - (include it manually in cppsharp package ?)

That said, when you copy above DLL by hand on a test project, the package works as intended.

The next step would be to have the same for 64-bit, then packing C++# would be a matter of running a batch file. By the way, I didn't even bother looking as there's no x64 configuration to switch to in the solution.

Review it and let me know what you think, cheers !


tritao:
Hey @aybe, this looks mostly OK to me, but is limited to Windows.

Would you be interested in helping out with integrating Cake with a task to do this? https://cakebuild.net/ That would work across all platforms.

aybe:
Hi @tritao, this 'cake' thing looked appealing at first, but it's an absolute piece of crap to work with. From the poor (generated) docs and not-without-issues setup on VSCode, this was not a funny exercise ! I thought the worst docs in the world were Unity3D's but I changed my mind :)

Anyway, it's here now, it works but should be tested and extended. And there's a strong possibility I did something wrong since it was my day 1 with cake, so check all that.

By the way, I've tried to look at 64-bit, but from what I've understood there's no 64-bit llvm 7z to fetch from the repo; and premake should be able to deploy in something like 'vs2017-platformname' so one can have both solutions at the same time; I will let you handle that part.

Cheers !

tritao:
@aybe Thanks for looking into it. Haven't had time yet to review this one, will try do so soon.

aybe:
ok

tritao:
Sorry for delay, thanks for contribution.

aybe:
You welcome !

PR #1018, opened 2017-12-01
Use correct LLVM build dir for includes from unpackaged LLVM builds

rokups:
`get_llvm_build_dir()` returns directory `<llvm_dir>/build`. It is incorrect when using non-packaged LLVM build, because LLVM is built into `'deps/llvm/'..get_llvm_package_name()`. This change fixes issue where using unpackaged LLVM build resulted in `error: ‘StrictFP’ is not a member of ‘llvm::Attribute’`.

### DIFF
--- a/build/LLVM.lua
+++ b/build/LLVM.lua
@@ -25,7 +25,13 @@ function SearchLLVM()
 end
 
 function get_llvm_build_dir()
-  return path.join(LLVMRootDir, get_llvm_package_name())
+  local packageDir = path.join(LLVMRootDir, get_llvm_package_name())
+  local buildDir = path.join(LLVMRootDir, "build")
+  if os.isdir(buildDir) then
+    return buildDir
+  else
+    return packageDir
+  end
 end
 
 function SetupLLVMIncludes()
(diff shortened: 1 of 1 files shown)


tritao:
Why not change `get_llvm_build_dir` itself to point to the correct dir?

rokups:
Indeed that would be much more cleaner solution. Will do

rokups:
I slipped one more commit that fixes lib linking on linux. Hope it is alright. I am PR-spamming too much already :)

To summarize it links `z` and `pthreads` on linux as well as on mac. In addition it links `tinfo` if this library exists. Explicitly disabling these libs in LLVM config did not work even though `config.h` has right preprocessor defines to not have symbols pulled in from these libs.

Aaaand i broke it. Will fix commit in a sec.

tritao:
I think this might break the build for pre-package LLVM builds, which have a root `build` folder. We should probably be checking `LLVMRootDir` for a `build` folder. If one exists, we use it, otherwise just use `get_llvm_package_name()` as in this PR.

PR #1022, opened 2017-12-08
Multiple fixes

rokups:
Please see commit messages for details on what was fixed. No point in repeating that here.

In addition to fixes i added `GenerateFormatted` option which forces producing formatted c# bindings code. This was not possible if wrapper was set to be compiled by CppSharp. In case this is unnecessary or unacceptable please let me know and i will nuke the commit.

### DIFF
--- a/src/Generator.Tests/GeneratorTest.cs
+++ b/src/Generator.Tests/GeneratorTest.cs
@@ -28,6 +28,7 @@ namespace CppSharp.Utils
             options.GeneratorKind = kind;
             options.OutputDir = Path.Combine(GetOutputDirectory(), "gen", name);
             options.Quiet = true;
+            options.GenerateDebugOutput = true;
             var testModule = options.AddModule(name);
             testModule.SharedLibraryName = $"{name}.Native";
 
--- a/src/Generator/Generators/CSharp/CSharpSources.cs
+++ b/src/Generator/Generators/CSharp/CSharpSources.cs
@@ -203,6 +203,27 @@ namespace CppSharp.Generators.CSharp
             return true;
         }
 
+        private IEnumerable<Class> EnumerateClasses()
+        {
+            foreach (var tu in TranslationUnits)
+            {
+                foreach (var cls in EnumerateClasses(tu))
+                    yield return cls;
+            }
+        }
+
+        private IEnumerable<Class> EnumerateClasses(DeclarationContext context)
+        {
+            foreach (var cls in context.Classes)
+                yield return cls;
+
+            foreach (var ns in context.Namespaces)
+            {
+                foreach (var cls in EnumerateClasses(ns))
+                    yield return cls;
+            }
+        }
+
         void GenerateNamespaceFunctionsAndVariables(DeclarationContext context)
         {
             var hasGlobalVariables = !(context is Class) && context.Variables.Any(
@@ -213,8 +234,11 @@ namespace CppSharp.Generators.CSharp
 
             PushBlock(BlockKind.Functions);
             var parentName = SafeIdentifier(context.TranslationUnit.FileNameWithoutExtension);
-            var @class = context.Classes.Find(c => c.Name == parentName);
-            var keyword = @class != null && @class.IsValueType ? "struct" : "class";
+
+            var keyword = "class";
+            var classes = EnumerateClasses().ToList();
+            if (classes.FindAll(cls => cls.IsValueType && cls.Name == parentName && context.QualifiedLogicalName == cls.Namespace.QualifiedLogicalName).Any())
+                keyword = "struct";
             WriteLine($"public unsafe partial {keyword} {parentName}");
             WriteStartBraceIndent();
 
--- a/src/Generator/Generators/CSharp/CSharpSources.cs
+++ b/src/Generator/Generators/CSharp/CSharpSources.cs
@@ -213,7 +213,9 @@ namespace CppSharp.Generators.CSharp
 
             PushBlock(BlockKind.Functions);
             var parentName = SafeIdentifier(context.TranslationUnit.FileNameWithoutExtension);
-            WriteLine("public unsafe partial class {0}", parentName);
+            var @class = context.Classes.Find(c => c.Name == parentName);
+            var keyword = @class != null && @class.IsValueType ? "struct" : "class";
+            WriteLine($"public unsafe partial {keyword} {parentName}");
             WriteStartBraceIndent();
 
             PushBlock(BlockKind.InternalsClass);
--- a/src/Generator/Generators/CodeGenerator.cs
+++ b/src/Generator/Generators/CodeGenerator.cs
@@ -61,14 +61,6 @@ namespace CppSharp.Generators
 
         public abstract void Process();
 
-        public override string Generate()
-        {
-            if (Options.IsCSharpGenerator && Options.CompileCode && !Options.GenerateDebugOutput)
-                return base.GenerateUnformatted();
-
-            return base.Generate();
-        }
-
         public virtual void GenerateFilePreamble(CommentKind kind, string generatorName = "CppSharp")
         {
             var lines = new List<string>
--- a/src/Generator/Generators/CodeGenerator.cs
+++ b/src/Generator/Generators/CodeGenerator.cs
@@ -92,7 +92,14 @@ namespace CppSharp.Generators
         public virtual void GenerateDebug(Declaration decl)
         {
             if (Options.GenerateDebugOutput && !string.IsNullOrWhiteSpace(decl.DebugText))
-                WriteLine("// DEBUG: " + decl.DebugText);
+            {
+                char[] newLineChars = {'\r', '\n'};
+                var text = decl.DebugText;
+                var index = text.IndexOfAny(newLineChars);
+                if (index >= 0)
+                    text = text.Substring(0, index);
+                WriteLine("// DEBUG: " + text);
+            }
         }
 
         #endregion
--- a/src/Generator/Generators/CodeGenerator.cs
+++ b/src/Generator/Generators/CodeGenerator.cs
@@ -83,22 +83,18 @@ namespace CppSharp.Generators
         public virtual void GenerateDeclarationCommon(Declaration decl)
         {
             if (decl.Comment != null)
-            {
                 GenerateComment(decl.Comment);
-                GenerateDebug(decl);
-            }
+
+            GenerateDebug(decl);
         }
 
         public virtual void GenerateDebug(Declaration decl)
         {
             if (Options.GenerateDebugOutput && !string.IsNullOrWhiteSpace(decl.DebugText))
             {
-                char[] newLineChars = {'\r', '\n'};
-                var text = decl.DebugText;
-                var index = text.IndexOfAny(newLineChars);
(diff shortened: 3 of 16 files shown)


src/Generator/Driver.cs
```diff
@@ -452,13 +449,16 @@ public static void Run(ILibrary library)
             if (!driver.Options.DryRun)
             {
                 driver.SaveCode(outputs);
-                if (driver.Options.IsCSharpGenerator && driver.Options.CompileCode)
+                if (driver.Options.IsCSharpGenerator && driver.Options.CompileCode &&
+                    !driver.Options.GenerateDebugOutput)
```
ddobrev: What's the reason for this change?

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -213,7 +213,8 @@ void GenerateNamespaceFunctionsAndVariables(DeclarationContext context)
 
             PushBlock(BlockKind.Functions);
             var parentName = SafeIdentifier(context.TranslationUnit.FileNameWithoutExtension);
-            WriteLine("public unsafe partial class {0}", parentName);
+            var @class = context.Classes.Find(c => c.Name == parentName);
+            WriteLine("public unsafe partial {0} {1}", @class != null && @class.IsValueType ? "struct" : "class", parentName);
```
ddobrev: Please use an interpolated string instead.

src/Generator/Options.cs
```diff
@@ -110,6 +110,11 @@ public Module AddModule(string libraryName)
         /// </summary>
         public bool GenerateDebugOutput;
 
+        /// <summary>
+        /// If set to true, the generated code will always be formatted.
+        /// </summary>
+        public bool GenerateFormatted;
```
ddobrev: Please drop this option and always format the code instead. I.e. remove GenerateUnformatted.

src/Generator/Driver.cs
rokups: Code with debug output is not compilable as debug output is not valid c# syntax.

src/Generator/Options.cs
rokups: This is even better, will do

src/Generator/Driver.cs
tritao: That should not be happening, do you have an example of generated debug output that does not compile?

tritao:
Looks good, can you add a test for https://github.com/mono/CppSharp/pull/1022/commits/3f227b1f5d8358e17c98ba0b885eb2576528b0cb?

src/Generator/Driver.cs
rokups: [Urho3D.cs.txt](https://github.com/mono/CppSharp/files/1543388/Urho3D.cs.txt)

ddobrev:
@rokups you seem to have fixed 3 unrelated bugs in this pull, none of them with tests. Would you mind sending separate commits for each?

tritao:
@rokups About the debug text issue, try to change the code here: https://github.com/mono/CppSharp/blob/426572bc8a5aee4ad178dc8c0228733c4222378a/src/Generator/Generators/CodeGenerator.cs#L103

We should only be printing the debug text up to the first new line, or else the generated code gets screwed.

src/Generator/Driver.cs
```diff
@@ -53,9 +53,6 @@ void ValidateOptions()
             {
                 if (string.IsNullOrWhiteSpace(module.LibraryName))
                     throw new InvalidOptionException("One of your modules has no library name.");
-
-                if (module.OutputNamespace == null)
-                    module.OutputNamespace = module.LibraryName;
```
ddobrev: The problem here is that our API is not intuitive enough. If you set the output name-space to the empty string, you won't get a repetitive name-space. In other words, we make a difference between null and empty in this case but we haven't documented it anywhere. And this code is actually important because it automatically provides the user with a decent name-space. So for now please use the empty string instead and report an issue about this. I'll think of a way to improve our API.

rokups:
`Consistently declare classes/structs` commit fixes issue in my case, however test i tried to add still fails. Still not sure why.

No idea what code i am supposed to add to a test to trigger delegate generation.

Also i slipped in some CI changes, added `sudo` requirement for travis so that it can build with my account as well. No idea how your account worked without that directive because `sudo` is required in order to use `apt-get` in the image. That should not have any impact on your end whatsoever.

Also added x64 windows build testing to appveyor. 32bit systems are absolute minority nowdays so there is no point in testing only x86. Hope that is OK.

src/Generator/Generators/CodeGenerator.cs
```diff
@@ -61,14 +61,6 @@ protected CodeGenerator(BindingContext context, IEnumerable<TranslationUnit> uni
 
         public abstract void Process();
 
-        public override string Generate()
-        {
-            if (Options.IsCSharpGenerator && Options.CompileCode && !Options.GenerateDebugOutput)
-                return base.GenerateUnformatted();
```
ddobrev: Please also delete the definition of GenerateUnformatted and any other methods no longer used.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -187,6 +187,15 @@ private DeclarationContext GetDeclContextForDelegates(DeclarationContext @namesp
                     ).GroupBy(d => d.Name).Where(g => g.Any(d => d.HasDeclarations)).ToList();
                 if (groups.Count == 1)
                     parent = groups.Last().Last();
+                else
+                {
+                    foreach (var g in groups)
+                    {
+                        parent = g.ToList().Find(ns => ns.Name == module.LibraryName || ns.Name == module.OutputNamespace);
```
ddobrev: For a test, you could add a method which takes a function pointer, or a virtual method, to a test binding where we do not define an output name-space, just like in your case. The function pointer or a virtual method would force the generation of a delegate so the issue should be reproduced.
In general, I would really like to just merge your code and make it "easy" to both you and me. Unfortunately, our system covers an extremely large area (C++) and the lack of tests is guaranteed to allow regressions in the long term. To this very day creating a test case is by far my most unpleasant part of the work but it is simply necessary. I really hope you understand.

ddobrev:
Also, what happened with the problem of GenerateDebugOutput?

src/Generator/Passes/DelegatesPass.cs
rokups: Completely agreed. Good test suite is essential for project like CppSharp. I would like to actually step it up and suggest adding more CI test builds. There could easily be tests for GCC 4.8 / 5 / 6 / and maybe 7 added, including LLVM builds by CI. To compensate long build times LLVM builds can be cached as well.

rokups:
I removed commit regarding debug output, will add fix as @tritao suggested instead.

src/Generator/Passes/DelegatesPass.cs
rokups: One more thing - i have patch in the works that changes llvm package name, adding extra info for linux builds. Possible names with the patch:

* `llvm-9bde5f-linux-x64-gcc-4.8-Release`
* `llvm-9bde5f-linux-x64-gcc-4.9-Debug`
* `llvm-9bde5f-linux-x64-gcc-5-Debug`
* `llvm-9bde5f-linux-x64-gcc-6-Release-no-cxx11`

It would make it easier to support multiple compiler versions on linux as well as pave the way for testing multiple compiler versions on travis.

ddobrev: The idea is more than brilliant but the problem is pure manpower. We need to update Clang every now and then (we try doing it twice a year) and it's cumbersome enough as it is. 3 packages to build each time, Windows, macOS and Linux, and your suggestion would make them 7. In addition we'd also have to maintain different parser bindings for each. I simply cannot imagine who could take responsibility for all of this grunt work.

rokups: Considering small coverage of linux binary builds you could even retire them altogether. Linux users can build LLVM locally. As for bindings - there is no need to maintain more bindings than supported at the moment. You have linux bindings generated with gcc 4.8 right? I generated them with gcc 7.2. The [difference](https://github.com/rokups/CppSharp/commit/b9a0abd41cf5b4af6d8812a6716fa5d827a01436) is minimal and that code is probably not even used by CppSharp, so you can keep maintaining bindings for the single compiler version that is officially supported. So all extra CI jobs would do is give wider test coverage. All of these compilers are supposed to work now anyway.

ddobrev:
@rokups thank you for all of your work. Please just add 3 tests: for the debug output, for the delegates and for the class/struct fix. When you're done, I'll take the responsibility of merging your code.

rokups:
Yeah i am trying.. Thing is i cant really test changes on my own machine because linux builds are still having issues. So i installed gcc 4.8 in attempt to verify and fix failing tests on supported compiler at least. This prompted llvm package name change i talked about earlier, because testing different compilers needs different llvm builds.. After implementing per-compiler-per-configuration llvm package naming it turned out that non-global compilers store includes in other directories not covered by current code, so i went ahead and fixed that part... Just finished doing that all and i can finally test multiple compilers on my machine. Will get to tests now, hopefully they will be ready some time tomorrow.

ddobrev:
@rokups excellent.

rokups:
Any idea why enabling debug output for tests would produce no debug output? For some reason [decl.Comments](https://github.com/mono/CppSharp/blob/master/src/Generator/Generators/CodeGenerator.cs#L93) is always null in tests.

ddobrev:
@rokups does the declaration in question actually have comments?

ddobrev:
Also, please check your work on Linux headers because it breaks our Linux CI.

rokups:
I tried enabling debug output for CSharp test, it has comments. But debug output is supposed to add fragments from c++ code as well.

ddobrev:
@rokups I see. Well, it's easiest to copy a piece of the header you got the real problem with and check if the test exhibits the same behaviour.

src/Generator.Tests/GeneratorTest.cs
```diff
@@ -28,6 +28,7 @@ public virtual void Setup(Driver driver)
             options.GeneratorKind = kind;
             options.OutputDir = Path.Combine(GetOutputDirectory(), "gen", name);
             options.Quiet = true;
+            options.GenerateDebugOutput = true;
```
ddobrev: Why do we need this for all test projects? Also, where is your actual test?

rokups: If debug output breaks then tests will fail. Consider it a debug output test. struct/class test is [here](https://github.com/mono/CppSharp/pull/1022/commits/4e735b5b61f7191e2140e39012d5ee517f314574) and delegates issue i cant replicate in a test. Already spent few too many hours on this.. So we will have to do without that test. I just cant spend that kind of time on nice-to-haves for free.

ddobrev: @rokups I understand. Consider it done. I'll merge it in the next few days.

tritao:
Thanks @rokups, changes look good, merged.

If you can send those changes you mentioned about supporting multiple toolchains on the LLVM packages that would be awesome, as well as testing more GCC versions on CI.

PR #1028, opened 2017-12-16
GetterSetterToPropertyPass customization

rokups:
Remove internal class from `GetterSetterToPropertyPass` and make some methods `protected virtual` in order to allow customization of this pass.

`GetterSetterToPropertyPass` is trying to be too smart and converts too many things to properties. One such example is `void Run()` method getting converted to property, which makes no sense. I can not tell original intention behind logic in `IsGetter()` so instead of changing it i propose moving everything from private `PropertyGenerator` class to `GetterSetterToPropertyPass` and making some methods virtual so user may inherit this class, override some methods and replace the pass.

My personal opinion is that CppSharp is trying to be too smart here. Converting obvious things to properties (methods stating with "get"/"set" and maybe "is") is fine, however applying heuristics to guess what user might wish is not a good idea. Heuristics get things wrong. I propose completely removing guessing part. I can do this if approved. User should still be able to customize this pass though.

Also logic in `IsGetter()` seems broken:
```cs
            return (firstWord.Length < method.Name.Length &&
                Match(firstWord, new[] { "get", "is", "has" })) ||
                (!Match(firstWord, new[] { "to", "new" }) && !verbs.Contains(firstWord));
```
A method is considered getter if it 1. does not begin with "to" or "new" and 2. is not in verbs.txt file which is full of weird words that are used rarely if ever.

### DIFF
--- a/src/AST/Method.cs
+++ b/src/AST/Method.cs
@@ -105,6 +105,7 @@ namespace CppSharp.AST
             SynthKind = method.SynthKind;
             AdjustedOffset = method.AdjustedOffset;
             OverriddenMethods.AddRange(method.OverriddenMethods);
+            ConvertToProperty = method.ConvertToProperty;
         }
 
         public Method(Function function)
@@ -171,6 +172,8 @@ namespace CppSharp.AST
         public int AdjustedOffset { get; set; }
 
         public List<Method> OverriddenMethods { get; } = new List<Method>();
+
+        public bool ConvertToProperty { get; set; }
 
         public Method GetRootBaseMethod()
         {
--- a/src/Generator/Library.cs
+++ b/src/Generator/Library.cs
@@ -4,6 +4,7 @@ using System.Globalization;
 using System.Linq;
 using System.Text.RegularExpressions;
 using CppSharp.AST;
+using CppSharp.Passes;
 
 namespace CppSharp
 {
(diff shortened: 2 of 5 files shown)


ddobrev:
@rokups we are not removing the heuristics. I think you are misunderstanding the concept here which is normal as you are not that familiar with the code and the approach is not documented which is my mistake. Allow me to explain.
The bug with Run() you have experienced is due to the verb "run" missing from the list for reasons most probably lying in the original. I.e. I took this list years ago from a categorised dictionary as I was sane enough not to write it myself. This is the reason you see it full of strange words. Any non-void parameterless C++ function the name of which does not start with a verb is mapped to a property. This, being heuristic in nature, makes mistakes every now and then but this is the price of automation. For the 5 mistakes it makes it generates 10 000 conveniences (properties). In addition to being few, such mistakes are not fatal because the code can still be called, one just needs a useless local variable in the extremely rare cases of not needing the returned result.
Given the above, I have to ask you to provide more examples of this pass behaving poorly before I consider merging your code. If not, your fix is nothing but adding "run" to verbs.txt.

rokups:
I have not run into any such cases yet, but i still think ability to customize this pass is necessary simply because heuristics do get things wrong or maybe user simply does not wish to turn certain methods to properties.

ddobrev:
You have a point that I haven't provided room for customisation so far. However, it can be done much more simply by marking specific methods as always such, similar to our methods for manually ignoring functions. I would really prefer not to have the entire pass rewritten just to expect users to then rewrite it once more themselves. Save this code of yours just in case but I really think what you want can be achieved with much less - if you really need it at this point, that is.

rokups:
Oh this code isnt really worth saving.. Diff is big but all i did was remove private class and un-indented most of the code..

Ok then.. How about methods like`ctx.IgnorePropertyConversion(pattern)`, `ctx.ForcePropertyConversion(pattern)` and `pass.SetHeuristicsEnabled(bool)`? That should probably be enough.

ddobrev:
I guess the latter is an option not to use heuristics at all, so it's place would be in DriverOptions with the proper name. The rest is fine, I would just rename a little - IgnoreConversionToProperty and ForceConversionToProperty.

rokups:
OK, will do it and push to this branch.

tritao:
We already have a mechanism to disallow passes to run for a specific decl. Let's try to re-use it here if possible: https://github.com/mono/CppSharp/blob/master/src/Generator/Passes/Pass.cs#L47

rokups:
Oh awesome, thank you for hint. I definitely would have done it wrong

ddobrev:
@tritao you're right, I had forgotten about the exclusion.
The issue is not what's attempted to be achieved but how it's achieved. While I stand behind the pass working by default, I agree we lack a point of customisation for its behaviour at present. However, I think overriding virtual methods is not much easier than entirely writing a custom pass, so a simpler solution such as the one I've already suggested will be more useful.

rokups:
You guys kind of forgot this one ;)

tritao:
Yep, we did 👍

PR #1040, opened 2018-01-06
Fix generated symbols checking when `Options.CheckSymbols = true`

rokups:
When `Options.CheckSymbols` was enabled CppSharp discarded generated symbols from `<name>-symbols.cpp` as non-existing, because they were not registered with `Context.Symbols`. This change registers newly generated symbols with `Context.Symbols` and they are no longer discarded.

Also previously incorrect library name for generated symbols was picked. CppSharp defaulted to picking library name of managed wrapper. As a result managed wrapper and symbols library would have name collision on windows where prefixes and suffixes of native libraries are the same as of managed libraries. This change registers generated symbols to `module.SymbolsLibraryName` and now generated symbols are correctly imported from `<name>-symbols.dll` or `lib<name>-symbols.so`.

### DIFF
--- a/src/AST/SymbolContext.cs
+++ b/src/AST/SymbolContext.cs
@@ -80,15 +80,20 @@ namespace CppSharp.AST
             {
                 foreach (var symbol in library.Symbols)
                 {
-                    if (!Symbols.ContainsKey(symbol))
-                        Symbols[symbol] = library;
-                    if (symbol.StartsWith("__", StringComparison.Ordinal))
-                    {
-                        string stripped = symbol.Substring(1);
-                        if (!Symbols.ContainsKey(stripped))
-                            Symbols[stripped] = library;
-                    }
+                    AddSymbol(library, symbol);
                 }
+            }
+        }
+
+        public void AddSymbol(NativeLibrary library, string symbol)
+        {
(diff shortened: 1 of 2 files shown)


rokups:
Tests broke and i am not sure what to do here. Tests apparently assume that actual wrapped cpp code and `*-symbols.cpp` are both built into the same native library. This approach creates chicken and egg problem with `CheckSymbols = true`. So... I think logical solution would be to add one more native project `*-symbols` when setting up tests for building generated symbols. After all tests already appear to be doing exactly this for `Std-symbols.cpp`. How does that sound?

ddobrev:
@rokups this is not how it works. Users should handle SymbolsCodeGenerated ( https://github.com/mono/CppSharp/blob/master/src/Generator/Passes/GenerateSymbolsPass.cs#L71 ), set the event args and then C++# will collect the symbols itself ( https://github.com/mono/CppSharp/blob/master/src/Generator/Passes/GenerateSymbolsPass.cs#L221 ). Please check https://gitlab.com/ddobrev/QtSharp/blob/master/QtSharp/QtSharp.cs#L222 for an example.
I need to document this but it's another matter.

rokups:
Oh! Thank you for clarification 👍

Issue #1043, opened 2018-01-10
Member variable properties create copies instead of returning references


rokups:
CppSharp generates following getter for `HashMap<StringHash, Variant> engineParameters_;`:
```cs
protected global::Urho3D.HashMap<global::Urho3D.StringHash, global::Urho3D.Variant> engineParameters_
{
    get
    {
        return global::Urho3D.HashMap<global::Urho3D.StringHash, global::Urho3D.Variant>.__CreateInstance(((global::Urho3D.Application.__Internal*) __Instance)->engineParameters_);
    }

    set
    {
        ((global::Urho3D.Application.__Internal*) __Instance)->engineParameters_ = ReferenceEquals(value, null) ? new global::Urho3D.HashMap.__Internal() : *(global::Urho3D.HashMap.__Internal*) value.__Instance;
    }
}
```
Problem is that `__CreateInstance` call invokes a constructor which in turn calls `__CopyValue(native)`. Property returns a new copy of hashmap. There is another constructor overload that does not copy object, but it is supposed to be called with IntPtr instance instead. This probably is not intended behaviour right?

Constructors:
```cs
internal static new global::Urho3D.HashMap<T, U> __CreateInstance(global::System.IntPtr native, bool skipVTables = false)
{
    return new global::Urho3D.HashMap<T, U>(native.ToPointer(), skipVTables);
}

internal static global::Urho3D.HashMap<T, U> __CreateInstance(global::Urho3D.HashMap.__Internal native, bool skipVTables = false)
{
    return new global::Urho3D.HashMap<T, U>(native, skipVTables);
}

private static void* __CopyValue(global::Urho3D.HashMap.__Internal native)
{
    var ret = Marshal.AllocHGlobal(sizeof(global::Urho3D.HashMap.__Internal));
    global::Urho3D.HashMap.__Internal.cctorc__N_Urho3D_S_HashMap____N_Urho3D_S_StringHash___N_Urho3D_S_Variant(ret, new global::System.IntPtr(&native));
    return ret.ToPointer();
}

private HashMap(global::Urho3D.HashMap.__Internal native, bool skipVTables = false)
    : this(__CopyValue(native), skipVTables)
{
    __ownsNativeInstance = true;
    NativeToManagedMap[__Instance] = this;
}

protected HashMap(void* native, bool skipVTables = false)
    : base((void*) null)
{
    __PointerAdjustment = 0;
    if (native == null)
        return;
    __Instance = new global::System.IntPtr(native);
}
[...]

PR #1044, opened 2018-01-11
Fix for #1043

rokups:
My attempt at least. I am unsure this is entirely correct. Especially `if (!field.Type.IsPrimitiveType() && !(field.Type is TagType))` check. Please verify.

P.S. also slipped in fix in `CSharpSourcesExtensions.cs` where invalid code was generated. Error complained about compiler being unable to deduce type.

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpSources.cs
+++ b/src/Generator/Generators/CSharp/CSharpSources.cs
@@ -906,13 +906,11 @@ namespace CppSharp.Generators.CSharp
             else
             {
                 var name = @class.Layout.Fields.First(f => f.FieldPtr == field.OriginalPtr).Name;
-                var printed = TypePrinter.PrintNative(@class);
-                ctx.ReturnVarName = string.Format("{0}{1}{2}",
-                    @class.IsValueType
-                        ? Helpers.InstanceField
-                        : $"(({printed}*) {Helpers.InstanceIdentifier})",
-                    @class.IsValueType ? "." : "->",
-                    SafeIdentifier(name));
+                var identifier = SafeIdentifier(name);
+                if (@class.IsValueType)
+                    ctx.ReturnVarName = $"{Helpers.InstanceField}.{identifier}";
+                else
+                    ctx.ReturnVarName = $"(({TypePrinter.PrintNative(@class)}*){Helpers.InstanceIdentifier})->{identifier}";
             }
             param.Visit(marshal);
 
--- a/src/Generator/Generators/CSharp/CSharpSources.cs
+++ b/src/Generator/Generators/CSharp/CSharpSources.cs
@@ -1182,7 +1182,7 @@ namespace CppSharp.Generators.CSharp
                 // IntPtr ensures that non-copying object constructor is invoked.
                 Class typeClass;
                 if (field.Type.TryGetClass(out typeClass) && !typeClass.IsValueType)
-                    returnVar = $"new global::System.IntPtr(&{returnVar})";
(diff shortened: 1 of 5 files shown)


tritao:
Can you add a test for this? It's likely to regress in the future without one.

ddobrev:
@rokups in addition, our CI has failed: https://ci.appveyor.com/project/tritao/cppsharp/build/1.0.2990/job/tn8n26ve3mjr08x5 .

rokups:
What is `TagType` though? I pulled that check out of debugger to skip enums. I am still not sure it is a correct check.

tritao:
Tag type is the base type for enums and struct/classes.

rokups:
How do i make VS2017 discover tests? For some reason it does not find any.

ddobrev:
https://marketplace.visualstudio.com/items?itemName=NUnitDevelopers.NUnit3TestAdapter . If it still doesn't work, rebuild.

rokups:
Have you ever experienced testcase on windows getting stuck? I cant really figure out which one it is.

Also if you have any workflow on debugging linux test crashes please do tell.

tritao:
If you have it crashing on the command line, then you can just add lldb on front of it to crash it on the debugger.

Something like: `lldb -- mono --debug nunit3-console.exe PathToTests.exe`

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -1174,13 +1172,24 @@ private static Property GetActualProperty(Property property, Class c)
...
+            String returnVar;
+            if (@class.IsValueType)
+                returnVar = $@"{Helpers.InstanceField}.{SafeIdentifier(name)}";
+            else
+            {
+                returnVar = $@"(({TypePrinter.PrintNative(@class)}*) {Helpers.InstanceIdentifier})->{SafeIdentifier(name)}";
+                // Class field getter should return a reference object instead of a copy. Wrapping `returnVar` in
+                // IntPtr ensures that non-copying object constructor is invoked.
+                var finalType = field.Type.Desugar(false);
+                if (!finalType.IsPrimitiveType() && !finalType.IsEnumType() && !finalType.IsAddress() && !(field.Type is ArrayType))
```
tritao: What is this check trying to achieve here?

Could we just check if the type is a non-value class type here?

rokups: Yeah i wanted this to be done for non-value classes, i guess i didnt put it into right words in my mind so ended up with this. Let me try

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -1174,13 +1172,24 @@ private static Property GetActualProperty(Property property, Class c)
...
+            if (@class.IsValueType)
+                returnVar = $@"{Helpers.InstanceField}.{SafeIdentifier(name)}";
+            else
+            {
+                returnVar = $@"(({TypePrinter.PrintNative(@class)}*) {Helpers.InstanceIdentifier})->{SafeIdentifier(name)}";
+                // Class field getter should return a reference object instead of a copy. Wrapping `returnVar` in
+                // IntPtr ensures that non-copying object constructor is invoked.
+                Class typeClass;
+                if (field.Type.TryGetClass(out typeClass) && !typeClass.IsValueType)
+                    returnVar = $"new global::System.IntPtr(&{returnVar})";
```
tritao: One last nit here that I remembered: You should use `CSharpTypePrinter.IntPtrType` here instead of hardcoding the type.

Issue #1051, opened 2018-01-19
Symbols generated for ignored methods of template classes


rokups:
Problematic code: https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Container/HashMap.h#L336-L347

I attempted to ignore these methods:
```cs
cls = ctx.FindCompleteClass("HashMap");
if (cls != null)
{
    foreach (var method in cls.Methods.Where(method => method.Name == "Populate"))
        method.ExplicitlyIgnore();
}
```
`ExplicitlyIgnore()` is called twice as one would expect, despite this symbols file contains lots of instances similar to  this:
```cpp
template Urho3D::HashMap<Urho3D::TextureUnit, Urho3D::SharedPtr<Urho3D::Texture>>& Urho3D::HashMap<Urho3D::TextureUnit, Urho3D::SharedPtr<Urho3D::Texture>>::Populate(const Urho3D::TextureUnit&, const Urho3D::SharedPtr<Urho3D::Texture>&);
template Urho3D::HashMap<Urho3D::StringHash, Urho3D::MaterialShaderParameter>& Urho3D::HashMap<Urho3D::StringHash, Urho3D::MaterialShaderParameter>::Populate(const Urho3D::StringHash&, const Urho3D::MaterialShaderParameter&);
```
It would not be a problem except that these specializations create ambiguous template specialization errors:
```
../../Urho3DNet/Urho3DNet-symbols.cpp:1379:64: error: ambiguous template specialization ‘Populate<>’ for ‘Urho3D::HashMap<Urho3D::StringHash, Urho3D::Variant>& Urho3D::HashMap<Urho3D::StringHash, Urho3D::Variant>::Populate(const Urho3D::StringHash&, const Urho3D::Variant&)’
 template Urho3D::HashMap<Urho3D::StringHash, Urho3D::Variant>& Urho3D::HashMap<Urho3D::StringHash, Urho3D::Variant>::Populate(const Urho3D::StringHash&, const Urho3D::Variant&);
                                                                ^~~~~~
In file included from ../../Urho3DNet/Urho3DNet-symbols.cpp:5:0:
/tmp/Urho3D-SDK/include/Urho3D/Container/HashMap.h:337:14: note: candidates are: Urho3D::HashMap<T, U>& Urho3D::HashMap<T, U>::Populate(const T&, const U&) [with T = Urho3D::StringHash; U = Urho3D::Variant]
     HashMap& Populate(const T& key, const U& value)
[...]

PR #1053, opened 2018-01-20
Fix for #1051

rokups:
As title says.

Also slipped in benign change adding extra include path for linux. Fixes error when including `limits.h`.

### DIFF
--- a/src/Generator/Passes/GenerateSymbolsPass.cs
+++ b/src/Generator/Passes/GenerateSymbolsPass.cs
@@ -55,7 +55,8 @@ namespace CppSharp.Passes
                             !m.IsImplicit && !m.IsDeleted && !m.IsDefaulted;
                         if (specialization.Methods.Any(m => m.IsInvalid && exportable(m)))
                             foreach (var method in specialization.Methods.Where(
-                                m => m.IsGenerated && exportable(m)))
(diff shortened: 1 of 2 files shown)


src/Generator/Passes/GenerateSymbolsPass.cs
```diff
@@ -55,7 +55,7 @@ private void GenerateSymbols()
                             !m.IsImplicit && !m.IsDeleted && !m.IsDefaulted;
                         if (specialization.Methods.Any(m => m.IsInvalid && exportable(m)))
                             foreach (var method in specialization.Methods.Where(
-                                m => m.IsGenerated && exportable(m)))
+                                m => m.IsGenerated && m.InstantiatedFrom.IsGenerated && exportable(m)))
```
ddobrev: As I warned you at the beginning, InstantiatedFrom may be null. You do not check it as I asked and the result is precisely https://ci.appveyor.com/project/tritao/cppsharp/build/1.0.3005/job/896ctkux248jd7y8#L497 . In addition, I understand you prefer to work on Linux but our tests can run there too. I have a feeling you don't ensure they do before sending code, not just in this particular pull request. Please correct me if I'm wrong. If not, please understand that valid tests mean work less work not just for me but for you as well.

rokups: NUnit actually crashes on my machine. Sorry about that. Please see if new commit is any better.

PR #1062, opened 2018-01-30
Implement xcode-select based path selection

### LINKED ISSUE #1061: GetXcodePath() should use xcode-select to retrieve the configured xcode path
##### Brief Description
[`XcodeToolchain.GetXCodePath()`](https://github.com/mono/CppSharp/blob/master/src/Core/Toolchains/XcodeToolchain.cs#L9) currently enumerates the folders in `"/Applications"` which start with the string `"XCode"` and chooses the last based on the sort order.

Because system-included headers are based on this value, the results can cause potential issues with developers who have multiple versions of XCode installed on their machine. For example, the stable and beta versions.

An ideal solution would use the output from `xcode-select --print-path` and default to the current behavior only when `xcode-select` fails.

OS: OS X (pretty much every version)

Target: GCC/Clang

KallynGowdyYetiCGI:
Fixes #1061.

When merged, this PR will improve `XcodeToolchain.cs` to use `xcode-select` instead of inferring where the XCode SDK exists. If `xcode-select` fails for some reason, we default to the original behavior.

### DIFF
--- a/src/Core/Toolchains/XcodeToolchain.cs
+++ b/src/Core/Toolchains/XcodeToolchain.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Diagnostics;
 using System.Linq;
 
 namespace CppSharp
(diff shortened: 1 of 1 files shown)


tritao:
Thanks, have always wanted to do this but never got around to it 👍

PR #1089, opened 2018-05-17
#1087 Fix stack overflow unless VS2017 is not installed

### LINKED ISSUE #1087: Stack overflow unless VS2017 is not installed
##### Brief Description

I built CppSharp on VS2017, but it doesn't work unless VS2017 is not installed even if VS2015 is installed.

OS: Windows
Target: MSVC

I found the cause is here:
https://github.com/mono/CppSharp/blob/master/src/Core/Toolchains/MSVCToolchain.cs#L168

It calls itself to find VS2017 but it cannot get valid vsSdk, then it calls itself to find **VS2017** again.

limenote135:
Fixes #1087


limenote135:
Thank you for the checking. I'll use my change for now.
Please let me know if there are any problems!

tritao:
Hey @limenote135.

I think we merged some VS fixes a little while ago.

Could you check if there is still an issue with latest revisions?

Issue #1096, opened 2018-05-21
C++ string parameter instance duplicates


mattparks:
##### Brief Description
The C# generated output from a virtual method that takes two strings (or three) as parameters generates duplicate names for string instances. 

C++ input: https://pastebin.com/HrBkVT87
Generated C# output: https://pastebin.com/4CyMcfag

Errors:
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0128	A local variable or function named '__basicStringRet' is already defined in this scope	TestSharp	C:\Users\mattp\Documents\Flounder Workspace\TestSharp\TestSharp\Test.cs	110	Active
```

OS: Windows
Target: MSVC

Driver: https://pastebin.com/vAWeTgua

tritao:
Thanks for the bug report.

The issue is in this code:

https://github.com/mono/CppSharp/blob/695a7538d11db4fe7fef6c1f1d77dd0284267ce7/src/Generator/Types/Std/Stdlib.cs#L101

https://github.com/mono/CppSharp/blob/695a7538d11db4fe7fef6c1f1d77dd0284267ce7/src/Generator/Types/Std/Stdlib.cs#L112

I think this just needs to be changed to something like this:

```
const string varBasicString = $"__basicStringRet{ctx.ParameterIndex}";
```

Do you want to try sending a PR fixing this issue?

PR #1097, opened 2018-05-21
Added parameter index to managed marshal variables

mattparks:
Fix for issue https://github.com/mono/CppSharp/issues/1096


tritao:
The code as-is does not compile, we also need a new test-case for this issue to merge the fix.

mattparks:
So I fixed the build error, what test do you think this should be added into?

tritao:
Try here in `ImplementsVirtualFunctionsWithStringParams `.

https://github.com/mono/CppSharp/blob/master/tests/Common/Common.h#L1340
https://github.com/mono/CppSharp/blob/master/tests/Common/Common.cpp#L896
https://github.com/mono/CppSharp/blob/master/tests/Common/Common.Tests.cs#L803

mattparks:
Ok added tests for this into `ImplementsVirtualFunctionsWithStringParams`, appears to have worked properly.

tritao:
👍

PR #1192, opened 2019-03-13
fixes #1191 CppSharp.CLI.exe --rtti sets -fno-rtti to clang

### LINKED ISSUE #1191: CppSharp.CLI.exe --rtti sets -fno-rtti to clang
##### Brief Description
The command line tool seems to not process --rtti correctly. Even if the parameter --rtti is given the parameter is lost on its way to clang.

OS: Windows

I have used the pre-compiled version inside the nuget package (packages\CppSharp.0.9.0\tools\CppSharp.CLI.exe) but I remember having the same issue on linux.

codingdave:
The problem being fixed is that although --rtti is given as argument to
CppSharp.CLI.exe the runtime type information is turned off. The error
is happening because the Generator options are copied over to the
parserOptions after the parserOptions have been used to generate the
cmake call. Thus, the default value of RTTI (off) is always evaluated
and cmake will always get the parameter -fno-rtti.
The same issue seems to apply to the unity build.

This patch fixes the issue by moving the 2 assignments of Generator
options properties into ParserOptions properties before they are
evaluated in parserOptions.Setup();

### DIFF
--- a/src/CLI/Generator.cs
+++ b/src/CLI/Generator.cs
@@ -152,6 +152,9 @@ namespace CppSharp
             if (abi == CppAbi.Microsoft)
                 parserOptions.MicrosoftMode = true;
 
+            parserOptions.UnityBuild = options.UnityBuild;
+            parserOptions.EnableRTTI = options.EnableRTTI;
+
             parserOptions.Setup();
 
             if (triple.Contains("linux"))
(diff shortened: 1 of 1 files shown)


tritao:
Awesome, thanks for the fix.

By the way, just to make things clear.
You have made some references to CMake, but it is not used by CppSharp at runtime.
Its only usage is to generate the projects for the initial LLVM/Clang build.

codingdave:
Ah, yes, sorry. I ment clang instead of cmake.

Issue #1203, opened 2019-04-14
Build fails on MacOS 10.14.4 with Xcode 10.2


Alovchin91:
I can't build CppSharp on MacOS 10.14.4 with Xcode 10.2 anymore.

The error I get is as follows:

```
==== Building Encodings.CSharp (release_x64) ====
Running prebuild commands
mono64 --debug "../lib/Release_x64/Encodings.Gen.exe"

Generating bindings for Encodings (CSharp)
Looking for tests in: /tmp/CppSharp-0.9.1/tests/Encodings

Unhandled Exception:
System.EntryPointNotFoundException: _ZNKSt3__112basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEE5c_strEv
  at (wrapper managed-to-native) Std.BasicStringExtensions+__Internal.CStr(intptr)
  at Std.BasicStringExtensions.CStr (Std.BasicString`3[_CharT,_Traits,_Allocator] this) [0x0001f] in /tmp/CppSharp-0.9.1/src/CppParser/Bindings/CSharp/x86_64-apple-darwin12.4.0/Std.cs:976 
  at CppSharp.Parser.CppParserOptions.get_TargetTriple () [0x0001e] in /tmp/CppSharp-0.9.1/src/CppParser/Bindings/CSharp/x86_64-apple-darwin12.4.0/CppSharp.CppParser.cs:47531 
  at CppSharp.Parser.ParserOptions..ctor (CppSharp.Parser.ParserOptions options) [0x00028] in /tmp/CppSharp-0.9.1/src/Parser/ParserOptions.cs:71 
  at CppSharp.Parser.ParserOptions.BuildForSourceFile (System.Collections.Generic.IEnumerable`1[T] modules, System.String file) [0x0000f] in /tmp/CppSharp-0.9.1/src/Parser/ParserOptions.cs:102 
  at CppSharp.Driver.ParseCode () [0x00123] in /tmp/CppSharp-0.9.1/src/Generator/Driver.cs:147 
  at CppSharp.ConsoleDriver.Run (CppSharp.ILibrary library) [0x00090] in /tmp/CppSharp-0.9.1/src/Generator/Driver.cs:428 
  at CppSharp.Tests.EncodingsTestsGenerator.Main (System.String[] args) [0x00007] in /tmp/CppSharp-0.9.1/tests/Encodings/Encodings.cs:27 
[...]

PR #1204, opened 2019-04-18
Fix Xcode 10.2+ build

Alovchin91:
This change fixes the issue described in #1203.

The old macro is left intact to ensure compatibility with older versions of Xcode.

There is no reason to make this change for x86 build but consistency.


ddobrev:
Thank you @Alovchin91 . However, this code is generated so we won't take a manual fix. Instead, here's what you need to do:
1. Add your change to https://github.com/mono/CppSharp/blob/0b825e11246e5ddb0e41305f13379b61f1bea995/src/Generator/Passes/SymbolsCodeGenerator.cs#L23 instead;
2. Set Parser.Gen as your starting project and run it - it's going to make all changes for you.
Please don't add new commits on top but rather push force - or open a new PR if you prefer.

Alovchin91:
@ddobrev Sure, no problem. Should I also commit the generated files (Std-symbols.cpp)?

ddobrev:
@Alovchin91 of course, along with anything else changed.

Issue #1144, opened 2018-12-27
ArgumentNullException while parse void(*func2)(void(*bug)());


AleXr64:
##### Brief Description
Windows 10 1803 build 17134.472
ArgumentNullException at [DelegatesPass.cs#L180](https://github.com/mono/CppSharp/blob/master/src/Generator/Passes/DelegatesPass.cs#L180)

##### Used headers
```
struct buggy
{
	void(*func2)(void(*bug)());
};
void func2(void(*bug)());
```

##### Used settings
Master branch, x86
Target: MSVC ver 14.16.27023
WinSDK ver 10.0.17763.0
```
ILibrary.Setup()
{
            var options = driver.Options;
            driver.ParserOptions.SetupMSVC();
            options.GeneratorKind = GeneratorKind.CSharp;
            var module = options.AddModule("Buggy");
            module.IncludeDirs.Add("C:\\sdk2\\");
            module.Headers.Add("Header.h");
            options.OutputDir = "C:\\sdk2\\";
}
```
Other defaults

AleXr64:
I made ugly [workaround](https://github.com/mono/CppSharp/compare/master...AleXr64:master), but don't know how do this better.
[Generated file](https://gist.github.com/AleXr64/2d7e71f468a4db5908b268ae9607cd50)

AleXr64:
```
#define CS_VALUE_TYPE
struct CS_VALUE_TYPE buggy
{
	void(*func2)(void(*bug)());
};
void func2(void(*bug)());
```
[Result](https://gist.github.com/AleXr64/0b1bb91633c0f360824036629ce687ca)
Problem with type of Action<> at lines 46, 51, 64

mwheeler:
Just confirming I too am encountering some variation of this bug, though crashing at a slightly different location within `DelegatePass`.  Due to the very similar nature of the problem, I've decided to add my report/trace to this existing ticket.

Parameter declarations whose function-pointer types are defined in-line with the parameter declarations, seem to lack a valid filename (and completely lack a valid a namespace and module - which causes issues around DelegatePass where modules and/or namespaces are assumed to always be valid/not-null).

However the suggested workaround above does not fix the issue as I encounter it.

[...]

PR #1206, opened 2019-04-20
Fix generating delegates for cases like https://github.com/mono/CppSharp/issues/1144

AleXr64:
```
struct fieldspointers
{
	int(*funcint)(int(*bug)());
};
```
[Result](https://gist.github.com/AleXr64/65b04836e000f4798db48df2cba88faf)
BTW, review requested

### DIFF
--- a/src/Generator/Passes/DelegatesPass.cs
+++ b/src/Generator/Passes/DelegatesPass.cs
@@ -94,6 +94,9 @@ namespace CppSharp.Passes
 
         public override bool VisitParameterDecl(Parameter parameter)
         {
+            if(parameter.Namespace?.TranslationUnit?.Module == null && namespaces.Count > 0)
+                parameter.Namespace = namespaces.Peek();
+
             if (!base.VisitDeclaration(parameter) || parameter.Namespace == null ||
                 parameter.Namespace.Ignore)
                 return false;
@@ -117,11 +120,16 @@ namespace CppSharp.Passes
 
         public override bool VisitFieldDecl(Field field)
         {
+            if (field.Namespace?.TranslationUnit?.Module != null)
+                namespaces.Push(field.Namespace);
+
             if (!base.VisitFieldDecl(field))
                 return false;
 
             field.QualifiedType = CheckForDelegate(field.QualifiedType,
                 field.Namespace);
+
+            namespaces.Clear();
 
             return true;
         }
@@ -297,5 +305,6 @@ namespace CppSharp.Passes
         /// iterating over it, so we collect all the typedefs and add them at the end.
         /// </summary>
         private readonly List<TypedefDecl> delegates = new List<TypedefDecl>();
+        private readonly Stack<DeclarationContext> namespaces = new Stack<DeclarationContext>();
     }
 }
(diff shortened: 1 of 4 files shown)


ddobrev:
@AleXr64 thank you for your work but I'm afraid our CI fails. Please ensure all tests pass for you before sending any code.

tritao:
Hey @AleXr64, as @ddobrev said, make sure all existing tests pass, and don't forget to add a new one for the issue that you are fixing.

Also looks like there's some unintended whitespace changes that should be fixed before merging.

ddobrev:
@AleXr64 I'm sorry for this back and forth but I'm afraid you've misunderstood @tritao . Indentation is not just quantity but kind too. You use a tab while we use 4 spaces. You can tell by the diff in this very PR.
In addition, when you feel you're done, please combine all of your commits into one and use your first commit message.

AleXr64:
I do not know why tests are not built on * nix. I guess that due to calling conventions other names are generated. But I do not know how to solve.

ddobrev:
@AleXr64 could you try this instead:

```cs
hasFunctionPtrField.FunctionPtrTakeFunctionPtrField = field => field(); 
Assert.That(hasFunctionPtrField.FunctionPtrTakeFunctionPtrField(() =>  42), Is.EqualTo(42));
```

AleXr64:
Now on OSX build and run. Linux fails with

> 1) Error : /home/travis/build/mono/CppSharp/build/gmake/lib/Release_x64/Common.Tests.CSharp.dll
> System.Runtime.Serialization.SerializationException : The object with ID 2 implements the IObjectReference interface for which all dependencies cannot be resolved. The likely cause is two instances of IObjectReference that have a mutual dependency on each other.
> --SerializationException

ddobrev:
@AleXr64 ignore this one, it's an error from the CI itself. Please just fix the remaining note in my review, combine your commits and add "Fixes <URL_to_your_issue>", this way it will be automatically closed.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -94,6 +94,8 @@ public override bool VisitFunctionDecl(Function function)
 
         public override bool VisitParameterDecl(Parameter parameter)
         {
+            if(parameter.Namespace?.TranslationUnit?.Module == null && namespaces.Count > 0)
+                parameter.Namespace = namespaces.Peek();
             if (!base.VisitDeclaration(parameter) || parameter.Namespace == null ||
```
tritao: Add a new line here.

src/Generator/Passes/DelegatesPass.cs
```diff
@@ -117,11 +119,15 @@ public override bool VisitProperty(Property property)
...
         {
+            if (field.Namespace?.TranslationUnit?.Module != null)
+                namespaces.Push(field.Namespace);
+
             if (!base.VisitFieldDecl(field))
                 return false;
 
             field.QualifiedType = CheckForDelegate(field.QualifiedType,
                 field.Namespace);
+            namespaces.Clear();
```
tritao: Same here.

tritao:
I added two small nits as well, just add some extra new lines. Apart from that it looks good, thanks.

PR #1214, opened 2019-05-02
Name anonymous types after the fields which use them

ddobrev:


src/Generator/Passes/CleanInvalidDeclNamesPass.cs
```diff
@@ -22,9 +23,31 @@ public override bool VisitClassDecl(Class @class)
...
+                return false;
+
+            field.Type.Visit(this, field.QualifiedType.Qualifiers);
+
+            Class @class;
+            if (field.Type.TryGetClass(out @class) && string.IsNullOrEmpty(@class.OriginalName))
+            {
+                @class.Name = field.Name;
+                foreach (var item in @class.Fields.Where(f => f.Name == @class.Name))
+                {
```
tritao: No need for {}, add new line.

src/Generator/Passes/CleanInvalidDeclNamesPass.cs
```diff
@@ -35,8 +58,11 @@ public override bool VisitEnumDecl(Enumeration @enum)
             if (!base.VisitEnumDecl(@enum))
                 return false;
 
-            CheckChildrenNames(@enum.Items,
-                string.IsNullOrEmpty(@enum.Name) ? 1 : 0);
+            Check(@enum.Items);
+            foreach (Enumeration.Item item in @enum.Items.Where(i => char.IsNumber(i.Name[0])))
```
tritao: No need for {}, add new line.

src/Generator/Passes/CleanInvalidDeclNamesPass.cs
```diff
@@ -117,17 +124,25 @@ private void CheckEnumName(Enumeration @enum)
...
+                nameBuilder[i] = char.ToLowerInvariant(nameBuilder[i]);
+            }
+            field.Name = nameBuilder.ToString();
+        }
 
-            return name;
+        private static void Check(IEnumerable<Declaration> decls)
+        {
+            var anonymousDecls = decls.Where(p => string.IsNullOrEmpty(p.Name)).ToList();
+            for (int i = 0; i < anonymousDecls.Count; i++)
```
tritao: No need for {}

tests/Common/Common.h
```diff
@@ -825,30 +825,6 @@ DLL_API void va_listFunction(va_list v);
 struct DLL_API TestNestedTypes
 {
 public:
-    struct
```
tritao: Move this to the new header, and comment it out with an explanation that we can't handle such code yet.

src/Generator/Passes/CleanInvalidDeclNamesPass.cs
```diff
@@ -22,9 +23,31 @@ public override bool VisitClassDecl(Class @class)
 
             if (@class.Layout != null)
             {
-                int order = 0;
-                foreach (var field in @class.Layout.Fields)
-                    field.Name = CheckName(field.Name, ref order);
+                var fields = @class.Layout.Fields.Where(
+                    f => string.IsNullOrEmpty(f.Name)).ToList();
+                for (int i = 0; i < fields.Count; i++)
```
tritao: Newline between statements

tests/Common/Common.h
ddobrev: I actually like this solution.

PR #1239, opened 2019-08-08
Fixed out of range exception when property name equals "Get".

martin-tuor:
Access on string array `name[3] ` generates an `IndexOutOfRangeException` when the given parameter euqals "Get".


ddobrev:
@martin-tuor I appreciate the work but would you be able to add a test too? Please use our CSharp test project.

PR #1310, opened 2020-03-31
Ensure typedefs of std::vector are mapped

IIFE:
We have situations where we are using a `std::vector` typedef. Make sure `Desugar` is called on the type before it is marshalled, so that we get the underlying vector type of the typedef.


tritao:
Thanks for the contribution.

PR #1311, opened 2020-03-31
Marshal non primitive fixed arrays.

IIFE:
Non primitive arrays are not being marshalled correctly for C++/CLI.

### DIFF
--- a/src/Generator/Generators/CLI/CLIMarshal.cs
+++ b/src/Generator/Generators/CLI/CLIMarshal.cs
@@ -52,6 +52,8 @@ namespace CppSharp.Generators.CLI
                     if (array.Type.IsPointerToPrimitiveType(PrimitiveType.Void))
                         Context.Before.WriteLineIndent("{0}[i] = new ::System::IntPtr({1}[i]);",
                             value, Context.ReturnVarName);
+                    else if (!array.Type.IsPrimitiveType())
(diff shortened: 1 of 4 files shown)


ddobrev:
@IIFE thank you for adding a test too but may I ask you to also extend your test to the run-time too? That is, add a unit test which reads from and writes to the generated array.
Please do not send the test in a new commit but rather amend and force-push.

IIFE:
@ddobrev What project do I need to use to add unit tests that test generated code?

ddobrev:
`Common.Tests.cs`

IIFE:
Is it possible to run the tests locally on Windows?

ddobrev:
@IIFE yes, with any NUnit test runner.

tritao:
Regarding Travis, it's hit by a failure that happens once in a while, you can safely ignore it.

IIFE:
@ddobrev @tritao Test added. Thanks.

tritao:
Thanks, Travis now has a real failure:

```
1) Error : CommonTests.TestNonPrimitiveFixedArrays
System.EntryPointNotFoundException : _ZN16NonPrimitiveType6GetFooEv
```

You need to move the `GetFoo` method definition to the source file (`.cpp`).

ddobrev:
@IIFE one final change, please: move the implementations of your functions to *.cpp. When functions are defined in the header, they're sometimes inlined and the work to handle such is ongoing.

IIFE:
@ddobrev @tritao Done.

PR #1312, opened 2020-03-31
add cli namespace to header to ensure the array type does not conflicts

IIFE:
CppSharp.h uses the cli array type, however it does not prefix it with the cli namespace. This causes conflicts with std array, or other types called array.


ddobrev:
Looks just fine, I'm just going to wait for the Windows CI to complete.

PR #1313, opened 2020-03-31
Distinguish free functions class names

IIFE:
A scenario can arise where a translation unit has inside it a class with the same name as the unit file, as well as free functions under a namespace. The CLI generator will generate a class for the free functions, with the same name as the translation unit. This causes a conflict because there is already a class with the same name. Make the generated free functions class name unique by appending the word "Helpers".

Not an issue for C# as it generates partial classes.

### DIFF
--- a/include/CppSharp.h
+++ b/include/CppSharp.h
@@ -118,7 +118,7 @@ namespace clix {
         }
 
         // Copy the C++ string contents into a managed array of bytes
-        array<unsigned char> ^bytes = gcnew array<unsigned char>(byteCount);
+        cli::array<unsigned char> ^bytes = gcnew cli::array<unsigned char>(byteCount);
         { pin_ptr<unsigned char> pinnedBytes = &bytes[0];
           memcpy(pinnedBytes, cxxString.c_str(), byteCount);
         }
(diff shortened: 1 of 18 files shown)


ddobrev:
The CI broke, there's some problem. Also, I'm not sure this is the way we would like this solved. I'll discuss it with @tritao .

IIFE:
@ddobrev No problem. Let me know what you feel is better.

IIFE:
@ddobrev The CI error has this line:
```
CLI.Tests.cs(24,44): error CS0234: The type or namespace name 'FreeFunctions' does not exist in the namespace 'CLI' (are you missing an assembly reference?) [C:\projects\cppsharp\tests\CLI\CLI.Tests.CLI.csproj]
```

Do new generated files get included automatically in in the gen projects, or do I need to do something special? I think the error is because the free functions generated file is not getting compiled.

ddobrev:
@IIFE might be the case. However, if @tritao agrees with the solution I'm going to propose, there won't be extra files.

ddobrev:
@IIFE we might be able to come up with an even better solution than before, even for C#. However, we want to discuss in in more detail tomorrow. Meanwhile there's something you can do to help us: [generate C++/CLI classes partial too](https://docs.microsoft.com/en-us/cpp/extensions/partial-cpp-component-extensions) and let us know if and how it works with your test but without your fix.

ddobrev:
@IIFE I've just tried it myself, don't bother. It's only supported in windows rt. So let me tell you what the problem is. We have logic which moves free functions to classes as static methods in such cases. However, our searching for a matching class is apparently inadequate and fails when a name-space stands in the way. So our real fix is to improve the pass [which you can find here.](https://github.com/mono/CppSharp/blob/master/src/Generator/Passes/MoveFunctionToClassPass.cs)

ddobrev:
@IIFE there are some other complications as well, unfortunately. So please leave this for now and we'll talk more about it tomorrow.

IIFE:
@ddobrev That's fine. I'll continue posting the other PRs in the meantime. Thanks.

tritao:
Me and @ddobrev have discussed this a bit, but have not yet had time to come up with a solution we're both happy with. I would ask you to continue sending other PRs in the meantime and we'll try to figure something out.

IIFE:
@tritao Thanks for the update. I can temporarily work around the issue, but it would be good to have a proper fix.

Any idea when it's possible to get a release with the merged fixes please, as I'm unable to find NuGet or binaries with the latest changed?

Thanks.

ddobrev:
@IIFE https://www.nuget.org/packages/CppSharp/0.10.3. Thank you for your contribution for this release. It hasn't been indexed at the moment but will be in a few minutes.

IIFE:
@tritao @ddobrev I had more thought about this, and I'm starting to think we may not need a fix on CppSharp side.

The issue is pretty much an edge case, and can be fixed in the CppSharp consumer by simply wrapping the free functions in a namespace so that they don't conflict with existing types that match the file name.

I'm happy to close this PR.

ddobrev:
@IIFE it actually can and should be fixed, it's just that we haven't agreed on the approach. So please leave the PR open to serve as a reminder.

PR #1318, opened 2020-04-01
Ctor, dtor and finalizer body blocks

IIFE:
Add body blocks for ctor, dtor and finalizer to allow consumer to modify these methods bodies.

Create a ctor overload that takes a bool from the caller to tell the wrapper if it should own the native pointer passed to it, therefore making it responsible for deleting it later, or the caller is the only owner of the pointer.

Didn't think there were any tests needed as it is only generating a new ctor, which can be verified by looking at the generated code. How the ctor is used is up to the consumer.

### DIFF
--- a/src/Generator/Generators/CLI/CLIHeaders.cs
+++ b/src/Generator/Generators/CLI/CLIHeaders.cs
@@ -376,6 +376,7 @@ namespace CppSharp.Generators.CLI
 
             // Output a default constructor that takes the native pointer.
             WriteLine("{0}({1} native);", @class.Name, nativeType);
+            WriteLine("{0}({1} native, bool ownNativeInstance);", @class.Name, nativeType);
             WriteLine("static {0}^ {1}(::System::IntPtr native);",
                 @class.Name, Helpers.CreateInstanceIdentifier);
 
--- a/src/Generator/Generators/CLI/CLIMarshal.cs
+++ b/src/Generator/Generators/CLI/CLIMarshal.cs
@@ -275,15 +275,19 @@ namespace CppSharp.Generators.CLI
             instance += Context.ReturnVarName;
             var needsCopy = Context.MarshalKind != MarshalKind.NativeField;
 
+            bool ownNativeInstance = false;
+
             if (@class.IsRefType && needsCopy)
             {
                 var name = Generator.GeneratedIdentifier(Context.ReturnVarName);
                 Context.Before.WriteLine("auto {0} = new ::{1}({2});", name,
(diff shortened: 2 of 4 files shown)


tritao:
So to continue discussion from the other PR:

For each specific block, we already have `PushBlock(BlockKind.Finalizer)` and  `PushBlock(BlockKind.Destructor)`, and which also have a general `BlockKind.MethodBody` inside.

So I can't why we need yet another block kind specific for each method body.

The only problem with the actual situation, is that for methods, the code looks like this:

```
            PushBlock(BlockKind.Method, method);
            if (method.IsConstructor || method.IsDestructor ||
                method.OperatorKind == CXXOperatorKind.Conversion ||
                method.OperatorKind == CXXOperatorKind.ExplicitConversion)
                Write("{0}::{1}(", QualifiedIdentifier(@class), GetMethodName(method));
            else
                Write("{0} {1}::{2}(", method.ReturnType, QualifiedIdentifier(@class),
                      method.Name);
```

So it only pushes a general `BlockKind.Method`, but you could update this code instead to generate a more specific `BlockKind.Constructor/Destructor` if needed. Alternatively, you could also look at the `Method` instance stored inside the method block to see what kind of method it is.

IIFE:
@tritao The existing blocks are not for the **body** part of the method. It's much easier to only intercept the body generation and change that as the consumer may only be interested in the body, rather than the parameters declaration.

The existing `MethodBody` does not cover finalizer, constructor, or destructor body for CLI as far as I can see.

tritao:
> @tritao The existing blocks are not for the **body** part of the method. It's much easier to only intercept the body generation and change that as the consumer may only be interested in the body, rather than the parameters declaration.
> 
> The existing `MethodBody` does not cover finalizer, constructor, or destructor body for CLI as far as I can see.

Correct, what I was suggesting is for you to add `MethodBody` to cover them as well.

Wouldn't that work for your use case?

IIFE:
@tritao Don't think it will because there is only context reference to the class rather than the method being generated, so it won't be possible to know from the consumer what method the method body is for. There is also no way to check if a method is a finalizer. Consumers might also only care about the class context rather than only the method being the block context.

This felt the simplest change to allow us to intercept the generation of the required methods. Using `MethodBody` will require passing the method to various methods inside `CLISources` that generate the bodies, as well as having a context that has both the class and the method.

tritao:
OK, I don't think the separate method body kinds for constructors, destructors and finalizers are really needed, and does make the code a little more complex than  needed, but I'm fine with it if it makes it easier on your side.

PR #1319, opened 2020-04-02
Move the GenerateInclude logic to GetTypeReference

IIFE:
We have the following situation that causes us an error without this fix:
- Class template `Template` that say takes a pointer of type `Foo`.
- A header that forward declares Foo and a typedef for `Template<Foo>`.
- Another class `Bar` that makes use of the typedef from the forward decl header.
- The class template is ignored by default, however we create a type map for it to extract the template type and map it to the generated CLI type. For example we create a type map for `Template` class template, and marshal the template type, in this case `Foo` to the generated CLI `Foo`.

The problem is because `Bar` cannot see `Foo` directly (it's forward declared), even though a CLI class for `Bar` is generated, the code generation does not add a `Bar.h` include so that the generated `Foo.cpp` can see the generated `Bar` class, which it now uses because we have mapped the class template to return the template type.

The issue is caused by `GetTypeReference` not creating a `CInclude` for `Bar.h` into `Foo.cpp`.

Hope it makes sense.


IIFE:
@ddobrev @tritao Hi both. Thanks so much for your help so far in getting the PRs merged.

Any chance we could have a look at this PR today as well? Got few other PRs to go through, but I'm trying to do them in order to make it more manageable.

Your help is greatly appreciated.

Thanks.

tritao:
I'll review this in a few hours, feel free to send more PR meanwhile.

IIFE:
@tritao Awesome cheers. Will start sending more PRs then. Thanks.

src/Generator/Generators/CLI/CLITypeReferences.cs
```diff
@@ -48,10 +48,27 @@ public CLITypeReference GetTypeReference(Declaration decl)
...
+            if (ShouldIncludeTranslationUnit(translationUnit) && decl.IsGenerated && !IsBuiltinTypedef(decl))
+            {
+                var @ref = new CLITypeReference { Declaration = decl };
+
+                @ref.Include = new CInclude
+                {
+                    File = GetIncludePath(translationUnit),
+                    TranslationUnit = translationUnit,
+                    Kind = translationUnit.IsGenerated
+                    ? CInclude.IncludeKind.Quoted
```
tritao, line 62: This should be indented either one level up for both the lines, or you can keep the `?` in the previous line, and have the rest of the conditional expression on a single line, indented.

tests/CLI/CLI.cs
```diff
@@ -1,9 +1,58 @@
...
+            return new CustomType("CLI::Employee^");
+        }
+
+        public override void CLIMarshalToManaged(MarshalContext ctx)
+        {
+            ctx.Return.Write($"gcnew CLI::Employee({ctx.ReturnVarName}.m_employee)");
+        }
+    }
+
+    public class CompleteIgnoredClassTemplateForEmployeeTypedefPass : TranslationUnitPass
```
tritao, line 24: Can you explain the meaning of this pass? 

Is this to workaround any bug in the generator or just to force what you need from the test?

IIFE: Just to force what I need from the test.

src/Generator/Generators/CLI/CLITypeReferences.cs
```diff
@@ -104,39 +121,27 @@ private void GenerateInclude(ASTRecord<Declaration> record)
             if (decl.Namespace == null)
                 return;
 
+            var typedefType = record.Value as TypedefNameDecl;
+
             // Find a type map for the declaration and use it if it exists.
             TypeMap typeMap;
-            if (TypeMapDatabase.FindTypeMap(record.Value, out typeMap))
+            if (TypeMapDatabase.FindTypeMap(record.Value, out typeMap) 
+                || (typedefType != null && TypeMapDatabase.FindTypeMap(typedefType.Type.Desugar(), out typeMap)))
```
tritao, line 129: Just FYI this should be fixed inside `FindTypeMap` itself.

I do have a PR that I intend to land soon that rework type map handling and so we should be able to remove this then.

IIFE: So do you want to keep this in the commit until your PR, or remove it from this commit and wait for your PR? I would sort of prefer the former to unblock the work I'm doing if that's OK?

tritao: Yes, that's fine to keep it.

tritao:
Overall looks fine, not sure I like it that `GetTypeReference` can return `null` as part of its behaviour, but I'm not coming up with any decent alternatives either, so I think it's fine as a solution. I'd just ask you to fix that minor indentation remark, and I'll merge this in.

IIFE:
Done all required changes. Hopefully it's fine now. Thanks.

tritao:
Unfortunately GitHub says there is a conflict now due to merging the other PR.

Can you fix it so I can merge it?

IIFE:
@tritao Done

PR #1321, opened 2020-04-02
By reference enum param fixes

IIFE:
- Enum passed by reference is getting treated as a raw pointer, where enum is being dereferenced when it shouldn't.
- Full qualified enum name is not being used when generating enum param.

### DIFF
--- a/src/Generator/Generators/CLI/CLITypePrinter.cs
+++ b/src/Generator/Generators/CLI/CLITypePrinter.cs
@@ -149,7 +149,8 @@ namespace CppSharp.Generators.CLI
                 var typeName = VisitDeclaration(@enum, quals);
 
                 // Skip one indirection if passed by reference
-                if (Parameter != null && (Parameter.Type.IsReference() || ((Parameter.IsOut || Parameter.IsInOut)
+                if (Parameter != null && (Parameter.Type.IsReference() 
+                    || ((Parameter.IsOut || Parameter.IsInOut)
                     && pointee == finalPointee)))
                     return typeName;
 
--- a/src/Generator/Generators/CLI/CLITypePrinter.cs
+++ b/src/Generator/Generators/CLI/CLITypePrinter.cs
@@ -146,11 +146,11 @@ namespace CppSharp.Generators.CLI
             Enumeration @enum;
             if (pointee.TryGetEnum(out @enum))
             {
-                var typeName = @enum.Visit(this);
+                var typeName = VisitDeclaration(@enum, quals);
 
                 // Skip one indirection if passed by reference
-                if (Parameter != null && (Parameter.IsOut || Parameter.IsInOut)
-                    && pointee == finalPointee)
+                if (Parameter != null && (Parameter.Type.IsReference() || ((Parameter.IsOut || Parameter.IsInOut)
+                    && pointee == finalPointee)))
                     return typeName;
 
                 return $"{typeName}*";
--- a/tests/CLI/CLI.h
+++ b/tests/CLI/CLI.h
@@ -56,3 +56,15 @@ typedef struct IncompleteStruct IncompleteStruct;
 
 DLL_API IncompleteStruct* createIncompleteStruct();
 DLL_API void useIncompleteStruct(IncompleteStruct* a);
+
+enum EnumParam
+{
+    E1,
+    E2
+};
+
+class DLL_API TestByRefEnumParam
+{
+public:
+    EnumParam GetPassedEnumParam(EnumParam& e);
+};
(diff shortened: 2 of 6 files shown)


src/Generator/Generators/CLI/CLITypePrinter.cs
```diff
@@ -146,11 +146,11 @@ public override TypePrinterResult VisitDelegate(FunctionType function)
             Enumeration @enum;
             if (pointee.TryGetEnum(out @enum))
             {
-                var typeName = @enum.Visit(this);
+                var typeName = VisitDeclaration(@enum, quals);
```
ddobrev, line 149: What's the reason for this change?

src/Generator/Generators/CLI/CLITypePrinter.cs
```diff
@@ -146,11 +146,11 @@ public override TypePrinterResult VisitDelegate(FunctionType function)
...
 
                 // Skip one indirection if passed by reference
-                if (Parameter != null && (Parameter.IsOut || Parameter.IsInOut)
-                    && pointee == finalPointee)
+                if (Parameter != null && (Parameter.Type.IsReference() || ((Parameter.IsOut || Parameter.IsInOut)
```
ddobrev, line 152: This line is now too long, you could break it at `((Parameter.IsOut`.

tests/CLI/CLI.h
```diff
@@ -55,4 +55,16 @@ struct CompleteIncompleteStruct;
 typedef struct IncompleteStruct IncompleteStruct;
 
 DLL_API IncompleteStruct* createIncompleteStruct();
-DLL_API void useIncompleteStruct(IncompleteStruct* a);
\ No newline at end of file
+DLL_API void useIncompleteStruct(IncompleteStruct* a);
```
ddobrev, line 58: We try placing non-generator specific tests in `Common.h`. Could you please try moving it there?

src/Generator/Generators/CLI/CLITypePrinter.cs
IIFE: The generator does not generate fully qualified parameter name. For example instead of `CLI::EnumParam` it just generates `EnumParam`, which does not build.

tests/CLI/CLI.h
IIFE: The fix is specific to CLI because it is fixing pointer dereferencing.

src/Generator/Generators/CLI/CLITypePrinter.cs
ddobrev: We should perhaps fix the real issue, that is, make visiting an enum generate the qualified name. @tritao what do you think?

tests/CLI/CLI.h
ddobrev: @IIFE I meant that functions taking a pointer to an enum can be mapped to C# as well. So please try moving the test there as I expect no trouble from the C# gen, it can already handle such pointers.

src/Generator/Generators/CLI/CLITypePrinter.cs
IIFE: It does generate qualified name for other enum scenarios, does not seem to do it if enum is passed by ref, hence this fix.

tritao: You should be able to do `@enum.Visit(this, quals);`.

Can you give that a try?

IIFE: There is no `Visit` overload that takes a second quals param.

tritao: You're right, it's only supported for types, my bad.

tritao: @IIFE Can you try removing the following method from `CLITypePrinter`?

```csharp
        public override TypePrinterResult VisitEnumDecl(Enumeration @enum)
        {
            return @enum.Name;
        }
``` 

That will call into the base `CppTypePrinter`, which should print enums using the general qualified declaration code.

tritao:
This looks mostly good, I would just like for you to try my suggestion to use the base printer for enums. And if you could move the test to `Common`, as @ddobrev suggested, we'd definitely prefer as that increases test coverage of other backends aswell.

tests/CLI/CLI.h
IIFE: Just tried common test. Unfortunately the C# generation also generates pointer to by ref enum param, so moving the test to common will break unless the C# part is fixed, which is probably worth doing in another PR, so maybe keep the CLI tests?

src/Generator/Generators/CLI/CLITypePrinter.cs
IIFE: Removing that method causes StackOverflowException when generator is run.

src/Generator/Generators/CLI/CLITypePrinter.cs
IIFE: Done

tests/CLI/CLI.h
tritao: Thanks for checking, we can fix it later.

PR #1322, opened 2020-04-03
Include type ref header in consumer header if type is nested enum

IIFE:
This fixes an issues where if a consumer is using an enum that is nested in another class, the generated header for the consumer cannot see the enum because the header where the enum lives is not included.

P.S - I think this might be the last fix I need to get things working on my side. I'll see what happens, so there may be other possible future PRs.

When does the master branch normally get published to NuGet?

Thanks.

### DIFF
--- a/src/Generator/AST/ASTRecord.cs
+++ b/src/Generator/AST/ASTRecord.cs
@@ -150,6 +150,19 @@ namespace CppSharp.Generators.AST
             return field.Type.Desugar().TryGetClass(out decl) && decl.IsValueType;
         }
 
+        public static bool IsEnumNestedInClass(this ASTRecord<Declaration> record)
+        {
+            Enumeration @enum = record.Value as Enumeration;
+            var typedDecl = record.Value as ITypedDecl;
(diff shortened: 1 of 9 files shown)


IIFE:
appveyor CI appears to be failing with:
```
NUnit.Engine.NUnitEngineException: No connection could be made because the target machine actively refused it 127.0.0.1:1602
```

tritao:
> P.S - I think this might be the last fix I need to get things working on my side. I'll see what happens, so there may be other possible future PRs.

Thanks a lot for your contributions so far, appreciated.

> When does the master branch normally get published to NuGet?

We really do not have a designated release schedule, it's more occasionally once we feel we have a good amount of fixes. I think all this recent work certainly calls for one soon.

IIFE:
@tritao Thank you for the prompt replies and help in getting PRs merged, much appreciated.

Yes it would be great if there was a NuGet release available ASAP please so that I can commence with using CppSharp through NuGet (preferred option to using the raw release).

Thanks.

PR #1333, opened 2020-04-08
Ensure complete template specializations in AST

ddobrev:


tritao:
@IIFE FYI We found the incomplete template bug you were working around in the test and fixed it.

IIFE:
@tritao Awesome thanks. Any news on when the NuGet package will be released?

PR #1339, opened 2020-04-10
Use UnsupportedType description for type name instead of empty string

IIFE:
`CppTypePrinter.VisitTagType` has changed recently to use type maps, which consequently calls `CppSignatureType` instead of the previous `tag.Declaration.Visit(this)` call. This is breaking code generation for us as we are using `CustomType` for some tag types mapping.

The fix is to return the `UnsupportedType` (`CustomType` base) description instead of an empty string so that we can generate the correct type name, instead of `void*`, which is what is currently getting generated.

@tritao @ddobrev If you're happy with this PR, it would be really appreciated if we can get another NuGet release please. Unfortunately I just found this issue when testing the latest NuGet release.

Thank you.

### DIFF
--- a/src/Generator/Generators/C/CppTypePrinter.cs
+++ b/src/Generator/Generators/C/CppTypePrinter.cs
@@ -363,7 +363,7 @@ namespace CppSharp.Generators.C
         public override TypePrinterResult VisitUnsupportedType(UnsupportedType type,
             TypeQualifiers quals)
         {
-            return string.Empty;
+            return type.Description;
         }
 
         public override TypePrinterResult VisitDeclaration(Declaration decl,
--- a/tests/CLI/CLI.Tests.cs
+++ b/tests/CLI/CLI.Tests.cs
@@ -44,4 +44,16 @@ public class CLITests : GeneratorTestFixture
             Assert.AreEqual(ClassWithNestedEnum.NestedEnum.E1, consumer.GetPassedEnum(ClassWithNestedEnum.NestedEnum.E1));
         }
     }
+
+    [Test]
+    public void TestChangePassedMappedTypeNonConstRefParam()
+    {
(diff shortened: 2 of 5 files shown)


ddobrev:
@IIFE please add a test. Besides guarding against regressions, it would also help us understand the problem.

IIFE:
@ddobrev Done.

The pass/map in the test is required to simulate the issue we are facing.

ddobrev:
@IIFE please replace the pass with a usage of our AST search, for example https://github.com/mono/CppSharp/blob/fb31a85b903c9028106c33c6ddcf9f89bf4a2ddf/src/AST/ASTContext.cs#L93. It's much shorter and simpler solution.

IIFE:
@ddobrev Do you mean use Preprocess/Postprocess to call that function? How can I call it on a function inside a class?

ddobrev:
@IIFE this is correct, you need the call in Preprocess or Postprocess.
I sent you that link as an example. In the same class you can see how to find classes, methods, whatever you need.

IIFE:
@ddobrev Done. Found a helper method to set param usage on class method name.

IIFE:
@ddobrev Pushed another change as I was using ASTContext from driver rather than the provided one.

IIFE:
@ddobrev Thanks very much for the merge 👍 

Any timeline when a NuGet release will take place please?

Thank you.

ddobrev:
@IIFE you understand that I wouldn't like to release this often in order not to confuse users. I think, however, that there's a way to publish private (or unlisted) releases. So this I can look into tomorrow.

IIFE:
@ddobrev A private release would be fine as well. NuGet supports pre-releases, which serves the required purpose.

It's just unfeasible to wait a long time for an important fix to become available through NuGet.

Thanks.

ddobrev:
@IIFE here you are: https://www.nuget.org/packages/CppSharp/0.10.4-alpha. Thank you for your contribution, I encourage you to send more as necessary.

PR #1342, opened 2020-04-13
Fixes around pointers and const arrays

IIFE:
- Ensure constant arrays, especially in value types, are correctly marshalled from managed to native by using a for loop, same way that a loop is used when a return var name is available. Ensure exception is thrown if provided array does not match size of array that we will fill.

- Pass address of pointer type if the native function call expects a pointer to pointer.

- Desugar the parameter type before visiting it so that we end up visiting the actual type rather than the typedef. This is required for typedef of pointers to primitive types to work.

- Ensure vector is dereferenced in loop if the native call gives us a pointer.

### DIFF
--- a/src/Generator/Generators/CLI/CLIMarshal.cs
+++ b/src/Generator/Generators/CLI/CLIMarshal.cs
@@ -305,7 +305,7 @@ namespace CppSharp.Generators.CLI
                 Context.Return.Write("({0} == nullptr) ? nullptr : gcnew ",
                     instance);
 
-            Context.Return.Write("{0}(", QualifiedIdentifier(@class));
+            Context.Return.Write("::{0}(", QualifiedIdentifier(@class));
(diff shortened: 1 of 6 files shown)


ddobrev:
@IIFE your PR has combined 4 different features. This makes it difficult to review and respectively easy to miss an error. Therefore I'd ask you to split it in 1 PR per feature. For faster merging please start with the simplest one.

IIFE:
@ddobrev The reason I combined 4 different fixes is that they are all quite related and mainly around how pointers are handled (arrays are just pointers at the end of the day).

I can split it in separate PRs if you really want to, even though I think the fixes are related.

ddobrev:
@IIFE if they are, this simply means they build on top of each other. If you send them in the correct order, it's going to work just fine.

IIFE:
@ddobrev Following PRs created;
#1343 #1344 #1345 #1346

PR #1345, opened 2020-04-14
pointer to primitive typedef

IIFE:


ddobrev:
@IIFE I've merged one of your PR-s and there conflicts now, please resolve.

IIFE:
@ddobrev Updated PR with latest non-conflicting changes now.

IIFE:
@ddobrev Travis seems to be failing with:
```
../../../tests/Common/Common.Tests.cs(1092,16): error CS1502: The best overloaded method match for `CommonTest.Common.PointerToPrimitiveTypedefPointerTestMethod(ref long, long)' has some invalid arguments
```

However appveyor is building fine.

IIFE:
@ddobrev Would it be possible to have a look at the travis build please as I'm not familiar with it or why it's failing in xcode.

Thanks.

tritao:
This might explain it.

```../../../tests/Common/Common.Tests.cs(1092,63): error CS1503: Argument `#1' cannot convert `ref int' expression to type `ref long'```

**Comment by tritao** [comment]
We upgraded our Mono version on CI, to make sure it's not a compiler bug.

Still fails with Mono 6.8, so I'd say it looks likely to be a real bug here.

You are using an int in the test, while the native code expects a long.

**Comment by IIFE** [comment]
@tritao The problem is that for some reason the .NET metadata seems to think the function takes an int, even though in the generated CLI it is declared as a long. So I can't even pass a long from the C# test because it thinks the param needs to be an int.

**Comment by IIFE** [comment]
OK I've changed the native code to take a pointer to an int rather than a long. See if this fixes it.

Don't think this is an issue with the generator as the code is building for Windows.

**Comment by tritao** [comment]
Mind pulling the latest master to resolve the conflicts?

**Comment by tritao** [comment]
By the way, as for a review for the PR, I think it would be best if we can move that typedef processing to inside the marshaler itself. That's usually the preferred way to go about these kind of fixes.

Did you try going with that approach? If yes, what problems did you find? If not, mind giving it a try ?

Thanks

**Comment by IIFE** [comment]
@tritao I did give it a try and found it would just be a lot of changes to achieve the same thing that can be done with this small change in CLISources. Additionally, some marshalling for typedefs is not implemented in CLIMarshall. So it is a combination of missing implementation, as well as complexity of changing some of the existing marshalling.

The typedef marshalling is really only needed for function pointers. For everything else, we know that we're going to have to visit the type from the typedef anyway, so best thing in my opinion is to start with that type from CLISources.

**Comment by IIFE** [comment]
On another note, why has the conflict resurface when i just resolved it?

**Comment by ddobrev** [comment]
@IIFE it's entirely my fault, I apologize. I pushed to master what I thought was a simple fix but it broke the build. I had to then force-push to fix it.

**Comment by IIFE** [comment]
@ddobrev No problem. The confusing bit is that I resolve the conflicts but it keeps saying it's conflicting here.

**Comment by IIFE** [comment]
OK this seems to have sorted it.

**Comment by ddobrev** [comment]
@IIFE indeed, thank you for your patience. I'm also going to assure you that since I got my fix right this won't happen any more.

**Comment by IIFE** [comment]
@ddobrev No problem at all it's absolutely fine :)

File: src/Generator/Generators/CLI/CLISources.cs [hunk 1]
```diff
@@ -1173,10 +1173,23 @@ public struct ParamMarshal
             var argName = Generator.GeneratedIdentifier("arg") + paramIndex.ToString(CultureInfo.InvariantCulture);
 
             var isRef = param.IsOut || param.IsInOut;
+
+            var paramType = param.Type;
```
ddobrev, line 1177: Please add desugaring at the end, delete the rest of your change in this file and try again.

IIFE: can you elaborate please? At the end where exactly?

ddobrev:
@IIFE never mind, I've pushed your code: https://github.com/mono/CppSharp/commit/471b6d4d2c0a92d81d571dec0932cfca0f199481.

IIFE:
@ddobrev Your change will not work because it's not respecting function pointers.

Per my comment in this PR, we need to keep visiting typedef only if the typedef is not a function pointer typedef. Function pointers need the typedef.

ddobrev:
@IIFE I've restored the previous master but I can't work any more on this tonight. I hope you can excuse me. We'll continue tomorrow.

IIFE:
@ddobrev No problem, of course.

IIFE:
@tritao @ddobrev Hi both. Any chance we can progress this PR today please? Thanks.

tritao:
It's OK to merge but it's got conflicts.

I think due to the earlier force push you pulled.

You'll need to reset the branch to an earlier commit, and re-commit and push.

IIFE:
@tritao Github says there are no conflicts with base branch?

IIFE:
Unless I'm missing something?

tritao:
I see "This branch cannot be rebased due to conflicts", along with 8 commits, including ones from @ddobrev.

IIFE:
@tritao Hmm very confusing as it tells me there are no conflicts.

What I'll do is abandon this PR and create a new one with a fresh branch from master to avoid any confusion.

One sec...

IIFE:
@tritao New PR created at #1355

Thanks.

PR #1346, opened 2020-04-14
correctly marshal constant arrays

IIFE:


ddobrev:
@IIFE I've merged one of your PR-s and there conflicts now, please resolve.

IIFE:
@ddobrev @tritao I've updated this PR with latest changes from master.

I've purposely only put tests for CLI because C# generator appears to have issues with the test cases. Obviously it needs to be looked at and fixed, but probably best done as a separate PR so that we're not holding off the CLI fixes.

Thanks.

IIFE:
@tritao @ddobrev Would it be possible to progress this PR please? Let me know if you have any questions.

Thanks.

src/Generator/Generators/CLI/CLIMarshal.cs
```diff
@@ -789,23 +793,26 @@ private void MarshalValueClassProperty(Property property, string marshalVar)
...
                 Context.Before.Write(marshal.Context.Before);
 
-            Type type;
-            Class @class;
-            var isRef = property.Type.IsPointerTo(out type) &&
-                !(type.TryGetClass(out @class) && @class.IsValueType) &&
-                !type.IsPrimitiveType();
-
-            if (isRef)
+            if (!string.IsNullOrWhiteSpace(marshal.Context.Return))
```
tritao, line 796: Why did you add this check here?

It is my understanding that the `Return` here should always provide a valid value upon invoking the marshaler. If not, then it should an error, no?

tritao:
> @tritao @ddobrev Would it be possible to progress this PR please? Let me know if you have any questions.
> 
> Thanks.

Sorry for the delay in processing this one.

Looks mostly good, I have just one question I left as a review.

src/Generator/Generators/CLI/CLIMarshal.cs
IIFE: Hi. This is for the scenario where we are marshalling an array member of a value/struct type. The array member is set and marshalled inside VisitArrayType on line 426. There is no return involved in that case, so we should not set a return here.

Basically without the if you end up with code like this:
```
    auto _marshal0 = ::UnionNestedInsideStruct();
    if (value.SzText != nullptr)
    {
        if (value.SzText->Length != 10)
        {
            throw gcnew System::InvalidOperationException("Source array size must equal destination array size.");
        }
        for (int i = 0; i < 10; i++)
            _marshal0.szText[i] = value.SzText[i];
    }
    _marshal0.szText = ;
    ((::StructWithNestedUnion*)NativePtr)->nestedUnion = _marshal0;
```

Notice the `    _marshal0.szText = ;`.

IIFE:
@tritao No problem. Hopefully my answer to your question makes sense.

tritao:
Thanks again for the contribution.

PR #1347, opened 2020-04-14
Implement abstract templates to call virtuals

ddobrev:
This allows for, just as with regular classes, the calling of virtual methods of abstract templates returned by functions.


ddobrev:
Already merged by accident but fixed beforehand.

tritao:
Nice, thanks for this fix.

PR #1368, opened 2020-05-06
CreateInstance factory overload to pass flag informing wrapper if it should own the native pointer passed to it.

IIFE:


ddobrev:
@IIFE I apologize for the late reply. I would ask you for a test, also because we'd understand your intention better — If the functionality you need isn't specific to the generator, we would have it in the C# one too.

IIFE:
@ddobrev No problem. Unfortunately couldn't find a way to test it as there isn't an example of passing a raw pointer from the tests.

Unlike C++/CLI generator, the C# one has no constructor that takes a bool about ownership, therefore having a `__CreateInstance` overload with that flag would be of no benefit at the minute.

Thanks.

IIFE:
@ddobrev @tritao Any possible update on this please? Thanks.

PR #1373, opened 2020-05-18
Option to allow caller to specify it does not want unchanged output files to be modified. This supports incremental build in VS.

IIFE:
@ddobrev 
@tritao 

We care generating a significant number of C++/CLI files with CppSharp. Currently CppSharp will overwrite files in `OutputDir` every time the generator is run, even if the source native code has not changed. This prevents incremental builds from taking place because Visual Studio will recompile C++ files if their timestamp changes, even if their content is the same, which slows the build time.

This PR improves the situation by allowing the caller to specify a flag that tells CppSharp not to overwrite generated files if the file content is the same as the generated code.

### DIFF
--- a/src/Generator/Driver.cs
+++ b/src/Generator/Driver.cs
@@ -330,19 +330,9 @@ namespace CppSharp
 
             // We only compare contents if the generated code length is the same as the length of the existing file. If their lengths are different then they are
             // definitely no the same so we should overwrite.
-            if (fi.Exists && fi.Length == generatedCode.Length)
-            {
-                var existingFileContent = File.ReadAllText(file);
-
-                if(!existingFileContent.Equals(generatedCode))
-                {
-                    File.WriteAllText(file, generatedCode);
-                }
-            }            
-            else
-            {
+            if (!fi.Exists || fi.Length != generatedCode.Length ||
+                File.ReadAllText(file) != generatedCode)
                 File.WriteAllText(file, generatedCode);
-            }
         }
 
         private static readonly Dictionary<Module, string> libraryMappings = new Dictionary<Module, string>();
--- a/src/Generator/Driver.cs
+++ b/src/Generator/Driver.cs
@@ -316,11 +316,37 @@ namespace CppSharp
                     var fileRelativePath = $"{fileBase}.{template.FileExtension}";
 
                     var file = Path.Combine(outputPath, fileRelativePath);
-                    File.WriteAllText(file, template.Generate());
+                    WriteGeneratedCodeToFile(file, template.Generate());
                     output.TranslationUnit.Module.CodeFiles.Add(file);
 
                     Diagnostics.Message("Generated '{0}'", fileRelativePath);
                 }
+            }
+        }
+
+        private void WriteGeneratedCodeToFile(string file, string generatedCode)
+        {
+            var writeGeneratedCodeToFile = true;
+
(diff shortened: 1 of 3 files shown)


src/Generator/Options.cs
```diff
@@ -160,6 +160,13 @@ public Module AddModule(string libraryName)
 
         public bool UsePropertyDetectionHeuristics { get; set; } = true;
 
+        /// <summary>
+        /// Flag to indicate if we should overwrite unchanged generated code files in the <see cref="OutputDir"/>. When set to false this flag
+        /// allows incremental builds for the generated code (i.e. Visual Studio won't need to compile files if they don't change).
+        /// Defaults to true to keep existing behaviour.
+        /// </summary>
+        public bool OverwriteUnchangedOutputDirFiles { get; set; } = true;
```
ddobrev, line 168: Delete the option and always compare.

src/Generator/Driver.cs
```diff
@@ -316,14 +316,40 @@ public void SaveCode(IEnumerable<GeneratorOutput> outputs)
...
+            {
+                var fi = new FileInfo(file);
+
+                // We only compare contents if the generated code length is the same as the length of the existing file. If their lengths are different then they are
+                // definitely no the same so we should overwrite.
+                if (fi.Exists && fi.Length == generatedCode.Length)
+                {
+                    var existingFileContent = File.ReadAllText(file);
+
+                    writeGeneratedCodeToFile = !existingFileContent.Equals(generatedCode);
```
ddobrev, line 343: Eliminate this local variable and simply write if the comparison fails.

src/Generator/Options.cs
IIFE: Done

src/Generator/Driver.cs
IIFE: Done

src/Generator/Driver.cs
```diff
@@ -316,14 +316,35 @@ public void SaveCode(IEnumerable<GeneratorOutput> outputs)
...
             }
         }
 
+        private void WriteGeneratedCodeToFile(string file, string generatedCode)
+        {
+            var fi = new FileInfo(file);
+
+            // We only compare contents if the generated code length is the same as the length of the existing file. If their lengths are different then they are
+            // definitely no the same so we should overwrite.
+            if (fi.Exists && fi.Length == generatedCode.Length)
```
ddobrev, line 333: Would you mind rewriting as a single statement, like this:
```C#
if (!fi.Exists || fi.Length != generatedCode.Length ||
    File.ReadAllText(file) != generatedCode)
    File.WriteAllText(file, generatedCode);
```

IIFE: Done

src/Generator/Driver.cs
```diff
@@ -316,14 +316,25 @@ public void SaveCode(IEnumerable<GeneratorOutput> outputs)
...
                     Diagnostics.Message("Generated '{0}'", fileRelativePath);
                 }
             }
         }
 
+        private void WriteGeneratedCodeToFile(string file, string generatedCode)
+        {
+            var fi = new FileInfo(file);
+
+            // We only compare contents if the generated code length is the same as the length of the existing file. If their lengths are different then they are
```
ddobrev, line 331: Please excuse me as I'm a little distracted today. Allow me to ask you for one final change: delete this comment as your code is self explanatory.

IIFE: No problem. Done

PR #1376, opened 2020-05-24
Add built-in compilation of extra symbols

ddobrev:


src/CppParser/Bindings/premake5.lua
```diff
@@ -1,6 +1,6 @@
 include ("CSharp")
-
-if EnableNativeProjects() and os.ishost("windows") and not os.getenv("CI") then
+-- TODO: fix the C# parser bindings on win
+if EnableNativeProjects() and os.ishost("windows") then
```
josetr, line 4: The C# parser bindings are good as we've discussed before. Remove this.

tests/CSharp/CSharp.Tests.cs
```diff
@@ -731,9 +731,9 @@ public void TestAlignment()
 
         foreach (var internalType in new Type[]
             {
-                typeof(CSharp.TwoTemplateArgs.__Internal_Ptr),
-                typeof(CSharp.TwoTemplateArgs.__Internalc__S_TwoTemplateArgs___I_I),
-                typeof(CSharp.TwoTemplateArgs.__Internalc__S_TwoTemplateArgs___I_f)
+                //typeof(CSharp.TwoTemplateArgs.__Internal_Ptr),
+                //typeof(CSharp.TwoTemplateArgs.__Internalc__S_TwoTemplateArgs___I_I),
+                //typeof(CSharp.TwoTemplateArgs.__Internalc__S_TwoTemplateArgs___I_f)
```
josetr, line 736: Why have you disabled these?

josetr:
* The previous macos build took 2:36 and this one took `6` minutes.
* This will allow us to test those generated symbol files but we wont be testing the exact same behavior that everyone else will be getting and I don't think that's good. Sometimes things will work in CppSharp and give us false confidence that it's working for everyone else.
* Most constructors / destructors that have been removed are `trivially_constructible or trivially_destructible` and as I've said before they don't even need to be generated at all.

josetr (changes requested):

ddobrev:
>     * The previous macos build took 2:36 and this one took `6` minutes.

The builds serve us, not the other way around. Removing any trivial constructors or destructors would hardly make it faster. But we need this feature and if it slows the build down a little, so be it.

>     * This will allow us to test those generated symbol files but we wont be testing the exact same behavior that everyone else will be getting and I don't think that's good. Sometimes things will work in CppSharp and give us false confidence that it's working for everyone else.

I don't understand why the behaviour would be different. Users would have the same code and options we do.

>     * Most constructors / destructors that have been removed are `trivially_constructible or trivially_destructible` and as I've said before they don't even need to be generated at all.

It should be easy to skip these in the symbols, I'll try it.

src/CppParser/Parser.cpp
```diff
@@ -4707,33 +4770,171 @@ ParserTargetInfo* Parser::GetTargetInfo()
...
-        }
+        auto res = new ParserResult();
+        HandleDiagnostics(res);
+        return res;
     }
-    if (MinLineNumberStart < std::numeric_limits<int>::max())
+    return 0;
+}
+
+void Parser::Link(const std::string& File, const LinkerOptions* LinkerOptions)
```
tritao, line 4802: All this code needs at the very least to be moved to a `Link.cpp` file.

src/Generator/Passes/GenerateSymbolsPass.cs
```diff
@@ -53,7 +53,38 @@ private void GenerateSymbols()
                 var e = new SymbolsCodeEventArgs(module);
                 SymbolsCodeGenerated?.Invoke(this, e);
                 if (string.IsNullOrEmpty(e.CustomCompiler))
+                {
+                    if (module == Options.SystemModule ||
```
tritao, line 57: Please factor this code better, right now it introduces a lot of nesting.

josetr:
> It should be easy to skip these in the symbols, I'll try it.

We also need to change the codegen to avoid calling them.

src/CppParser/Parser.cpp
ddobrev: This might be useful now what we use a parallel build.

src/Generator/Passes/GenerateSymbolsPass.cs
ddobrev: I'll see what I can do.

ddobrev: Done.

src/CppParser/Parser.cpp
ddobrev: Done.

tests/CSharp/CSharp.Tests.cs
ddobrev: Fixed.

josetr:
> > ```
> > * The previous macos build took 2:36 and this one took `6` minutes.
> > ```
> 
> The builds serve us, not the other way around. Removing any trivial constructors or destructors would hardly make it faster.

The build taking so long was a hint that something was being done wrong. Things look much better now that you are using better compiler flags but it's weird how you care very little about making builds 2.5 times slower. Without all the optimizations that we've done last week the build would take about `10` minutes and maybe only then you would notice or care about it. We also need to be aware that there are people with much bigger projects than CppSharp and making their builds 2.5 times slower might be an eternity and not just 4 minutes.

> Removing any trivial constructors or destructors would hardly make it faster.

The reason why I want to remove trivial constructors / destructors is because I also want to remove the calls to those constructors / destructors and avoid a `managed to native` transition that are always expensive and should be avoided whenever it's possible. 

> But we need this feature and if it slows the build down a little, so be it.

`Need` is overstated here: the only thing that this give us that couldn't be done by just creating c wrappers is partial support for extending "native abstract classes" in the managed side but as I've mentioned before this is not really an issue because we never had complete support for it and we won't until someone decides to add it and that could take `unknown` time.
* There is no easy way for the user that extends that class to implement it properly since there is no easy way to access the copy/move constructor nor the copy/move assignment operators.
* The user cannot implement a pure function from the native side because for this to happen the vtable needs to be modified and  we are not doing that.

You only noticed about 1 of the many problems that these abstract classes have because there was a test giving false positives. I still think that it would have been better to ignore this test and add a TODO because the support for the whole thing wasn't trivial rather than merging the previous commit that is supposed to fix it but the CI couldn't event test so we had no idea whether the fix was fully correct or not.

> > ```
> > * This will allow us to test those generated symbol files but we wont be testing the exact same behavior that everyone else will be getting. 
> > ```
> 
> I don't understand why the behaviour would be different. Users would have the same code and options we do.
> 

My bad here. I thought that this would only be used internally.

-----

Overall if I had to chose between making builds slower and slowing all of us that are trying to contribute to CppSharp vs adding partial support for something that we have never supported  and none has requested or complained then I'd chose the first one for as long as it's possible.

Here is a suggestion to dodge these build time problems (if they are still a problem): add an option that will allow us to enable or disable these special compiler flags that are required for the symbols of these abstract classes to show up.

This way we can create a specific project that tests those abstract classes and we only would have to use those special compiler flags that slow down the build for a single project and not all of them, thus slowing the build time as little as possible.

PR #1423, opened 2020-10-14
Added the ability to remove a built-in pass.

fsinisi90:


src/Generator/Passes/PassBuilder.cs
```diff
@@ -32,6 +32,14 @@ public void AddPass(T pass)
             Passes.Add(pass);
         }
 
+        /// <summary>
+        /// Remove a previously-added pass, if exists.
+        /// </summary>
+        public bool RemovePass(T pass)
+        {
+            return pass is TranslationUnitPass && Passes.Remove(pass);
```
tritao, line 40: Any specific reason you are checking for `pass is TranslationUnitPass` here?

Since this method is supposed to be generic, I think keeping the code generic would be best here.

PR #1425, opened 2020-10-16
Fix anonymous member name colliding with enclosing type name

### LINKED ISSUE #1414: Inheriting from a base class that uses unions and lives in another file generates uncompilable code
I've added a comment in the generated code which shows the exact line where the error `CS0542` happens. This only happens if the base class lives in another file. If both classes live in the same file then a completely different code is generated which doesn't have this issue.

file1.h
```.cs
struct StructBase {
    union {
        float arr[2];
        struct {
            float a, b;
        };
    };
};
```

file2.h
```
#include "file.h"
struct Struct : public StructBase { };
```

OS: Windows

##### Used settings

Target: MSVC

```.cs
// DEBUG: struct StructBase
// DEBUG: {
// DEBUG:     union
// DEBUG:     {
// DEBUG:         float arr[2];
// DEBUG:         struct
// DEBUG:         {
// DEBUG:             float a, b;
// DEBUG:         };
// DEBUG:     };
// DEBUG: }
public unsafe partial class StructBase : IDisposable
{
    [StructLayout(LayoutKind.Explicit, Size = 8)]
    public partial struct __Internal
    {
        [FieldOffset(0)]
        internal fixed float arr[2];

        [FieldOffset(0)]
        internal global::TestLib.StructBase._0._0.__Internal _0;

        [DllImport("TestLib", CallingConvention = global::System.Runtime.InteropServices.CallingConvention.Cdecl,
            EntryPoint="??0StructBase@@QEAA@AEBU0@@Z")]
        internal static extern global::System.IntPtr cctor(global::System.IntPtr __instance, global::System.IntPtr _0);
    }

    // DEBUG: union
    // DEBUG:     {
    // DEBUG:         float arr[2];
    // DEBUG:         struct
    // DEBUG:         {
    // DEBUG:             float a, b;
    // DEBUG:         };
    // DEBUG:     }
    public unsafe partial struct _0
    {
        [StructLayout(LayoutKind.Explicit, Size = 8)]
        public partial struct __Internal
        {
            [FieldOffset(0)]
            internal fixed float arr[2];

            [FieldOffset(0)]
            internal global::TestLib.StructBase._0._0.__Internal _0;
        }

        // DEBUG: struct
        // DEBUG:         {
        // DEBUG:             float a, b;
        // DEBUG:         }
        public unsafe partial class _0 // CS0542: member names cannot be the same as their enclosing type
        {
            [StructLayout(LayoutKind.Explicit, Size = 8)]
            public partial struct __Internal
            {
                [FieldOffset(0)]
                internal float a;

                [FieldOffset(4)]
                internal float b;
            }
        }
    }

```

josetr:
Fixes #1414

### DIFF
--- a/src/Generator/Passes/CleanInvalidDeclNamesPass.cs
+++ b/src/Generator/Passes/CleanInvalidDeclNamesPass.cs
@@ -146,7 +146,14 @@ namespace CppSharp.Passes
         {
             var anonymousDecls = decls.Where(p => string.IsNullOrEmpty(p.Name)).ToList();
             for (int i = 0; i < anonymousDecls.Count; i++)
-                anonymousDecls[i].Name = $"_{i}";
+            {
+                var anonymousDecl = anonymousDecls[i];
+
+                if (anonymousDecl.Namespace != null && anonymousDecl.Namespace.Name == anonymousDecl.Name)
+                    anonymousDecl.Name = $"__{i}";
+                else
+                    anonymousDecl.Name = $"_{i}";
+            }
         }
     }
 }
--- a/tests/CSharp/CSharp.Tests.cs
+++ b/tests/CSharp/CSharp.Tests.cs
@@ -1429,4 +1429,10 @@ public unsafe class CSharpTests : GeneratorTestFixture
     {
         Assert.That(CSharp.CSharp.TakeTypemapTypedefParam(false), Is.False);
     }
+
+    [Test]
+    public void TestAnonymousMemberNameCollision()
+    {
+        StringAssert.EndsWith(nameof(CSharp.TestAnonymousMemberNameCollision._0.__0), "__0");
+    }
 }
--- a/tests/CSharp/CSharp.h
+++ b/tests/CSharp/CSharp.h
@@ -6,6 +6,7 @@
 #include <limits>
 #include <string>
 #include "AnotherUnit.h"
+#include "ExcludedUnit.hpp"
 #include "CSharpTemplates.h"
 
 class DLL_API Foo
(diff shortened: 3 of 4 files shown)


src/Generator/Passes/CleanInvalidDeclNamesPass.cs
```diff
@@ -146,7 +146,14 @@ private static void Check(IEnumerable<Declaration> decls)
         {
             var anonymousDecls = decls.Where(p => string.IsNullOrEmpty(p.Name)).ToList();
             for (int i = 0; i < anonymousDecls.Count; i++)
-                anonymousDecls[i].Name = $"_{i}";
+            {
+                var anonymousDecl = anonymousDecls[i];
+
+                if (anonymousDecl.Namespace != null && anonymousDecl.Namespace.Name == anonymousDecl.Name)
```
ddobrev, line 152: You could simplify to:
`anonymousDecl.Namespace?.Name == anonymousDecl.Name`

ddobrev:
@josetr I'd like to start by saying I'm extremely happy to finally talk to you. I had been too busy for the last week and a half but now I'm happy with the opportunity to personally thank you for the extensive testing of our product.
Your fix is quite simple and therefore good by itself. I've added a minor suggestion. However, it lacks a just as simple unit test. So let me ask you to simply move the test case from your bug report in one of our test projects and this one will be ready to go.

src/Generator/Passes/CleanInvalidDeclNamesPass.cs
josetr: I avoided that because I thought that if `anonymousDecl.Name` is null and `anonymousDecl.NameSpace` is null then it would be detected as a collision even tho there isn't one.

josetr:
No problem. I was just a bit surprised to find many bugs but I've realized that is mainly because I'm using the opposite `ParseOptions` that you guys use to test CppSharp.

Issue #1421, opened 2020-10-12
Constant variables are always resolved at runtime.


josetr:
```.cpp
static constexpr int degrees = 180;
const int degrees = 180;
```

Both are translated into `*(int*)CppSharp.SymbolResolver.ResolveSymbol("Lib.dll", "degrees")` which is not efficient and also forces you to export the variable to make it work.

The current workaround is using `#define degrees 180`.

Issue #1419, opened 2020-10-09
Generated std::string destructor is wrong when target is i686-pc-win32-msvc


josetr:
The parameter `deleted` should be removed.

OS: Windows

##### Used settings

`Driver.ParserOptions.TargetTriple = "i686-pc-win32-msvc"`

Target: MSVC

##### Stack trace or incompilable generated code

```.cs
[SuppressUnmanagedCodeSecurity]
[DllImport("Std-symbols", CallingConvention = global::System.Runtime.InteropServices.CallingConvention.ThisCall,
    EntryPoint="??1?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@QAE@XZ")]
internal static extern void dtorc__N_std_S_basic_string__C___N_std_S_char_traits__C___N_std_S_allocator__C
    (global::System.IntPtr __instance, int deleted); 

global::Std.BasicString.__Internalc__N_std_S_basic_string__C___N_std_S_char_traits__C___N_std_S_allocator__C.dtorc__N_std_S_basic_string__C___N_std_S_char_traits__C___N_std_S_allocator__C
       (__Instance, 0);
```

PR #1428, opened 2020-10-16
Fix wrong marshalling for static member arrays

### LINKED ISSUE #1417: static T array[] generates (uncompilable | incorrect) and slow code
##### Brief Description

Using `char [] array` generates uncompilable code but using a type which uses more storage like short generates compilable but incorrect code.

OS: Windows

##### Used settings

Target: MSVC

```.cs
// DEBUG: static char static_array[100]
public static sbyte[] StaticArray
{
    get
    {
        var __ptr = (byte*)CppSharp.SymbolResolver.ResolveSymbol("TestLib", "?static_array@Class@@2PADA");
        sbyte[] __value = null;
        if (__ptr != null)
        {
            __value = new sbyte[100];
            for (int i = 0; i < 100; i++)
                __value[i] = __ptr[i]; // CS0266: Cannot implicitly convert type 'byte' to 'sbyte'. An explicit conversion 
        }
        return __value;
    }
}
```

josetr:
Fixes #1417

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpSources.cs
+++ b/src/Generator/Generators/CSharp/CSharpSources.cs
@@ -880,12 +880,12 @@ namespace CppSharp.Generators.CSharp
 
             var arrayType = var.Type as ArrayType;
             string ptr = Generator.GeneratedIdentifier("ptr");
-            if (arrayType != null && var.Namespace is Class @class && @class.IsRefType)
+            if (arrayType != null)
             {
-                WriteLine($@"var {ptr} = {
-                   (arrayType.Type.IsPrimitiveType(PrimitiveType.Char) &&
-                    arrayType.QualifiedType.Qualifiers.IsConst ?
-                       string.Empty : "(byte*)")}{location};");
+                if (arrayType.Type.IsPrimitiveType(PrimitiveType.Char) && arrayType.SizeType != ArrayType.ArraySize.Constant)
+                    WriteLine($@"var {ptr} = {location};");
+                else
+                    WriteLine($@"var {ptr} = ({arrayType.Type.Visit(TypePrinter)}*){location};");
             }
             else
             {
--- a/tests/CSharp/CSharp.Tests.cs
+++ b/tests/CSharp/CSharp.Tests.cs
@@ -735,6 +735,23 @@ public unsafe class CSharpTests : GeneratorTestFixture
     }
 
     [Test]
+    public void TestStaticVariables()
+    {
+        Assert.That(StaticVariables.ConstexprBoolean, Is.EqualTo(true));
+        Assert.That(StaticVariables.ConstexprChr, Is.EqualTo('G'));
+        Assert.That(StaticVariables.ConstexprUChr, Is.EqualTo('G'));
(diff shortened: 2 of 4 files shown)


tritao:
The build is failing on the bot:

```
C:\projects\cppsharp\build\vs2019\gen\Common\Common_GenerateName.cs(1413,21): error CS0019: Operator '==' cannot be applied to operands of type 'sbyte*' and 'IntPtr' [C:\projects\cppsharp\tests\Common\Common.CSharp.csproj]
```

I'm assuming its something caused by these changes, can you take a look?

josetr:
Now it's failing also due to #1419 as well.

ddobrev:
Thank you, @josetr .
`C:\projects\cppsharp\build\vs2019\gen\Common\Common_GenerateName.cs(1413,21): error CS0019: Operator '==' cannot be applied to operands of type 'sbyte*' and 'IntPtr' [C:\projects\cppsharp\tests\Common\Common.CSharp.csproj]` is no longer a problem but now `C:\projects\cppsharp\build\vs2019\gen\CSharp\CSharp.cs(33471,68): error CS0021: Cannot apply indexing with [] to an expression of type 'IntPtr' [C:\projects\cppsharp\tests\CSharp\CSharp.CSharp.csproj]`'s appeared.

ddobrev:
@josetr much better, Windows passed now. Unfortunately, Unix doesn't. The reason is that default initialization of members is often inlined, and we don't automatically support inlined code yet (I've been working on it but it's difficult and not done yet). Therefore for this test to work all initialization must move to C++ as you've already done with 2 of your constants.

josetr:
Yes they issue is related to https://github.com/mono/CppSharp/issues/1421 which I've reported as well. What I've done in my local repo is expose an `internal` `Initializer` in the `Variable` class and store the expression from `clang::VarDecl::getAnyInitializer()`. It's storing an `ExpressionObsolete` and I'm using bits like `CheckForSimpleExpressions` from the `HandleDefaultParamValuesPass` class therefore it has similar issues like not supporting every possible valid C++ expression but the most common usage for static const variables is storing integral types and that is supported.

ddobrev:
You're correct that having the feature of "constants as constants" would solve this issue too. But while we don't have it yet, tests should keep working with whatever approach we use at present. Therefore there's little option but to move the initializations to `cpp`.
The pieces of code you mention are definitely going to be useful when we go about this feature. So let's merge this PR after the tests pass and we can discuss proceeding.

josetr:
I agree and I was about to do that. I just wanted to share what I've done in case that is good to turn into a pull request.

ddobrev:
It does sound good. Even if there's anything to improve, we'll catch it in the review.

tritao:
> Yes they issue is related to #1421 which I've reported as well. What I've done in my local repo is expose an `internal` `Initializer` in the `Variable` class and store the expression from `clang::VarDecl::getAnyInitializer()`. It's storing an `ExpressionObsolete` and I'm using bits like `CheckForSimpleExpressions` from the `HandleDefaultParamValuesPass` class therefore it has similar issues like not supporting every possible valid C++ expression but the most common usage for static const variables is storing integral types and that is supported.

That sounds reasonable, feel free to send a PR.

Issue #1420, opened 2020-10-09
Wrong code generated for copying a std::string 


josetr:
##### Brief Description

The class `std::string` is not trivially copyable therefore it shouldn't be copied by just assigning the bits from the source.  Right now a crash will happen at `__basicString0.Dispose();` because it will try to free the same internal buffer which is no longer valid.

OS: Windows

##### Used headers

string

##### Used settings

Target: MSVC

##### Stack trace or incompilable generated code
```.cs
public static void Func(string s)
{
    var __basicString0 = new global::Std.BasicString<sbyte, global::Std.CharTraits<sbyte>, global::Std.Allocator<sbyte>>();
    global::Std.BasicStringExtensions.Assign(__basicString0, s);
    var __arg0 = *(global::Std.BasicString.__Internalc__N_std_S_basic_string__C___N_std_S_char_traits__C___N_std_S_allocator__C*) __basicString0.__Instance;
    __Internal.Func(__arg0);
    __basicString0.Dispose();
}
```

josetr:
Fixed by #1427.

PR #1429, opened 2020-10-17
Fix wrong generated destructor

### LINKED ISSUE #1419: Generated std::string destructor is wrong when target is i686-pc-win32-msvc

josetr:
Fixes #1419

### DIFF
--- a/src/Generator/Passes/CheckAbiParameters.cs
+++ b/src/Generator/Passes/CheckAbiParameters.cs
@@ -56,7 +56,7 @@ namespace CppSharp.Passes
 
             // Deleting destructors (default in v-table) accept an i32 bitfield as a
             // second parameter in MS ABI.
-            if (method != null && method.IsDestructor && Context.ParserOptions.IsMicrosoftAbi)
+            if (method != null && method.IsDestructor && method.IsVirtual && Context.ParserOptions.IsMicrosoftAbi)
             {
                 method.Parameters.Add(new Parameter
                 {


ddobrev:
@josetr while I trust this fix works are you sure it is correct? These hidden parameters are strictly defined by the ABI. In fact, we've had for quite some time the idea of reading these from Clang instead of writing them ourselves based on ms's scattered documentation on the subject. This would need to happen somewhere around https://github.com/mono/CppSharp/blob/master/src/CppParser/Parser.cpp#L3301 as the info about code generation contains these ABI parameters if any.

josetr:
@ddobrev No, I can't be sure due to lack of documentation like you said. The target i686-pc-win32-msvc certainly needs it since my other 2 pull requests (and my library that is using CppSharp) fail every time without it.

Maybe the Microsoft 64bit ABI doesn't need the `IsVirtual` check, but I highly doubt it. The memory corruption for the 32 bit target was silent and just doesn't show until your library gets complex enough to trigger it. The same is probably happening for the 64 bit target.

ddobrev:
@tritao what do you think?

tritao:
Fix looks correct to me, deleted destructors should always be virtual.

ddobrev:
@josetr it's possible that you've fixed part of the problem with this PR but [not all of it](https://ci.appveyor.com/project/tritao/cppsharp/builds/35874369/job/1agpg1krqegspps3#L1017). I cannot tell if it's related to this kind of hidden parameters but the only reliable way to get them correctly is reading them from Clang as I mentioned [earlier](https://github.com/mono/CppSharp/pull/1429#issuecomment-710779337). If you feel you could give it a try, I'm definitely available to help.

josetr:
@ddobrev I already saw this same test breaking CI build failing before so I've managed to reproduce the issue and it seems the code generated to copy a string is still wrong so the problem may be related to #1420 instead.

PR #1430, opened 2020-10-17
Add missing `Variable.Type` null check.

### LINKED ISSUE #1412: Valid c++ code using variadic templates is not supported
##### Brief Description
```.cpp
template<typename... Args>
int make(Args... args)
{
    return {};
}

template<int... args>
struct Struct
{
    static const auto s = make(args...);
};
```

OS: Windows
##### Used settings

Target: MSVC

`Driver.ParserOptions.LanguageVersion = LanguageVersion.CPP17`

```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at CppSharp.AST.AstVisitor.VisitVariableDecl(Variable variable)
   at CppSharp.AST.Variable.Visit[T](IDeclVisitor`1 visitor)
   at CppSharp.AST.AstVisitor.VisitDeclarationContext(DeclarationContext context)
   at CppSharp.AST.AstVisitor.VisitClassDecl(Class class)
   at CppSharp.Passes.ResolveIncompleteDeclsPass.VisitClassDecl(Class class)
   at CppSharp.AST.Class.Visit[T](IDeclVisitor`1 visitor)
   at CppSharp.AST.AstVisitor.VisitDeclarationContext(DeclarationContext context)
   at CppSharp.Passes.TranslationUnitPass.VisitTranslationUnit(TranslationUnit unit)
   at CppSharp.Passes.TranslationUnitPass.VisitASTContext(ASTContext context)
   at CppSharp.Passes.ResolveIncompleteDeclsPass.VisitASTContext(ASTContext c)
   at CppSharp.Generators.BindingContext.<RunPasses>b__29_0(TranslationUnitPass pass)
   at CppSharp.PassBuilder`1.RunPasses(Action`1 action)
   at CppSharp.Generators.BindingContext.RunPasses()
   at CppSharp.Driver.ProcessCode()
   at CppSharp.ConsoleDriver.Run(ILibrary library)
   at Program.Main(String[] args) in C:\Test\Program.cs:line 25```

### PR DESCRIPTION
**Pull request by josetr**
Fixes #1412.

### DIFF
--- a/src/AST/ASTVisitor.cs
+++ b/src/AST/ASTVisitor.cs
@@ -470,6 +470,8 @@ namespace CppSharp.AST
             if (!VisitDeclaration(variable))
                 return false;
 
+            // TODO: Remove this null check once CppParser can properly handle auto types.
+            // This is workaround for https://github.com/mono/CppSharp/issues/1412.            
             if (variable.Type == null)
                 return false;
 
--- a/src/AST/ASTVisitor.cs
+++ b/src/AST/ASTVisitor.cs
@@ -470,6 +470,9 @@ namespace CppSharp.AST
             if (!VisitDeclaration(variable))
                 return false;
 
+            if (variable.Type == null)
+                return false;
+
             return variable.Type.Visit(this, variable.QualifiedType.Qualifiers);
         }
 
--- a/tests/CSharp/CSharp.h
+++ b/tests/CSharp/CSharp.h
@@ -1430,3 +1430,15 @@ namespace CXXRecordDeclWithoutDefinition
     template <> struct it<list<>> { };
     template <> struct it<list<> const> { };
 }
+
+template<int... n>
+struct TestVariableWithoutType
+{
+    template<typename... Args>
+    static constexpr int create(Args... args)
+    {
+        return {};
+    }
+
+    static constexpr auto variable = create(n...);
+};

### CODE REVIEW

**Comment by ddobrev** [comment]
The fix is simple and so is the test but I'm worried. Such variables most probably have types but we just as probably skip walking these types of types (that's right!). So could you debug the case of such a variable and see if this is the case?
The code's at https://github.com/mono/CppSharp/blob/master/src/CppParser/Parser.cpp#L3388. You could help yourself with:
```c++
if (Var->name == "variable") // the name from your test
    __debugbreak();
```
You'd just have to enable native debugging in the project settings of the respective executable (presumably CSharp.Gen.exe since your test is in CSharp).

josetr:
I assumed it was ok for an `auto` variable that depends on templates to be null. I just checked and the nullptr is returned from                         https://github.com/mono/CppSharp/blob/3ffa42aa2fd4ea9ff432660f5eb579701c609f17/src/CppParser/Parser.cpp#L2809

ddobrev:
@josetr thank you for tracking it down. @tritao do you think there's anything we can do here?

tritao:
We usually shy away from null checks for types in the AST layer since it usually just hides bugs.

In this specific case, we probably need to implement the type parsing in the parser layer.
That said, I don't think that would be a good use of time for anyone right now, since its not blocking anything.

So I am fine with merging this workaround, I would just request that you add a comment above the null check explaining the problem and/or linking to the issue.

ddobrev:
@josetr the variable should definitely have some type but I cannot tell how long it would take to find it. It might as well be a bug in our version of Clang. So please just add 2 TODO comments, at your fix in C# and at the line you found in C++, and I'll merge.

PR #1431, opened 2020-10-17
Fix generated constructor cant use `placement new` when class overloads `operator new`

### LINKED ISSUE #1418: Generated code for class constructor in Symbols.cpp is incorrect when a class overloads operator new
##### Brief Description
```.cpp
struct Struct
{
    void* operator new ( size_t n );
};
```

OS: Windows

##### Used settings

Target: MSVC

```.cpp
extern "C" { void Struct_Struct(void* __instance) { new (__instance) Struct(); } } // operator new cannot be called with the given argument
extern "C" { void Struct_Struct___1__S_Struct(void* __instance, const Struct& _0) { new (__instance) Struct(_0); } } // operator new cannot be called with the given argument
```

josetr:
Fixes #1418.

### DIFF
--- a/src/Generator/Passes/SymbolsCodeGenerator.cs
+++ b/src/Generator/Passes/SymbolsCodeGenerator.cs
@@ -137,7 +137,7 @@ namespace CppSharp.Passes
                     p => cppTypePrinter.VisitParameter(p)))})");
                 WriteLine($": {@namespace}({@params}) {{}} }};");
                 Write($"extern \"C\" {{ void {wrapper}({signature}) ");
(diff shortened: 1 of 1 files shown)


ddobrev:
This looks a useful fix, thank you. Just add the usual test and I'll merge.

josetr:
@ddobrev I should have left a comment that I didn't know where to add the test. Maybe there should be a project that actually uses the code spit out by `SymbolsCodeGenerator`. I don't use premake so I think I'll leave that task to someone else.

ddobrev:
You're right, it's in the generated C++ only. It would've been easy to test if I was done but alas.

PR #1432, opened 2020-10-17
Use C# 'default literal' to avoid using a type's name that may be renamed

### LINKED ISSUE #1410: Code generated for default argument is invalid when class name starts with underscore
OS: Windows

Target: MSVC

`struct _Class { };`

##### Incompilable generated code

```.cs

// DEBUG: void Method(_Class* arg = nullptr)
public static void Method(global::TestLib.Class arg = default(global::TestLib._Class))
{
    var __arg0 = ReferenceEquals(arg, null) ? global::System.IntPtr.Zero : arg.__Instance;
    __Internal.Method(__arg0);
}
```

josetr:
Fixes #1410 but not completely since using a nested expression that uses a class with leading underscore will still break but those 2 cases combined should be extremely rare. I'm not using C#'s `default literal` in nested expressions because there is potential for ambiguity.

According to the source code comment the order of `HandleDefaultParamValuePass` can't be changed. If there is a better way let me know but regardless it can always be implemented at a later time.

### DIFF
--- a/build/Helpers.lua
+++ b/build/Helpers.lua
@@ -164,7 +164,7 @@ end
 
 function SetupManagedProject()
   language "C#"
-  csversion "7.2"
+  csversion ("7.2")
   location ("%{wks.location}/projects")
   buildoptions {"/platform:".._OPTIONS["arch"]}
 
--- a/build/Helpers.lua
+++ b/build/Helpers.lua
@@ -164,8 +164,8 @@ end
 
 function SetupManagedProject()
   language "C#"
-  csversion ("7.2")
   location ("%{wks.location}/projects")
+  buildoptions {"/langversion:7.2"}
   buildoptions {"/platform:".._OPTIONS["arch"]}
 
   dotnetframework "4.7.2"
--- a/build/Helpers.lua
+++ b/build/Helpers.lua
@@ -164,6 +164,7 @@ end
 
 function SetupManagedProject()
   language "C#"
+  csversion "7.2"
   location ("%{wks.location}/projects")
   buildoptions {"/platform:".._OPTIONS["arch"]}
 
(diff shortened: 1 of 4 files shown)


ddobrev:
Looks OK but it means we'd have to drop support for C# below 7.1. I personally think it's a little too early for this. @tritao what do you think?
In general this, `HandleDefaultParamValuePass`, is entirely mine and pure garbage. I wrote it years ago to add quick support for default values. The reason was and is that the proper support requires printing expressions which is much work at least in quantity. Assuming you don't have the opportunity to do it, we could think of some other way to hack around this while keeping the generated code compatible. For example, this pass could run last, after renaming.

tritao:
Using newer C# sounds totally fine to me.

ddobrev:
@josetr agreed. Please just actually use your fix :), because the build's broken. Furthermore: would it actually be possible to always use the literal? If so, the code can become much simpler — no parameter.

ddobrev:
@josetr this is already good to merge. However, I want to ensure: have you been able to try always using the default literal? If it doesn't work, why?

josetr:
@ddobrev 
```.cs
// error CS0121: The call is ambiguous between the following methods or properties: 'Foo.Foo(string)' and 'Foo.Foo(char)'

// DEBUG: virtual void abstract(const Foo& foo = Foo()) = 0
public override void Abstract()
{
    Abstract(new global::CSharp.Foo(default));
}
```

ddobrev:
@josetr fair enough. However, you need to update the version of C# we use as it's too old — the Linux CI's broken. So please add:
`csversion "7.2"`
below https://github.com/mono/CppSharp/blob/0a158717519e0552930a9270521f377a25c4c534/build/Helpers.lua#L166.

ddobrev:
Thank you. I'm going to wait for the build just in case and merge.

Issue #674, opened 2016-07-10
Map primitive C++ constants to C# constants


ddobrev:
We use dynamic calls in C# to resolve symbols for C++ constants. This has the following three problems:
1. It's much slower than reading a hard-coded value;
2. It prevents constants from being used in default value of parameters, attributes and anywhere else a compile-time value is required;
3. It's needlessly complex because constants rarely if ever change their values across versions and because of the tight integration C++# creates between the binding and the native library it's unsafe anyway to use a binding generated for one version with a newer one.

We can actually support (in most cases) complex constants (objects) too. But let's just start with primitive constants: integers, Booleans, possibly strings.

rokups:
I am looking at this issue. Do you have any advice how i could extract constant values from variable declarations? I can see `DebugText` has this info but it probably is a very wrong and fragile approach to use that.

rokups:
My attempt: https://github.com/rokups/CppSharp/commit/2f0a417a6aba75a030cd34c6ef72cd613cc02c71
However `VisitStatement(decl.Expression);` call throws `NullPointerException` exception on [statement.Class](https://github.com/rokups/CppSharp/blob/2f0a417a6aba75a030cd34c6ef72cd613cc02c71/src/Parser/ASTConverter.cs#L1126) access. I basically copied exact same steps that were used by `Parameter` for initializing it's `DefaultArgument`. Any clue what is going on here?

ddobrev:
@rokups you have added a new member in C++ without regenerating the parser bindings so what happens is normal. In addition, please initialise the new field to 0 in the constructor rather than when using it. Finally, please do not stack your commits on top - produce a single commit by amending.

rokups:
Oh but i did regenerate parser bindings. If i would not have done it then CppSharp would not compile because `VisitStatement(decl.Expression);` would be accessing something that does not exist in parser bindings.

ddobrev:
@rokups for some reason the large diffs are not even noted in that link. Anyway, then try the second step, initialisation to 0 in the constructor instead.

rokups:
Alright i have some kind of success. Please see [my branch](https://github.com/rokups/CppSharp/commits/var-decl-expr).

I am pretty sure there is million things i did wrong though. For starters i added constants pass that forces constant variables to be included in generation by excluding them from symbols check. Is that a valid approach?

[...]

PR #1435, opened 2020-10-19
Basic support for static constexpr variables

### LINKED ISSUE #1421: Constant variables are always resolved at runtime.

josetr:
Fixes #1421


josetr:
Not sure why InternalsVisibleTo in  linux / mono doesn't work. I'll check it tomorrow.

tritao:
This looks nice.

The CI issue seems related to the parser bindings, you only updated the generated C++/CLI bindings, which are only used on Windows, on other platforms generated C# code is used.

They need to be re-generated for all platforms by running `https://github.com/mono/CppSharp/blob/master/src/CppParser/ParserGen/ParserGen.cs`. You can generate them all from Windows, but you need a folder with the header files for all platforms.

@ddobrev Can you upload those to our releases page on Github?

ddobrev:
@tritao it was loooong due...
@josetr please get the headers [here](https://github.com/mono/CppSharp/releases/download/CppSharp/headers.zip) and unpack them in `build` (so that you have `build/headers/...`), then rerun the generator for the parser.

src/AST/Variable.cs
```diff
@@ -10,6 +10,7 @@ public override T Visit<T>(IDeclVisitor<T> visitor)
 
         public Type Type { get { return QualifiedType.Type; } }
         public QualifiedType QualifiedType { get; set; }
+        internal ExpressionObsolete Initializer { get; set; }
```
ddobrev, line 13: Why is this internal? Just make it public and save your self all of the trouble.

josetr: Probably because the type is named Obsolete. I can just make it public but I still need to generate the bindings anyways.

josetr:
I just generated the bindings and I've realized that they were already outdated with #1429

src/AST/Variable.cs
ddobrev: Never mind that, it's only because we've been building a new automated tree of expressions which should over time replace these.

src/CppParser/APValuePrinter.h
```diff
@@ -0,0 +1,91 @@
...
+#include <llvm/ADT/ArrayRef.h>
+#include <clang/AST/APValue.h>
+#include <clang/AST/Type.h>
+#include <clang/AST/ASTContext.h>
+
+namespace CppSharp { namespace CppParser {
+
+struct APValuePrinter
+{
+	clang::ASTContext& ctx;
```
ddobrev, line 14: You've used tabs here instead of 4 spaces.

ddobrev:
Thank you, @josetr . An outstanding feature.
We had actually [noted it years ago](https://github.com/mono/CppSharp/issues/674) but I think you extended it even further. At the bottom, though, I can see a note that could've simplified your work. I apologize for not linking it earlier. Still, you could check it if you like.
To my surprise, there's a regression in your code [which breaks the build](https://ci.appveyor.com/project/tritao/cppsharp/builds/35838764/job/2qvk0n3eli7c7pvl#L127). It wasn't there in your previous commit. I apologize for merging in a hurry instead of waiting to make sure but could you please send a new PR to fix this?

josetr:
@ddobrev Anytime 👍 . To be honest I was looking into using CppSharp in a personal project that requires high performance and some of the pull requests like this one were really needed for me.

I actually realized my mistake very fast but you were too quick at merging it 😄

ddobrev:
It's all right now. :) Let me ask you: do you intend to send further improvements? We're considering a release to update our NuGet which has stalled for a few months now. However, we'd rather wait to include anything else you might have.

josetr:
@ddobrev I have made pull requests for nearly everything I had in my local repo except 3 that are probably a bit incomplete like supporting string_view, copying arrays using `System.Buffer.MemoryCopy` rather than using a loop and fixing `const wchar_t*` marshalling.

ddobrev:
OK, we won't release today. If there's nothing new from you in a day or two, we'll release. If you send sth later, we'll have it in the subsequent release.

PR #1439, opened 2020-10-21
Fix marshalling for bool arrays

josetr:


tests/CSharp/CSharp.cpp
```diff
@@ -1413,6 +1413,7 @@ const std::string StaticVariables::String = "Str";
 const char StaticVariables::ChrArray[2] { 'A', 'B' };
 const int StaticVariables::IntArray[2] { 1020304050, 1526374850 };
 const float StaticVariables::FloatArray[2] { 0.5020f, 0.6020f };
+const bool StaticVariables::BoolArray[2] { false, true };
```
ddobrev, line 1416: Now that you've added mapping to real constants, does it work to initialize these directly in the header?

tn07: It does. Perhaps tests should be generated as well otherwise it's very time consuming to cover every single case.

josetr: I've added another test to cover that case. In the future we should try to inline these kind of constants as well but I wonder how expensive that is and of course the library would need to be referenced, not just the headers. To be honest I'm still unsure currently in which situations I would need to add a library to Module.Libraries

ddobrev: Mostly when working with dependencies. Anyway, I understand your point and I agree it's nice to have both approaches tested.

PR #1440, opened 2020-10-21
Copy arrays of primitive values using System.Buffers.MemoryCopy

josetr:


src/AST/TypeExtensions.cs
```diff
@@ -435,5 +435,10 @@ public static Module GetModule(this Type type)
 
             return declaration.TranslationUnit.Module;
         }
+
+        public static long GetSizeInBytes(this ArrayType array)
+        {
+            return array.Size * (array.ElementSize / 8);
```
ddobrev, line 441: I think this `8` should be `TargetInfo.PointerWidth`.

josetr: This method is just converting from bits to bytes and the magic number to do that is 8

josetr: @ddobrev should I create a pull request to ignore the test that's breaking the CI build for now?

ddobrev: I'm just afraid if you do, we'll simply never fix it... Do you perhaps have an idea how to?

josetr: The issue is #1441 and the the main problem is that the destructor is called by both the callee and the caller. I'm guessing that when `Indirect` is true then the caller is supposed to do the cleanup and call the object destructor. Then there's the other issue is that the address for the storage of the object is not stable. If you do `Console.WriteLine(new IntPtr(&__arg0))` and then `printf("%p", this)` at the std::string destructor you will get different values and I don't know if a custom marshaller will let you fix that. Maybe the answer is generating wrappers so that we can always call those methods via pointers.

PR #1444, opened 2020-10-22
Add support for const char32_t* and fix const wchar_t* in linux / osx

josetr:


src/Generator/Types/Std/Stdlib.cs
```diff
@@ -198,6 +194,8 @@ public uint GetCharPtrWidth(PointerType pointer)
                     return Context.TargetInfo.WCharWidth;
                 if (pointee.IsPrimitiveType(PrimitiveType.Char16))
                     return Context.TargetInfo.Char16Width;
+                if (pointee.IsPrimitiveType(PrimitiveType.Char32))
```
ddobrev, line 197: We have an overload which takes an `out` parameter. Could you please use this overload and then switch on the result?

tests/CSharp/CSharp.Tests.cs
```diff
@@ -1485,7 +1486,14 @@ public void TestAnonymousMemberNameCollision()
     [Test]
     public void TestStringMarshall()
     {
-        Assert.IsTrue(StringMarshall.CSharpString8(StringMarshall.CSharpString));
-        Assert.IsTrue(StringMarshall.CSharpString16(StringMarshall.CSharpString));
+        const string cs = "C#";
```
ddobrev, line 1489: Could you please test with a longer string which also contains non-ASCII characters? You can use https://github.com/mono/CppSharp/blob/master/tests/Common/Common.Tests.cs#L974 as a reference.

src/Generator/Types/Std/Stdlib.cs
```diff
@@ -266,50 +264,45 @@ public override void CSharpMarshalToManaged(CSharpMarshalContext ctx)
...
+            textGenerator.WriteLine($"int {length} = 0;");
+            textGenerator.WriteLine($"while (*({retPtr}++) != 0) {length} += sizeof({type});");
 
             ctx.Return.Write($@"global::System.Text.Encoding.{
-                encoding}.GetString((byte*) {returnVarName}, __length)");
+                encoding}.GetString((byte*) {returnVarName}, {length})");
         }
 
-        private string GetEncodingClass(Parameter parameter)
+        private (Encoding Encoding, string Name) GetEncoding()
```
ddobrev, line 283: Why have you complicated this method? You can as well use `Context.Options.Encoding` at the caller's site.

ddobrev: By the way, even though I've written it, it can be much better. You can delete the entire method and then simply use:
```C#
string encoding = Context.Options.Encoding.GetType().FullName;
```
and then drop the manually written name-space:
```C#
ctx.Return.Write($@"global::{encoding}.GetString((byte*) {returnVarName}, __length)");
```

tests/CSharp/CSharp.h
```diff
@@ -1487,7 +1487,8 @@ struct TestVariableWithoutType
 class DLL_API StringMarshall
 {
 public:
-    static constexpr const char* CSharpString = "C#";
-    static bool CSharpString8(const char* in);
-    static bool CSharpString16(const char16_t* in);
+    static const char* TestCSharpString(const char* in, CS_OUT const char** out);
```
ddobrev, line 1490: I've been wondering why your tests are so complicated. You could write them in the way I have: [you pass a parameter and then return the parameter](https://github.com/mono/CppSharp/blob/master/tests/Common/Common.h#L1527). This way you easily and fully test both directions.
Also, you don't need a class for this particular test, free functions are enough.

ddobrev (changes requested):

src/Generator/Types/Std/Stdlib.cs
josetr: First I would like to say that I've tried to keep nearly the same logic as before even tho I don't completely understand the logic behind it.

 Using Encoding = Unicode with a 8 bit array will end up using `MarshalAs(UnmanagedType.LPWStr)` which will end up storing a unicode string in a 8 bit char array and I don't know if that's what people would expect to happen since we could also store it as UTF8. Maybe it should be allowed to specify more than one encoding per char type or width.

> Why have you complicated this method? You can as well use `Context.Options.Encoding` at the caller's site.

> By the way, even though I've written it, it can be much better. You can delete the entire method and then simply use:
> 
> ```cs
> string encoding = Context.Options.Encoding.GetType().FullName;
> ```

The method is checking for a supported encoding and it's also looking for the name of the static variable that holds an instance to a specific Encoding, not just the type.

ddobrev: I understand, it's fine. I'll fix it myself after merging.

tests/CSharp/CSharp.h
```diff
@@ -1484,11 +1484,7 @@ struct TestVariableWithoutType
...
 
-class DLL_API StringMarshall
-{
-public:
-    static const char* TestCSharpString(const char* in, CS_OUT const char** out);
-    static const wchar_t* TestCSharpStringWide(const wchar_t* in, CS_OUT const wchar_t** out);
-    static const char16_t* TestCSharpString16(const char16_t* in, CS_OUT const char16_t** out);    
-    static const char32_t* TestCSharpString32(const char32_t* in, CS_OUT const char32_t** out);
-};
+DLL_API const char* TestCSharpString(const char* in, CS_OUT const char** out);
```
ddobrev, line 1487: Exactly the kind of simplification I meant, thank you. I think that you no longer need the out parameters so please remove them and I'll merge.

josetr: I can do that but note that when I added those output parameters some issues appeared like https://github.com/mono/CppSharp/blob/7055068107057e0bfb4e8abf62dac78fedb1f8c5/src/Generator/Types/Std/Stdlib.cs#L270 and https://github.com/mono/CppSharp/blob/7055068107057e0bfb4e8abf62dac78fedb1f8c5/src/Generator/Types/Std/Stdlib.cs#L277 since __retPtr would be declared twice (one for the output parameter and one for the return)

ddobrev: So they help you test syntax too. OK, let them stay.

Issue #1443, opened 2020-10-22
Consuming CppSharp.Runtime


josetr:
If you generate 32 bit bindings then you no longer can use the nuget package to consume the `CppSharp.Runtime` since the nuget package is for 64 bit only.

An alternative would be to add an option to just embed the runtime.

PR #1445, opened 2020-10-22
Fix string marshalling bug and move hard to read code to CppSharp.Runtime

josetr:
// DEBUG: DLL_API const wchar_t* TestCSharpStringWide(const wchar_t* in, CS_OUT const wchar_t** out)

Before

```.cs
public static string TestCSharpStringWide(string @in, out string @out)
{
    var __arg1 = global::System.IntPtr.Zero;
    var __ret = __Internal.TestCSharpStringWide(@in, &__arg1);
    if (__arg1 == global::System.IntPtr.Zero)
    {
        @out = default(string);
        return default; // Bug: completely ignoring __ret
    }
    var __retPtr1 = (short*) __arg1;
    int __length1 = 0;
    while (*(__retPtr1++) != 0) __length1 += sizeof(short);
    @out = global::System.Text.Encoding.Unicode.GetString((byte*) __arg1, __length1);
    if (__ret == global::System.IntPtr.Zero)
        return default(string);
    var __retPtr0 = (short*) __ret;
    int __length0 = 0;
    while (*(__retPtr0++) != 0) __length0 += sizeof(short);
    return global::System.Text.Encoding.Unicode.GetString((byte*) __ret, __length0);
}
```

After

```.cs
		
public static string TestCSharpStringWide(string @in, out string @out)
{
    var __arg1 = global::System.IntPtr.Zero;
    var __ret = __Internal.TestCSharpStringWide(@in, &__arg1);
    @out = CppSharp.Runtime.MarshalUtil.GetString(global::System.Text.Encoding.Unicode, __arg1);
    return CppSharp.Runtime.MarshalUtil.GetString(global::System.Text.Encoding.Unicode, __ret);
}
```

### DIFF
--- a/src/Generator/Types/Std/Stdlib.cs
+++ b/src/Generator/Types/Std/Stdlib.cs
@@ -242,45 +242,18 @@ namespace CppSharp.Types.Std
             }
 
             string returnVarName = ctx.ReturnVarName;
-            string nullPtr = "global::System.IntPtr.Zero";
             if (ctx.Function != null)
             {
                 Type returnType = ctx.Function.ReturnType.Type.Desugar();
                 if (returnType.IsAddress() &&
                     returnType.GetPointee().Desugar().IsAddress())
                 {
-                    returnVarName = $"*{returnVarName}";
-                    nullPtr = "null";
+                    returnVarName = $"new global::System.IntPtr(*{returnVarName})";
                 }
             }
 
-            TextGenerator textGenerator;
-            if (ctx.Parameter == null)
-            {
-                textGenerator = ctx.Before;
-                textGenerator.WriteLine($"if ({ctx.ReturnVarName} == {nullPtr})");
-                textGenerator.WriteLineIndent($"return default({ctx.ReturnType});");
-            }
-            else
-            {
-                textGenerator = ctx.Cleanup;
-                textGenerator.WriteLine($"if ({ctx.ReturnVarName} == {nullPtr})");
-                textGenerator.WriteOpenBraceAndIndent();
-                textGenerator.WriteLine($"{ctx.Parameter.Name} = default({Type.Desugar()});");
-                textGenerator.WriteLine($"return{(ctx.Function.ReturnType.Type.IsPrimitiveType(PrimitiveType.Void) ? "" : " default")};");
-                textGenerator.UnindentAndWriteCloseBrace();
-            }
-
-            string encoding = GetEncoding().Name;
-            string type = GetTypeForCodePoint(encoding);
-            var retPtr = Generator.GeneratedIdentifier($"retPtr{ctx.ParameterIndex}");
-            var length = Generator.GeneratedIdentifier($"length{ctx.ParameterIndex}");
-            textGenerator.WriteLine($"var {retPtr} = ({type}*) {returnVarName};");
-            textGenerator.WriteLine($"int {length} = 0;");
-            textGenerator.WriteLine($"while (*({retPtr}++) != 0) {length} += sizeof({type});");
-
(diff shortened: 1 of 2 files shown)


ddobrev:
@josetr your idea is good and I'm sorry you had to write the code without knowing our internal intentions. Namely, we've been thinking of eliminating the Runtime library and generating the code instead. It's too late here now but tomorrow we can discuss it in more detail. So far one of the suggestions has been to generate it in `Std.cs` which is always created anyway. We'll continue tomorrow also with @tritao 's participation.

josetr:
@ddobrev I don't know if you guys are reading the issues but that's what I mentioned at https://github.com/mono/CppSharp/issues/1443 

> An alternative would be to add an option to just embed the runtime.

The runtime is fine, just embed it so that we can rely on it and make the generated code much easier to read and less bug prone.

ddobrev:
We had definitely not thought of this option, it's worth considering. We can all discuss it tomorrow.

josetr:
Any update ❓

tritao:
Looks good to me.

tritao:
> @josetr your idea is good and I'm sorry you had to write the code without knowing our internal intentions. Namely, we've been thinking of eliminating the Runtime library and generating the code instead. It's too late here now but tomorrow we can discuss it in more detail. So far one of the suggestions has been to generate it in `Std.cs` which is always created anyway. We'll continue tomorrow also with @tritao 's participation.

By the way, the idea we discussed was never to fully eliminate the CppSharp.Runtime project, if anything we need to expand it, and this is a good PR in that direction. What we discussed was the possibility of generating the runtime files along with the bindings.

josetr:
@tritao My motives are not a 100% pure, I'm also trying to get it ready for my consumption 😄 

I also wonder if you guys have maybe considering using a template engine. String interpolation is a bit hard to read and it's not even fast.

ddobrev:
@josetr I'm not sure what you mean by a template engine. Could you elaborate?

josetr:
@ddobrev You have a template, you give it a model, and you get some output. Like string interpolation but with more features like loops, conditionals, etc.

ddobrev:
If you're talking about generation of code, we've considered options such as CodeDom. The problem is that it's an enormous effort for little gain. Our code generation is relatively fast compared to other parts and it's them that we'd focus on if optimizing performance.

josetr:
@ddobrev Also why do we call `TextGenerator.WriteLine` rather than just using multiline lines and then indenting them? I think code would be more readable that way. Then there's also cases where WriteLine is used and its split in multiple lines like https://github.com/mono/CppSharp/blob/0cfc06001812a98d3149399649f8beec02f14303/src/Generator/Types/Std/Stdlib.cs#L227 and for me at least it makes things worse.

String interpolation is probably just fine if we can try to stick to simple one liners like I've done in this pull request. Even some complex scenarios can probably be handled in the runtime with some overloads.

tritao:
The initial version of CppSharp (called Cxxi at the time) was based on T4 templates, which is a text template system for C#. But it soon became apparent that it was a big maintenance mess and we switched to the current text generation system. Maybe there are better implementations of that idea around that could help.

As @ddobrev said, he also prototyped an approach based on CodeDOM, but if I recall correctly, it had big performance regressions compared to the current system.

If I was doing CppSharp from scratch now, I would generate everything based on a C# AST code model (which could even be implemented on top of the current AST). I think would make some stuff a lot cleaner, but we are stuck with the current approach for now. Converting the codebase would be a massive task at this point and no one has the energy to do it.

Using the multiline approach you proposed sounds interesting and something we should try. Feel free to send a PR if you feel like it.

PR #1447, opened 2020-10-24
Load symbols at startup to avoid very expensive ResolveSymbol calls everywhere

josetr:
```.cs


public static byte UChr
{
    get
    {	
	// Performance killer: ResolveSymbol calls GetCurrentDirectory / File.Exists / LoadLibrary / GetProcAddress
        var __ptr = (byte*)CppSharp.SymbolResolver.ResolveSymbol(image, "?UChr@StaticVariables@@2EB");
        return *__ptr;
    }
}

public static byte UChr
{
    get
    {
        // Now its as simple as accessing a pointer from a static variable that doesn't change
        var __ptr = (byte*)global::CSharp.__Symbols.CSharp_Native_dll._UChr_StaticVariables__2EB; 
        return *__ptr;  
    }
}

```

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpSources.cs
+++ b/src/Generator/Generators/CSharp/CSharpSources.cs
@@ -20,6 +20,8 @@ namespace CppSharp.Generators.CSharp
         public CSharpTypePrinter TypePrinter { get; set; }
         public CSharpExpressionPrinter ExpressionPrinter { get; protected set; }
 
+        Dictionary<string, LibrarySymbolInfo> LibrarySymbols = new Dictionary<string, LibrarySymbolInfo>();
+
         public override string FileExtension => "cs";
 
         public CSharpSources(BindingContext context)
@@ -95,6 +97,9 @@ namespace CppSharp.Generators.CSharp
                 UnindentAndWriteCloseBrace();
                 PopBlock(NewLineKind.BeforeNextBlock);
             }
+
+            foreach (var lib in LibrarySymbols)
+                WriteLine(lib.Value.Generate());
 
             GenerateExternalClassTemplateSpecializations();
         }
@@ -874,9 +879,12 @@ namespace CppSharp.Generators.CSharp
             TypePrinter.PushContext(TypePrinterContextKind.Native);
 
             string mangled = var.GetMangled(Context.ParserOptions.TargetTriple);
+            var libraryPath = GetLibraryOf(var);
 
-            var location = $@"CppSharp.SymbolResolver.ResolveSymbol(""{
-                GetLibraryOf(var)}"", ""{mangled}"")";
+            if (!LibrarySymbols.TryGetValue(libraryPath, out var lib))
+                LibrarySymbols[libraryPath] = lib = new LibrarySymbolInfo(libraryPath, Module.OutputNamespace + ".__Symbols");
+
+            var location = lib.GetFullVariablePath(mangled);
 
             var arrayType = var.Type as ArrayType;
             string ptr = Generator.GeneratedIdentifier("ptr");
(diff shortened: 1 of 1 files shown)


ddobrev:
@josetr could you please paste examples of the generated code? It's too long for here but you could use for example pastebin. It has syntax colouration too.

josetr:
Pastebin.com has a 512kb limit

https://pastebin.pl/view/f0da5632

tritao:
This is the interesting part:

```csharp
        static CSharp_Native_dll()
        {
            var path = "CSharp.Native.dll";
            var image = CppSharp.SymbolResolver.LoadImage(ref path);
            if (image == IntPtr.Zero) throw new System.DllNotFoundException(path);
            _FREE_CONSTANT_IN_NAMESPACE_HasFreeConstant__3HB = CppSharp.SymbolResolver.ResolveSymbol(image, "?FREE_CONSTANT_IN_NAMESPACE@HasFreeConstant@@3HB");
            _STD_STRING_CONSTANT_HasFreeConstant__3V__basic_string_DU__char_traits_D_std__V__allocator_D_2__std__B = CppSharp.SymbolResolver.ResolveSymbol(image, "?STD_STRING_CONSTANT@HasFreeConstant@@3V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@B");
            _Marker_TestDestructors__2HA = CppSharp.SymbolResolver.ResolveSymbol(image, "?Marker@TestDestructors@@2HA");
            _stringConstant_MethodsWithDefaultValues__2PEBDEB = CppSharp.SymbolResolver.ResolveSymbol(image, "?stringConstant@MethodsWithDefaultValues@@2PEBDEB");
            _intConstant_MethodsWithDefaultValues__2HA = CppSharp.SymbolResolver.ResolveSymbol(image, "?intConstant@MethodsWithDefaultValues@@2HA");
            _Destroyed_CallDtorVirtually__2_NA = CppSharp.SymbolResolver.ResolveSymbol(image, "?Destroyed@CallDtorVirtually@@2_NA");
            _variableWithFixedArrayType_TestVariableWithFixedArrayType__2PAVFoo__A = CppSharp.SymbolResolver.ResolveSymbol(image, "?variableWithFixedArrayType@TestVariableWithFixedArrayType@@2PAVFoo@@A");
            _instance_QApplication__2PEAV1_EA = CppSharp.SymbolResolver.ResolveSymbol(image, "?instance@QApplication@@2PEAV1@EA");
            _dtorCalled_VirtualDtorAddedInDerived__2_NA = CppSharp.SymbolResolver.ResolveSymbol(image, "?dtorCalled@VirtualDtorAddedInDerived@@2_NA");
            _Boolean_StaticVariables__2_NB = CppSharp.SymbolResolver.ResolveSymbol(image, "?Boolean@StaticVariables@@2_NB");
            _Chr_StaticVariables__2DB = CppSharp.SymbolResolver.ResolveSymbol(image, "?Chr@StaticVariables@@2DB");
            _UChr_StaticVariables__2EB = CppSharp.SymbolResolver.ResolveSymbol(image, "?UChr@StaticVariables@@2EB");
            _Int_StaticVariables__2HB = CppSharp.SymbolResolver.ResolveSymbol(image, "?Int@StaticVariables@@2HB");
            _Float_StaticVariables__2MB = CppSharp.SymbolResolver.ResolveSymbol(image, "?Float@StaticVariables@@2MB");
            _String_StaticVariables__2V__basic_string_DU__char_traits_D_std__V__allocator_D_2__std__B = CppSharp.SymbolResolver.ResolveSymbol(image, "?String@StaticVariables@@2V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@B");
            _ChrArray_StaticVariables__2QBDB = CppSharp.SymbolResolver.ResolveSymbol(image, "?ChrArray@StaticVariables@@2QBDB");
            _IntArray_StaticVariables__2QBHB = CppSharp.SymbolResolver.ResolveSymbol(image, "?IntArray@StaticVariables@@2QBHB");
            _FloatArray_StaticVariables__2QBMB = CppSharp.SymbolResolver.ResolveSymbol(image, "?FloatArray@StaticVariables@@2QBMB");
            _BoolArray_StaticVariables__2QB_NB = CppSharp.SymbolResolver.ResolveSymbol(image, "?BoolArray@StaticVariables@@2QB_NB");
            _VoidPtrArray_StaticVariables__2PAPEBXA = CppSharp.SymbolResolver.ResolveSymbol(image, "?VoidPtrArray@StaticVariables@@2PAPEBXA");
            _variableWithFixedPrimitiveArray__3QBEB = CppSharp.SymbolResolver.ResolveSymbol(image, "?variableWithFixedPrimitiveArray@@3QBEB");
            _variableWithVariablePrimitiveArray__3QBIB = CppSharp.SymbolResolver.ResolveSymbol(image, "?variableWithVariablePrimitiveArray@@3QBIB");
            _ArrayOfVariableSize__3QBUComplexArrayElement__B = CppSharp.SymbolResolver.ResolveSymbol(image, "?ArrayOfVariableSize@@3QBUComplexArrayElement@@B");
            _kotlin__3U_unnamed_type_kotlin___A = CppSharp.SymbolResolver.ResolveSymbol(image, "?kotlin@@3U<unnamed-type-kotlin>@@A");
        }
```

I like the overall idea, though for big libraries I wonder if there will be any startup hit.

That said, I think each static variable declaration should be generated next to the original declaration, not all together like this internal symbols class.

Optionally we could also use a lazy approach, where each variable is looked up as needed, though I can see the benefits of this approach of resolving everything at once.

Code-wise, I see a lot of references to the text generator like this ` output.`, which I see as kinda a code smell.

Maybe if you inherit from TextGenerator / CodeGenerator instead, you could clean it up a bit and the code becomes simpler.

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -3326,5 +3334,95 @@ public int GetHashCode(Parameter obj)
                 return obj.Type.Desugar().GetHashCode();
             }
         }
+
+        class LibrarySymbolInfo
+        {
+            private static readonly Regex identifierCleanerRegex = new Regex(@"[^\w]", RegexOptions.Compiled);
```
ddobrev, line 3340: You don't need mangled names for identifiers since they're limited in scope (the containing class). For the case of two variables with the same name within a library you already have a counter (or you could prepend their containing class if any). So this regex and all of the related logic is useless.

ddobrev (changes requested):

src/Generator/Generators/CSharp/CSharpSources.cs
josetr: I don't follow. The regex is there to strip characters that can potentially generate invalid identifiers and doing that with such a simple regex will likely generate collisions, so there's code to handle that. If there's useless logic can you be more specific where that is.

ddobrev: The invalid characters come from basing identifiers on mangled symbols as I've explained.

josetr: Got it but I saw the problem from another angle and I'm too tired to change that today. If I prepend the containing class, there could be a collision as well. Then I prepend the namespace or use CSharpTypePrinter.GetName or GetQualifiedName, then I'll have to remove invalid identifiers again. At the end of the day seems like the same work.

josetr:
> I like the overall idea, though for big libraries I wonder if there will be any startup hit.

I highly doubt it. `ResolveSymbol(image, name)` will just call `GetProcAddress` in windows, which is probably just a table mapping symbol names to symbol addresses. I see no reason why that would be slow. Plus its only looking for symbols that we actually need, not all of them.

> Optionally we could also use a lazy approach, where each variable is looked up as needed, though I can see the benefits of this approach of resolving everything at once. With that approach, we could also get rid of the static ctor.
> 

I'm trying to make CppSharp generate the fastest code possible, if that's not the goal of this library, I really need to know that. Using `Lazy` is slower and gives less room to the JIT to optimize things. Since the addresses are just static constants, I'm guessing the JIT will just optimize them out and the final code would be just *(byte*)(0x10203040) and we can't do better than that, even in C++.

> Maybe if you inherit from TextGenerator / CodeGenerator instead, you could clean it up a bit and the code becomes simpler.

Done.

src/Generator/Generators/CSharp/CSharpSources.cs
ddobrev: It's fine, we can continue tomorrow. This will also give me some time to think about the obstacle you've presented.

tritao:
That sounds about right, though we wouldn't necessarily need to use `System.Lazy`, we would just need to introduce a branch to check if the variable was already initialized.

You make a great point though, performance has always been one of our goals, so from that perspective I agree that this approach is optimal.

All good from me.

src/Generator/Generators/CSharp/CSharpSources.cs
ddobrev: @josetr what do you think about this: you use the name of the variable if it's free and _class_variable_ if it's a static member. In theory this could still lead to collisions but in practice hardly.
Even if you decide it's simpler to use `GetQualifiedName`, you'd only have to replace `::` which is still better than mangling and regexes.

josetr: In practice isn't good enough tho. I don't know what bothers you most: the ugly names or the extra 4 lines to avoid collisions that will only trigger 1% of the time. The regex will always be executed but it is very trivial and it's compiled as well.

The idea was to create a compile time cache for the symbols, similar to what happens when you compile an executable and in assembly you can many places dereferencing pointers that have not yet been resolved until the dlls are loaded.

You could also imagine an scenario where you create an ILRewriter and replace calls to DllImport methods with `calli`, which seems to be faster. You can use this cache to easily do that.

On a similar subject I also was wondering what's faster: using a custom `ICustomMarshaller` or just marshalling the string yourself and then calling `DllImport` methods with their native types. 

You could also merge it and improve it any way you want it later with another pull request.

ddobrev: I don't mind the idea, in fact, I tend to agree with you (and @tritao ) that it would be in general faster. I just avoid making even the generated code difficult to read unless absolutely necessary. So I think we can reach a quick solution: you could keep the regex but apply it to qualified instead of mangled names. How does this sound?
We've known about `calli` for years now but it involves a large amount of work since, as you said, the instruction isn't available in C#. The issue is further complicated by the generated code still relatively often having errors to manually correct which precludes the automatic compilation necessary for rewriting of IL. Therefore this would be a significant challenge.
I've wondered about `ICustomMarshaller` myself. We don't use it anywhere which isn't deliberate but rather just a habit. It might as well be a superior solution, including in performance. @tritao what do you think? Could we perhaps complement or replace type maps with custom marshallers?

PR #1453, opened 2020-10-25
Fix PInvokeStackImbalance when calling C++ methods

### LINKED ISSUE #1450: Calling C++ methods that return small POD structures doesn't work in .NET Framework 4.8 [windows ABI]
```.cs
struct POD
{
    int a;
    int b;
};

// ThisCall
[DllImport] POD TestPOD(IntPtr Instance);
```

Apparently calling C++ methods using this signature doesn't work in net4.0~net4.8 (windows) but it does work in Mono and it also works when targeting netcoreapp3.1 in windows.

It seems like in those versions an alternative signature is supported but I haven't tested if it will work reliably everywhere. 
`[DllImport] void TestPOD(IntPtr Instance, IntPtr POD);`

josetr:
Fixes #1450

### DIFF
--- a/src/Generator/Passes/CheckAbiParameters.cs
+++ b/src/Generator/Passes/CheckAbiParameters.cs
@@ -1,4 +1,5 @@
 ﻿using CppSharp.AST;
+using CppSharp.AST.Extensions;
 using System.Linq;
 
 namespace CppSharp.Passes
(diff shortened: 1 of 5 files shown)


tritao:
I wonder if we need to check if its a method, or `__thiscall`. Think it might be a good idea to add a few more overloads with different calling conventions, if that is possible.

josetr:
@tritao `__thiscall` is used in x86 and this issue happens in x64 as well which seems to use `__fastcall` by default so it's testing 2 calling conventions, that's why I added the `is Method` check instead. Are there more call conventions that can actually be used to call methods?

josetr:
> Nice find.
> 
> I wonder if we need to check if its a method, or `__thiscall`. Think it might be a good idea to add a few more overloads with different calling conventions, if that is possible.
> 



> @tritao `__thiscall` is used in x86 and this issue happens in x64 as well which seems to use `__fastcall` by default so it's testing 2 calling conventions, that's why I added the `is Method` check instead. Are there more call conventions that can actually be used to call methods?

tritao:
I can think of at least [__vectorcall](https://docs.microsoft.com/en-us/cpp/cpp/vectorcall?view=vs-2019). And I am not sure at the moment if you can force methods to use any other convention, like `__cdecl` or `__stdcall` (I guess not?). Anyway that's precisely why I think it makes sense to at least add more coverage on this (if the combinations are valid).

tritao:
Thanks

PR #1454, opened 2020-10-25
Remove unsupported calling convention usage

josetr:
> System.TypeLoadException : Invalid unmanaged calling convention: must be one of stdcall, cdecl, or thiscall.

### DIFF
--- a/tests/CSharp/CSharp.Tests.cs
+++ b/tests/CSharp/CSharp.Tests.cs
@@ -164,7 +164,7 @@ public unsafe class CSharpTests : GeneratorTestFixture
     {
         using (var f = new Foo())
         {
-            foreach(var pod in new[] { f.SmallPodCdecl, f.SmallPodStdcall, f.SmallPodFastcall, f.SmallPodThiscall, f.SmallPodVectorcall })
+            foreach(var pod in new[] { f.SmallPodCdecl, f.SmallPodStdcall, f.SmallPodThiscall })
             {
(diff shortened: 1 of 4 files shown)


tritao:
About `vectorcall`, looks like its not possible to support with C#, maybe we should give a warning about these in the future.

About `fastcall` test, maybe we can keep it, but either ignore the test or not define the macro under Windows 32-bits?

And there is another issue: you added this line `#define ` in `Tests.h`, which fails the build on non-Windows CI.

josetr:
In Windows 64 bit most calling conventions were translated into `DllImport(CallingConvention.Cdecl)` except `__vectorcall` which was translated into `CallingConvention.Winapi` so I don't see the point of keeping the `fastcall` test.

tritao:
OK, I think you just need to move the new tests definition to the source file instead of the header to fix the CI.

ddobrev:
Thanks. This time we'll wait for the CI to finish, and then merge.

josetr:
@ddobrev  That sounds like a good idea 😄. Maybe things would be easier if the VS solution was generated with both x64 and x86.

ddobrev:
@josetr we might do it but in a little different way. We need to update LLVM soon and its new versions seem not to build any more on 32-bit Windows any longer. Even if they were, it's a burden to support the 32-bit host for little gain.
We need to rework our AppVeyor script first, though, to make it generate twice: once for 32-bit and once for 64-bit targets, and run the tests each time. Once we've done this, we could over time merge the solutions.

PR #1458, opened 2020-10-28
Fix ambiguous conversion functions

### LINKED ISSUE #1415: Two conversion operators (T* and T&) end up generating two operators returning the same type (T*)
```.cpp
class Class
{
public:
    operator short *();
    operator short &();
};
```

OS: Windows

##### Used settings

Target: MSVC

```.cs
// DEBUG: operator short*()
public static implicit operator short*(global::TestLib.Class __op)
{
    if (ReferenceEquals(__op, null))
        throw new global::System.ArgumentNullException("__op", "Cannot be null because it is a C++ reference (&).");
    var __arg0 = __op.__Instance;
    var __ret = __Internal.OperatorConversion(__arg0);
    return __ret;
}

// DEBUG: operator short&()
public static implicit operator short*(global::TestLib.Class __op) // CS0057: Duplicated user-defined conversion in type 'Class'
{
    if (ReferenceEquals(__op, null))
        throw new global::System.ArgumentNullException("__op", "Cannot be null because it is a C++ reference (&).");
    var __arg0 = __op.__Instance;
    var __ret = __Internal.OperatorConversion_1(__arg0);
    return __ret;
}
```

josetr:
Fixes #1415

### DIFF
--- a/src/Generator/Passes/CheckAmbiguousFunctions.cs
+++ b/src/Generator/Passes/CheckAmbiguousFunctions.cs
@@ -42,18 +42,19 @@ namespace CppSharp.Passes
 
             foreach (var overload in overloads)
             {
-                if (function.OperatorKind == CXXOperatorKind.Conversion)
-                    continue;
-                if (function.OperatorKind == CXXOperatorKind.ExplicitConversion)
-                    continue;
-
                 if (overload == function) continue;
 
                 if (!overload.IsGenerated) continue;
 
-                if (CheckConstnessForAmbiguity(function, overload) ||
-                    CheckDefaultParametersForAmbiguity(function, overload) ||
-                    CheckParametersPointerConstnessForAmbiguity(function, overload))
+                var ambiguous = 
+                    function.OperatorKind == CXXOperatorKind.Conversion ||
+                    function.OperatorKind == CXXOperatorKind.ExplicitConversion
+                    ? CheckConversionAmbiguity(function, overload)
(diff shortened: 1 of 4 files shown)


src/Generator/Passes/CheckAmbiguousFunctions.cs
```diff
@@ -172,6 +173,22 @@ private static bool CheckConstnessForAmbiguity(Function function, Function overl
...
         }
 
+        private bool CheckConversionAmbiguity(Function function, Function overload)
+        {
+            var method1 = function as Method;
+            var method2 = overload as Method;
+            if (method1 == null || method2 == null)
+                return false;
+
+            if (method1.ReturnType.Type.ToString() == method2.ReturnType.Type.ToString())
```
ddobrev, line 183: Please use `Type.Equals` instead.

josetr: @ddobrev I tried that and it only works for `BuiltinType` types.

ddobrev: If so, we have a bug. What did you try: `ReturnType.Equals` or `ReturnType.Type.Equals`?

josetr: `ReturnType.Type` because `ReturnType` includes the `const` qualifier which I didn't care about.

PR #1461, opened 2020-10-30
Generate C++ for constructors of abstract types

ddobrev:


src/Generator/Generators/C/CppTypePrinter.cs
```diff
@@ -522,11 +523,23 @@ public override TypePrinterResult VisitMethodDecl(Method method)
...
+                case TypePrintScopeKind.Qualified:
+                    @class = $"{method.Namespace.Visit(this)}::";
+                    break;
+                case TypePrintScopeKind.GlobalQualified:
+                    @class = $"::{method.Namespace.Visit(this)}::";
+                    break;
+                default:
+                    @class = string.Empty;
+                    break;
+            }
```
josetr, line 542: I think we should move this to a static method otherwise `VisitMethodDecl` will keep getting bigger. We could do the same for the `noexcept` specifier.

src/Generator/Passes/SymbolsCodeGenerator.cs
```diff
@@ -169,45 +221,36 @@ private string CastIfRVReference(Parameter p)
...
-            if (isProtected)
+            Write("extern \"C\" { ");
+            Write($"{GetExporting()}void {wrapper}");
+            if (needSubclass)
                 Write("Protected");
 
             string instance = Helpers.InstanceField;
-            string @class = isProtected ? wrapper : @namespace;
+            string @class = needSubclass ? wrapper : @namespace;
             Write($"({@class}* {instance}) {{ {instance}->~{method.Namespace.Name}(); }} }};");
```
josetr, line 242: Same `extern "C"` block issue here.

src/Generator/Passes/SymbolsCodeGenerator.cs
```diff
@@ -124,31 +180,27 @@ private void WrapConstructor(Method method)
...
                 Write("extern \"C\" ");
-                if (method.Namespace.Access == AccessSpecifier.Protected)
+                if (needSubclass)
                     Write($@"{{ class {wrapper}{method.Namespace.Namespace.Name} : public {
                         method.Namespace.Namespace.Visit(cppTypePrinter)} ");
-                Write($"{{ void {wrapper}({signature}) ");
-                Write($"{{ ::new ({Helpers.InstanceField}) {@namespace}({@params}); }} }}");
-                if (method.Namespace.Access == AccessSpecifier.Protected)
+                Write($"{{ {GetExporting()}void {wrapper}({signature}) ");
+                Write($"{{ new ({Helpers.InstanceField}) {@namespace}({@params}); }} }}");
```
josetr, line 202: Why did you remove the scope operator `::` from `new` here?

josetr, line 202: Same `extern "C"` block issue here.

src/Generator/Passes/SymbolsCodeGenerator.cs
```diff
@@ -51,29 +51,91 @@ public override bool VisitMethodDecl(Method method)
...
+                        Type returnType = @abstract.OriginalReturnType.Type.Desugar();
+                        if (!returnType.IsPrimitiveType(PrimitiveType.Void))
+                        {
+                            if (returnType.IsReference())
+                            {
+                                var pointee = returnType.GetPointee().Desugar().Visit(cppTypePrinter);
+                                Write($" static const auto __value = {pointee}{{}}; return __value; ");
+                            }
+                            else
+                                Write(" return {}; ");
```
josetr, line 99: You can simplify this by removing the outer if block and moving the condition to the end.

josetr (changes requested):
I've mentioned before that a big problem is that we do not currently have a way to test these generated files. Will #1461 and #1376 change that?

This pull request also follows the trend of generating everything in a single line which I don't like and I don't see what we gain by doing that. It makes these files much harder to read and less whitespaces wont obviously make them compile faster.

josetr:
@ddobrev I think the naming is wrong here because abstract constructors do not exist. I think you are talking about implicitly-declared default constructors that are generated by the compiler when they are not marked as `deleted`.

src/Generator/Passes/SymbolsCodeGenerator.cs
```diff
@@ -51,29 +51,91 @@ public override bool VisitMethodDecl(Method method)
...
+                            continue;
+                        }
+                        Write(@abstract.Visit(cppTypePrinter) + " {");
+                        Type returnType = @abstract.OriginalReturnType.Type.Desugar();
+                        if (!returnType.IsPrimitiveType(PrimitiveType.Void))
+                        {
+                            if (returnType.IsReference())
+                            {
+                                var pointee = returnType.GetPointee().Desugar().Visit(cppTypePrinter);
+                                Write($" static const auto __value = {pointee}{{}}; return __value; ");
```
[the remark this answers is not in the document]
ddobrev: Not sure about the reason but it doesn't so let it stay, might generate less ASM or something.

src/Generator/Passes/SymbolsCodeGenerator.cs
```diff
@@ -124,31 +180,27 @@ private void WrapConstructor(Method method)
...
+            if (needSubclass)
             {
-                Write(GetDerivedType(@namespace, wrapper));
-                Write(wrapper + @namespace);
-                Write($@"({string.Join(", ", method.Parameters.Select(
-                    p => cppTypePrinter.VisitParameter(p)))})");
-                WriteLine($": {@namespace}({@params}) {{}} }};");
                 Write($"extern \"C\" {{ void {wrapper}({signature}) ");
-                WriteLine($"{{ ::new ({Helpers.InstanceField}) {wrapper}{@namespace}({@params}); }} }}");
+                WriteLine($"{{ ::new ({Helpers.InstanceField}) {wrapper}{method.Namespace.Name}({@params}); }} }}");
```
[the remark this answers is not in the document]
ddobrev: The only reason I use these is to get C-mangling i.e. predictable names of functions.

src/Generator/Passes/SymbolsCodeGenerator.cs
```diff
@@ -169,45 +221,36 @@ private string CastIfRVReference(Parameter p)
...
-            string @class = isProtected ? wrapper : @namespace;
+            string @class = needSubclass ? wrapper : @namespace;
             Write($"({@class}* {instance}) {{ {instance}->~{method.Namespace.Name}(); }} }};");
-            if (isProtected)
+            if (needSubclass)
             {
                 NewLine();
-                Write($@"extern ""C"" {{ void {wrapper}({wrapper}* {instance}) {{ {
+                Write($@"extern ""C"" {{ {GetExporting()}void {wrapper}({wrapper}* {instance}) {{ {
                     instance}->{wrapper}Protected({instance}); }} }}");
```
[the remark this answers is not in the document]
ddobrev: Same as above.

src/Generator/Passes/SymbolsCodeGenerator.cs
ddobrev: Same as above.

src/Generator/Passes/SymbolsCodeGenerator.cs
ddobrev: No idea, I guess finding and replacing. Restored.

ddobrev: Same as above.

src/Generator/Passes/SymbolsCodeGenerator.cs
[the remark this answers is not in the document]
josetr: @ddobrev You are missing my point. I told you to remove the braces, not the extern specifier.

`extern C void method();` rather than `extern C { void method(); }`.

ddobrev:
> I've mentioned before that a big problem is that we do not currently have a way to test these generated files. Will #1461 and #1376 change that?

Yes, https://github.com/mono/CppSharp/pull/1376 is going to make these symbols used in our tests and for users.

> This pull request also follows the trend of generating everything in a single line which I don't like and I don't see what we gain by doing that. It makes these files much harder to read and less whitespaces wont obviously make them compile faster.

This is generated code and quite simple too. Besides, I don't want to put too much effort in this approach because I find https://github.com/mono/CppSharp/issues/1503 superior. If we agree there, this would all be rewritten anyway.

ddobrev:
> @ddobrev I think the naming is wrong here because abstract constructors do not exist. I think you are talking about implicitly-declared default constructors that are generated by the compiler when they are not marked as `deleted`.

It was mostly for shortness, it's actually "constructors of abstract classes". I'll change the message.

src/Generator/Generators/C/CppTypePrinter.cs
ddobrev: I agree it's neater, done.

src/Generator/Passes/SymbolsCodeGenerator.cs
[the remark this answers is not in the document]
ddobrev: Oh, OK.

src/AST/ClassExtensions.cs
```diff
@@ -30,6 +30,21 @@ public static class ClassExtensions
             }
         }
 
+        public static List<Method> GetAbstractMethods(this Class @class)
+        {
+            var abstractMethods = @class.Methods.Where(m => m.IsPure).ToList();
+            var abstractOverrides = abstractMethods.Where(a => a.IsOverride).ToList();
```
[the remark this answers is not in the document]
ddobrev: Done.

src/Generator/Passes/SymbolsCodeGenerator.cs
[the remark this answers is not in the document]
josetr: The new code no longer has the issue because you are using a C cast that basically says "I know what I'm doing" but it may be more prone to give warnings. What I meant before was this:

```.cpp
int& method() {
  static const int value; 
  return value; // error- cant remove const
}

const int& method() {
  static int value;
  return value; // ok - adding const is safe
}
```

src/Generator/Passes/SymbolsCodeGenerator.cs
ddobrev: Done.

PR #1468, opened 2020-11-02
Create VTables class to cache delegates pointing to native virtual methods

josetr:


src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -1580,98 +1580,127 @@ public void GenerateVTable(Class @class)
...
             NewLine();
+
+            if (!hasDynamicBase)
+                WriteLine("protected CppSharp.Runtime.VTables __vtables;");
+
+            WriteLines($@"
+            internal {(hasDynamicBase ? "override" : "virtual")} CppSharp.Runtime.VTables __VTables
+            {{ 
+                get {{
+                    if (__vtables.IsEmpty)
```
ddobrev, line 1677: Is this necessary? We initialize tables in constructors.

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -1580,98 +1580,127 @@ public void GenerateVTable(Class @class)
...
+                }}
+
+                set {{        
+                    __vtables = value;
+                }}
+            }}
+
+            internal {(hasDynamicBase ? "override" : "virtual")} void SetupVTables(bool destructorOnly = false)
+            {{
+                if (__VTables.IsTransient)
```
ddobrev, line 1689: See above.

src/Generator/Generators/CSharp/CSharpSources.cs
josetr: `CreateInstance` and `GetOrCreateInstance` both have a `skipVTables` parameter which is used in some cases.

PR #1471, opened 2020-11-03
Fix GenerateSequentialLayout implementation

josetr:


src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -764,24 +767,63 @@ public override void GenerateClassSpecifier(Class @class)
...
-            Write($"internal {retType}");
-            if (field.Expression != null)
-            {
-                var fieldValuePrinted = field.Expression.CSharpValue(ExpressionPrinter);
-                Write($" = {fieldValuePrinted}");
+            if (fields.Count > 1)
+            { 
+                for (var i = 1; i < fields.Count; ++i)
+                {
+                    if (fields[i].Offset == fields[i - 1].Offset)
```
ddobrev, line 781: Why is this necessary? Unless the type is a union which is already checked, fields at the same offset are impossible. If our AST reports this in any case, then it has a bug.

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -764,24 +767,63 @@ public override void GenerateClassSpecifier(Class @class)
...
+                if (sequentalLayout && fields[i].IsVTablePtr)
+                {
+                    var nativePointerSize = (Context.TargetInfo.PointerWidth / 8);
+                    var padding = 
+                        i + 1 < fields.Count
+                        ? fields[i + 1].Offset - field.Offset - nativePointerSize
+                        : (uint)@class.Layout.Size - nativePointerSize;
+
+                    if (padding > 0 && padding <= @class.Layout.Size)
+                        WriteLine($"public fixed byte {field.Name}Padding[{padding}];");
```
ddobrev, line 822: I didn't get why this padding is necessary. It looks like a bug, a field missing from the AST.

src/Generator/Generators/CSharp/CSharpSources.cs
josetr: > fields at the same offset are impossible. 

```
class Time
{
public:
    // a & b will share the same offset
    union 
    {
        int a;
        double b;
    };
    int test;
};```

ddobrev: Indeed they are – we just inline fields of empty types. Perhaps we need to reconsider this.

ddobrev:
Merged at https://github.com/mono/CppSharp/commit/0d32121eee5b814493c86656f2f67e59ed21696a.

PR #1476, opened 2020-11-07
Fallback to LayoutKind.Explicit when a custom alignment is specified.

josetr:


src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -782,6 +789,12 @@ private bool CanUseSequentialLayout(Class @class)
                 {
                     if (fields[i].Offset == fields[i - 1].Offset)
                         return false;
+
+                    if (fields[i].QualifiedType.Type.Desugar().TryGetDeclaration(out Declaration declaration))
+                    {
+                        if (declaration.AlignAs != 0)
```
ddobrev, line 795: Merge the conditions.

src/CppParser/Parser.cpp
```diff
@@ -4188,6 +4194,20 @@ Declaration* Parser::WalkDeclaration(const clang::Decl* D)
     return Decl;
 }
 
+int Parser::GetAlignAs(const clang::AlignedAttr* alignedAttr)
+{
+    using namespace clang;
+
+    if (alignedAttr->isAlignas())
+    {
+        if (!alignedAttr->isAlignmentErrorDependent() &&
```
ddobrev, line 4203: Merge the conditions.

josetr: Originally it was a much longer method where that wasn't possible. I've also removed `using namespace clang` which is no longer needed.

src/Generator/Generators/CSharp/CSharpSources.cs
josetr: I did that to avoid a long line but I've created a variable for the "desugared" type and merged both conditions.

src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -812,13 +830,10 @@ private void GenerateClassInternalsFields(Class @class, bool sequentalLayout)
                 }
                 WriteLine(";");
 
-                // HACK: work around the lack of alignment in StructLayout
```
ddobrev, line 815: Link https://github.com/dotnet/runtime/issues/44378 here.

PR #1486, opened 2020-11-08
Avoid generating an interface with a Dispose method if we already inherit from IDisposable.

josetr:


src/Generator/Generators/CSharp/CSharpSources.cs
```diff
@@ -480,7 +480,9 @@ private void GenerateInterface(Class @class)
 
             GenerateClassSpecifier(@class);
 
-            if (@class.Bases.Count == 0)
+            var inherithFromIDisaposable = @class.Bases.Count == 0;
+
+            if (inherithFromIDisaposable)
```
ddobrev, line 485: I think you should use `@class.HasRefBase()` instead.

ddobrev (changes requested):

src/Generator/Generators/CSharp/CSharpSources.cs
josetr: Is that needed tho? we only generate interfaces for ref types. Interfaces on structures cause boxing. Maybe I should use `@class.HasBase` instead which is the prettier version of this check?

ddobrev: There are details such as value types, ignored types, etc. This is why using existing methods is recommended.

josetr: The method already contains a check to see if the type was generated at the beggining. The only extra check that `HasRefBase` does is checking if it's a `RefType` but for performance reasons we don't want to use interfaces on structs because then boxing can happen. Like I said, maybe the prettier `HasBase` version makes more sense but if you want me to use the other one let me know and I will do it.

ddobrev: It's not the class itself being ignored, it's about the base. If the base is ignored, the number of bases would be > 1 and your code will skip the interface.

ddobrev: In addition, if the base is a value type, you get > 0 but no base because value types are sealed. So once more you'll skip the interface when you shouldn't.

josetr: I got it but then shouldnt the property "HasBase" check if the class is generated as well?

ddobrev: We haven't thought about it. It might be useful but it's for another PR.

josetr: We gonna handle this potential issue in another PR. I've created an issue for it at https://github.com/mono/CppSharp/issues/1488

PR #1489, opened 2020-11-09
Fix warnings about unnecessary using directives.

josetr:


tests/NamespacesDerived/NamespacesDerived.cs
```diff
@@ -18,7 +18,7 @@ public override void Setup(Driver driver)
             base.Setup(driver);
             driver.Options.GenerateDefaultValuesForArguments = true;
             driver.Options.GenerateClassTemplates = true;
-
+            driver.Options.DependentNameSpaces.Add("System.Runtime.CompilerServices");
```
ddobrev, line 21: We must find a way to work without this line. We have a mechanism which generates dependencies between generated libraries corresponding to those between native ones. The requirement for dependent modules to add this directive by hand breaks the automation.

ddobrev (changes requested):

PR #1514, opened 2020-11-19
A temporary hack to deal with a bug in MSVC v142 - VS 2019 build tools

josetr:


ddobrev:
Please link the relevant issue so that we know when it's fixed: https://github.com/microsoft/STL/issues/1300.

Issue #1518, opened 2020-11-20
Consider supporting passing value types by reference `ref` / `in`


josetr:
This seems by design since `in` is only supported since C# 7.2 but even if it was supported when CppSharp was created, it's more cumbersome to pass these kind of parameters since you always have to place the `in` modifier at the call site, unlike in C++. This effort usually only makes sense when the value type is very expensive to copy.

#### C++ signature
```.cpp
void TakeConstCharRef(const char& c);
```

#### Generated code
```.cs
void TakeConstCharRef(char c);
```

PR #1519, opened 2020-11-20
Ignore test because something that has not been implemented cant be tested

josetr:
See https://github.com/mono/CppSharp/issues/1518.


ddobrev:
Make the commit message more precise.

josetr:
I assumed someone would change the commit message before merging it since we had different opinions about this test. The option is still there :smile:.

ddobrev:
> I assumed someone would change the commit message before merging it since we had different opinions about this test. The option is still there 😄.

OK, no trouble. Let me merge now.

PR #1524, opened 2020-11-23
Upgrade to .NET Core 3.1 and improve build system

josetr:


build/premake/premake.extensions.lua
```diff
@@ -0,0 +1,66 @@
...
+  base(prj);
+
+  if prj.dotnetxml ~= nil then
+    for _, xml in ipairs(prj.dotnetxml) do
+        premake.out(string.gsub(string.gsub(xml, "\r", ""), "\n", premake.eol()))
+    end
+  end
+end)
+
+-- https://github.com/premake/premake-core/issues/1061#issuecomment-441417853
```
ddobrev, line 53: Please add _HACK_ here.

build/premake/premake.fixes.lua
```diff
@@ -0,0 +1,20 @@
+-- https://github.com/premake/premake-core/issues/1559
```
ddobrev, line 1: Please add _HACK_ here.

build/premake/premake.fixes.lua
```diff
@@ -0,0 +1,20 @@
...
+premake.override(premake.vstudio.vc2010, "targetFramework", function(oldfn, prj)
+  if string.endswith(prj.filename, ".CLI") then 
+    local action = premake.action.current()
+    local tools = string.format(' ToolsVersion="%s"', action.vstudio.toolsVersion)
+    local framework = prj.dotnetframework or action.vstudio.targetFramework or "4.0"    
+    premake.w('<TargetFramework>%s</TargetFramework>', framework)  
+  end
+end)
+  
+-- https://github.com/premake/premake-core/issues/1549
```
ddobrev, line 11: Please add _HACK_ here.

ddobrev (changes requested):
Mark the hacks we're forced to use so that they stand out and are easy to find later. This way we can remove them over time as premake fixes its bugs.

build/premake/premake.fixes.lua
josetr: Do you want me to rename the file to premake.hacks instead?

josetr:
> Mark the hacks we're forced to use so that they stand out and are easy to find later. This way we can remove them over time as premake fixes its bugs.

Do you want me to rename the file to premake.hacks instead?

This is the reason why I specific files for these so they would be easy to find. Adding "hack" or "fix" is redundant.

josetr:
> Mark the hacks we're forced to use so that they stand out and are easy to find later. This way we can remove them over time as premake fixes its bugs.

This is the exact reason why I created specific files for these "fixes" and "extensions" so that they would be extremely easy to find. They are isolated in the same folder as the premake executable. Adding "hack" or "fix" feels redundant.

I can rename the file to `premake.hacks.lua` if you want.

ddobrev:
> > Mark the hacks we're forced to use so that they stand out and are easy to find later. This way we can remove them over time as premake fixes its bugs.
> 
> This is the exact reason why I created specific files for these "fixes" and "extensions" so that they would be extremely easy to find. They are isolated in the same folder as the premake executable. Adding "hack" or "fix" feels redundant.
> 
> I can rename the file to `premake.hacks.lua` if you want.

I hadn't noticed the names of the files. Fair enough.

PR #1533, opened 2020-11-28
publish artifacts

IIFE:
publish cppsharp artifacts along with required dependencies

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,4 +59,5 @@ jobs:
           path: |
             bin/**/CppSharp*
             bin/**/Std-symbols*
+            bin/**/lib/**
             include/**/*.h
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,3 +51,12 @@ jobs:
       - name: Test
         shell: bash
         run: build/test.sh -platform $PLATFORM
+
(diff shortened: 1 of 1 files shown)


tritao:
This is not correct, as the paths are not correct, builds will be under a `bin/Release_x86` or similar naming scheme.

You are also not adding the include folder to the artifacts.

If you want to send a PR, take my branch and just add the `include` folder to the paths.

IIFE:
@tritao OK I'll do that thanks.

So what paths do I need to set so that it only upload CppSharp dlls and the required std-symbols one? Is it:
```
bin/Release_x86/CppSharp*
```

?

tritao:
You can try `bin/Release_-${{ matrix.config.platform }}/...` I guess.

By the way, you might have to publish a branch under your own fork to test the CI changes.

I think by default, for security reasons, Actions will not take into account workflow changes in submitted PRs.

IIFE:
@tritao The actions seem to have run fine, they just couldn't find the paths. I've committed a change to try and use globbing instead.

IIFE:
@tritao That's all working fine now. I've tested the published artifacts and it integrates as expected. It's publishing the minimal required files.

Let me know please if you're happy with this PR.

Thanks.

tritao:
Cool thanks for the PR.

PR #1550, opened 2020-12-14
Remove win x86 from the NuGet for smaller size

ddobrev:
Too few people would use a win 32 host nowadays.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -48,7 +48,7 @@
     <DotNetCmd></DotNetCmd>
   </PropertyGroup>
 
-  <Target Name="prepack" DependsOnTargets="Build" Condition="'$(IsPackable)' == 'true'">
+  <Target Name="prepack" DependsOnTargets="Build" Condition="'$(IsPackable)' == 'true' AND '$(Platform)' == 'x64'">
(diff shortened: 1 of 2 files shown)


Directory.Build.props
```diff
@@ -48,7 +48,7 @@
     <DotNetCmd></DotNetCmd>
   </PropertyGroup>
 
-  <Target Name="prepack" DependsOnTargets="Build" Condition="'$(IsPackable)' == 'true'">
+  <Target Name="prepack" DependsOnTargets="Build" Condition="'$(IsPackable)' == 'true' AND '$(RID)' != 'win-x86'">
```
josetr, line 51: Perhaps its better to just use `$(PlatformTarget)' == 'x64'` instead. We can discuss how to deploy more configurations when the the time comes.

src/Parser/CppSharp.Parser.csproj
```diff
@@ -29,7 +29,7 @@
     <Copy SourceFiles="@(ClangHeaders)" DestinationFiles="@(ClangHeaders->'$(PackageDir)contentFiles\any\any\lib\%(RecursiveDir)%(Filename)%(Extension)')" />
   </Target>
 
-  <Target Name="PackNativeLibs" AfterTargets="prepack">
+  <Target Name="PackNativeLibs" AfterTargets="prepack" Condition="'$(RID)' != 'win-x86'">
```
josetr, line 32: Same as above.

PR #1555, opened 2020-12-29
Make auto-compiling of C# work with .NET Core

ddobrev:


src/Generator/Driver.cs
```diff
@@ -353,66 +350,55 @@ private void WriteGeneratedCodeToFile(string file, string generatedCode)
...
-            compilerParameters.ReferencedAssemblies.AddRange(module.ReferencedAssemblies.ToArray());
-
-            Diagnostics.Message($"Compiling {module.LibraryName}...");
-            CompilerResults compilerResults;
-            using (var codeProvider = new CSharpCodeProvider(
-                new Dictionary<string, string> {
-                    { "CompilerDirectoryPath", ManagedToolchain.FindCSharpCompilerDir() } }))
+            string csproj = Path.Combine(Options.OutputDir, $"{module.LibraryName}.csproj");
+            File.WriteAllText(Path.Combine(Options.OutputDir, "Directory.Build.props"), "<Project />");
+            File.WriteAllText(csproj,
```
tritao, line 358: Please move this out of the driver into its own generator.

ddobrev: Could you explain, please? Do you want me to move the code to an external function? I don't get what generator you want. The only generators I'm aware of are all for source code and they have nothing to do with this piece because it doesn't work with the AST.

PR #1571, opened 2021-01-27
Extracting interfaces

MattKey95:
Most of the code is the same as the MultipleInheritancePass with a minor change to the VisitClassDecl method. Left it as a new class for now but if people think it would better to refactor to reduce code duplication that can be done. 

Also, off the back of this, it would be nice in the MultiInheritancePass to be able to specify opt out of inheritance completely (only composition). From what I understand, it defaults to inherit from the first base and then all the other base classes are transformed to interfaces. Using both these passes you end up inheriting from a class that has an interface which seems a little odd to me, especially if the class is abstract (although I have limited testing on this scenario).


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/mono/CppSharp?pullRequest=1571) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/mono/CppSharp?pullRequest=1571) before we can accept your contribution.<br/><br/>:x: MattKey95 [sign now](https://cla.dotnetfoundation.org/mono/CppSharp?pullRequest=1571)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/mono/CppSharp?pullRequest=1571) it.</sub>

tritao:
@MattKey95 Thanks for the contribution so far.

I think it would make sense to try to refactor to reduce duplication as much as possible before merging.

MattKey95:
Something more like this?

tritao:
Awesome, looks great.

PR #1574, opened 2021-01-31
Generate nuget package for CppSharp.Runtime

josetr:


InflexCZE:
Well that was fast 👍 

I think the "Generating bindings" doc should be amended as well to inform about this new requirement to include `CppSharp.Runtime` to "production app".

caihongxu:
When will this be merged? Really looking forward to it to reduce my binary size..

InflexCZE:
There are currently two ways to overcome absence of standalone NuGet.
1) Download official CppSharp NuGet, open it as zip archive, extract CppSharp.Runtime.dll, link it to your production app as standalone managed assembly
2) Compile sources from src/Runtime directly into your app

tritao:
I don't remember why we didn't merge this at the time... @josetr?

josetr:
> I don't remember why we didn't merge this at the time... @josetr, do you?

I believe I just forgot about it as I've been using a custom version of CppSharp. Someone other than me will need to publish the package on nuget.org tho.

caihongxu:
Thank you everyone! Look forward for it to show up in Nuget once published! The reason I felt reluctant with other hacks mentioned is the worries about version inconsistencies between the generated code and the run time. Nuget would make it a lot easier to manage!

Issue #1598, opened 2021-04-12
Two issues when generating classes with embedded arrays of structs 


jpnoir:
Target: C#. Will submit PR.

Consider the structs

```c++
struct DLL_API EmbeddedStruct
{
    uint64_t ui64;
};
struct DLL_API StructWithEmbeddedArrayOfStructObjectAlignment
{
    bool boolean;
    EmbeddedStruct embedded_struct[2];
};
```

*Issue 1:* With the existing implementation, `StructWithEmbeddedArrayOfStructObjectAlignment` is translated to

```c#
[StructLayout(LayoutKind.Sequential, Size = 24)]
public partial struct __Internal
{
    internal byte boolean;
    internal fixed byte embedded_struct[16];
    
    ...
}
```

Note that there is no padding following the `boolean` field. Since `embedded_struct` has been translated to a `fixed byte[]`, on the C# side, its offset is 1. On the C++ side, `embedded_struct` has an offset of 8. See `CSharpTests.TestAlignment` in the associated PR for a repro.

*Issue 2:* Observe that for the same struct(s), the accessor method for `Embeddedstruct` is

```c#
get
{
    global::CSharp.EmbeddedStruct[] __value = null;
    if (((__Internal*)__Instance)->embedded_struct != null)
    {
        __value = new global::CSharp.EmbeddedStruct[2];
        for (int i = 0; i < 2; i++)
            __value[i] =  global::CSharp.EmbeddedStruct.__CreateInstance(*((global::CSharp.EmbeddedStruct.__Internal*)&(((__Internal*)__Instance)->embedded_struct[i * sizeof(global::CSharp.EmbeddedStruct.__Internal)])));
    }
    return __value;
}
```

In particular, note that every time the accessor is called, it initializes a new instance of the `EmbeddedStruct[]` array by calling `__CreateInstance` using the overload that makes a copy of the relevant values in `embedded_struct[]` and creates a new instance of managed class, `EmbeddedStruct`, with that copy. Any updates to the array elements are written to the copy and not to `embedded_struct[16]`. The copied unmanaged memory is leaked. See `CSharpTests.TestEmbeddedArrayOfStructAccessor` in the associate PR for a repro.

##### Proposed fixes:

[...]

PR #1599, opened 2021-04-12
Embedded array of struct issues. Resolves mono/CppSharp#1598

jpnoir:
See issue #1598 for description.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ src/generator/generator
 *.opensdf
 *.pdb
 *.config
+!nuget.config
 *.vcxproj
 *.filters
 *.sln
--- a/nuget.config
+++ b/nuget.config
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+  </packageSources>
+</configuration>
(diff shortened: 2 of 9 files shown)


tritao:
Thanks for the fixes, it looks good.

The Windows CI with some Nuget issues, I've re-run all jobs in the meanwhile to see if it was just some temporary thing.

jpnoir:
Thanks, I have no idea why the restores (NUnit, etc.) are failing on the windows side. I didn’t change anything there and they obviously worked locally.

tritao:
Even though the PR CI was fully green, master is now complaining: https://github.com/mono/CppSharp/runs/2350856410#step:6:11

I am not sure why and don't really have time to check this right now.

If you got time to check this, it would be appreciated.

jpnoir:
I don't understand it either. The message(s) indicate some sort of CI-time SDK configuration error. 

I tried pulling the newly merged main branch and (of course) had no trouble restoring / building locally.

The projects that are failing are all C++/CLI projects which seems to match the docs pointed to by the URL in the error message. I haven't yet found where Visual Studio / MS Build establishes the expected AppHost version. However, if I'm interepreting the error message correctly, the version it found is (3.1.14). That version is included in the latest VS2019 version, v16.9.4. The previous Visual Studio version (v16.9.3) uses 3.1.13. The doc that the URL in the message refers you to suggests that scenario might be the issue: 

>  "This is typically caused by having a newer SDK than the one included in Visual Studio for C++/CLI projects. "

I can't see any version info for Visual Studio in the logs. Could this be the problem? Is it possible to update Visual Studio on the CI server?

PR #1603, opened 2021-04-27
Enums from macros issues

jpnoir:
##### Problem descriptions

*Issue 1:* Consider the macro definition:

```c++
#define MY_MACRO_TEST3_1    2147483647	// Int32.MaxValue
```

The resulting macro definition is:

```c#
public enum MyMacroTest3Enum1 : uint
{
    MY_MACRO_TEST3_1 = 2147483647
}
```

Note that the underlying base type is set to a `uint` rather than an `int` because `Library.GetUnderlyingTypeForEnumValue(ulong value)` uniformly use less-than comparisons where less-or-equal comparisons were wanted/intended.

*Issue 2:* Consider the macro definition:

```c++
#define MY_MACRO_TEST3_2    2147483648	// Int32.MaxValue + 1
```

The resulting macro definition is:

```c#
public enum MyMacroTest3Enum2 : sbyte
{
    MY_MACRO_TEST3_2 = 0
}
```

This result is due to the fact unless a literal type suffix (e.g. 'u') is present, the `ExpressionEvaluator` always attempts to parse numbers as an Int32. The method `Library.ParseEnumItemMacroExpression` eats the exception and sets the return value to zero with no hint that anything went wrong.

*Issue 3:* If you attempt to work around *Issue 2* by using a type suffix as in:

```c++
#define MY_MACRO_TEST3_2    2147483648u	// Int32.MaxValue + 1
```

the translator throws an `InvalidCastException` in `Library.ParseEnumItemMacroExpression` where it attempts to first cast the boxed `object` value returned by the `ExpressionEvaluator` to an `int` so that it can then cast it to a `ulong`. Since the actual boxed value returned from the `ExpressionEvaluator` is a `uint`, the cast fails.

*Issue 4:* Consider the macro definition:

```c++
#define MY_MACRO_TEST3_3    -101
```

If you attempt to convert this macro definition into an `enum` item, the translator converts it to:

```c#
public enum MyMacroTest3Enum3 : uint
{
    MY_MACRO_TEST3_3 = 18446744073709551515
}
```

Note that the `ExpressionEvaluator` handles negative numbers, but the `GenerateEnumFromMacros`  mechanism does not. In particular, `Library.ParseEnumItemMacroExpression` silently coerces the negative value to an unsigned long resulting in the value 18446744073709551515. Negative numbers are of course supported in C# `enum` items.

Note that the coerced result won't fit into a `uint` which is the underlying base type for the `enum`. That results in compile-time errors of the generated C#. The issue here is that `Library.GetUnderlyingTypeForEnumValue` types the value as a `PrimitiveType.ULong` rather than a `PrimitiveType.ULongLong`.

##### Proposed improvements:

- Mimic the C# compiler when parsing numbers in `ExpressionEvaluator.EvaluateNumber` to implicitly type values. That is, in C# if you write 

  ```c#
  var x = 2147483648;  // int.MaxValue + 1
  ```

  'x' will be typed as `uint`. Extend `ExpressionEvaluator` to do the same.

- Use `Enumeration.BuiltinType` / `Enumeration.Type` to track the underlying type of the enumeration while it is being populated. This allows us to support any primitive (numeric) type for `enum` values.

- Use the type information described above when populating symbols provided to the `ExpressionEvaluator` so that symbol values are appropriately typed when used in expressions evaluated by the `ExpressionEvaluator`.

- Add more care when casting from the boxed type returned from `ExpressionEvaluator` to the actual numeric type to avoid cast exceptions.

- Fix issues in `Library.GetUnderlyingTypeForEnumValue` and add support for supplying a minimum value so that primitive types representing signed values are supported.

- Use the `Diagnostics.Warning` mechanism to at least write something to the console if there are issues evaluating an expression.

- More tests.


tritao:
Thanks for all the fixes, and all the tests!

PR #1604, opened 2021-04-29
Const char star to string issues

jpnoir:
#### const char* to string issues 

Consider the test class (from CSharp.h)

```c++
class DLL_API TestString
{
public:
    TestString();
    ~TestString();
    const wchar_t* unicodeConst;
    wchar_t* unicode;
};
```

The existing test (CSharp.Tests.CSharp.TestUnicode) verifies that an initial string value for `unicodeConst` established in `TestString`'s native constructor is correctly marshaled to the C# wrapper class.

On the other hand, when a value is assigned to `unicodeConst` in C#, the following wrapper class property is used:

```c#
public string UnicodeConst
{
    get
    {
        return CppSharp.Runtime.MarshalUtil.GetString(
            global::System.Text.Encoding.Unicode, 
            ((__Internal*)__Instance)->unicodeConst);
    }

    set
    {
        byte[] __bytes0 = global::System.Text.Encoding.Unicode.GetBytes(value);
        fixed (byte* __bytePtr0 = __bytes0)
        {
            ((__Internal*)__Instance)->unicodeConst = (__IntPtr) new 
                global::System.IntPtr(__bytePtr0);
        }
    }
}
```

There are several problems here:

1. `Unicode.GetBytes(value)` used in the setter does not nul-terminate the resulting byte array as expected by the native C++ code, and by the `GetString` method used in the property's getter.

2. If the C# caller attempts to set a `const char*` value to `null` via this setter, `Unicode.GetBytes(value)` will throw a `NullReferenceException`.

3. In the setter, the `IntPtr` value corresponding to `unicodeConst` is set to an *unpinned* pointer to  otherwise unreferenced managed memory since the scope of the `fixed` operator ends at its closing brace. That is (from the Microsoft docs):

   > After the code in the statement is executed, any pinned variables are unpinned and subject to garbage collection. Therefore, do not point to those variables outside the `fixed` statement. 

   As a result, the memory that `unicodeConst` points to will get recycled the next time the garbage collector runs. The value of `unicodeConst` is indeterminate.

4. Property wrapper support for `const char16_t*` and `const char32_t*` was only partially implemented resulting in compile-time errors for the generated code if these character types were used for fields in classes.

#### Proposed improvements:

1. Change the setter for `const char*` values to allocated unmanaged memory rather than using managed memory. 
2. Track whether unmanaged memory has been allocated by the wrapper for a `const char*` field or not.
3. When setting a string value, check whether unmanaged memory has been previously allocated and if so, free it.
4. If the string value is `null`, set the `IntPtr` to `IntPtr.Zero`.
5. Arrange to nul-terminate the unmanaged byte array in the setter by:
   - Encoding the string as before, allocating sufficient unmanaged memory to hold the byte array plus nul termination character and copying the managed byte array to the unmanaged buffer.
   - Or, if the desired encoding is one of the cases that Marshal can handle: Unicode or ANSI, use `Marshal.StringToHGlobalUni` or `Marshal.StringToHGlobalAnsi` to encode and nul-terminate the byte array without an extraneous copy.
6. Perform a deep copy of wrapper-allocated strings in the copy constructor to avoid having to implement more complex tracking (e.g. reference counting) for a presumably rare case.
7. Add support for generating property wrappers for types of `const char16_t*` and `const char32_t*`.
8. If unmanaged memory is allocated by the wrapper, free it in the Dispose implementation.

The above setter becomes:

```c#
set
{
    if (__unicodeConst_OwnsNativeMemory)
        Marshal.FreeHGlobal(((__Internal*)__Instance)->unicodeConst);
    __unicodeConst_OwnsNativeMemory = true;
    if (value == null)
    {
        ((__Internal*)__Instance)->unicodeConst = global::System.IntPtr.Zero;
        return;
    }
    var __bytePtr0 = Marshal.StringToHGlobalUni(value);
    ((__Internal*)__Instance)->unicodeConst = (__IntPtr) __bytePtr0;
}
```

The Dispose method becomes:

```c#
public virtual void Dispose(bool disposing)
{
    if (__Instance == IntPtr.Zero)
        return;
    NativeToManagedMap.TryRemove(__Instance, out _);
    if (disposing)
        __Internal.dtor(__Instance);
    if (__unicodeConst_OwnsNativeMemory)
        Marshal.FreeHGlobal(((__Internal*)__Instance)->unicodeConst);
    if (__ownsNativeInstance)
        Marshal.FreeHGlobal(__Instance);
    __Instance = IntPtr.Zero;
}
```

The copy constructor becomes:

```c#
public TestString(global::CSharp.TestString _0)
{
    __Instance = Marshal.AllocHGlobal(sizeof(global::CSharp.TestString.__Internal));
    __ownsNativeInstance = true;
    NativeToManagedMap[__Instance] = this;
    *((global::CSharp.TestString.__Internal*) __Instance) = 
        *((global::CSharp.TestString.__Internal*) _0.__Instance);
    if (_0.__unicodeConst_OwnsNativeMemory)
        this.UnicodeConst = _0.UnicodeConst;
}
```

##### Assumption

We don't support APIs where both the native and managed sides allocate memory for a string within the lifetime of an object. If an API requires this, custom methods that know the ownership protocol will be required.


tritao:
Another excellent PR, thank you!

PR #1615, opened 2021-05-28
Fix incorrect dtor call for non-owned instances

szilvaa:
This PR fixes an issue that caused the generated code to call the dtor for the underlying native instance during Dispose() even when the managed wrapper doesn't own the native instance. I'm quite surprised that nobody else complained about this behavior. 

Steps to repro:
Commit 0875efb in this PR introduces a test that demonstrates the problem.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,7 +2,6 @@
   <Import Project="build/config.props" />
 
   <PropertyGroup>
-    <CI>True</CI>
     <RootDir>$(MSBuildThisFileDirectory)</RootDir>
     <Platforms>x86;x64</Platforms>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,6 +2,7 @@
   <Import Project="build/config.props" />
 
   <PropertyGroup>
+    <CI>True</CI>
     <RootDir>$(MSBuildThisFileDirectory)</RootDir>
     <Platforms>x86;x64</Platforms>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
--- a/src/CppParser/Bindings/CLI/AST.cpp
+++ b/src/CppParser/Bindings/CLI/AST.cpp
@@ -9,7 +9,7 @@
 using namespace System;
 using namespace System::Runtime::InteropServices;
 
-CppSharp::Parser::AST::NativeLibrary::NativeLibrary(::CppSharp::CppParser::AST::NativeLibrary* native)
+CppSharp::Parser::AST::NativeLibrary::NativeLibrary(struct ::CppSharp::CppParser::AST::NativeLibrary* native)
     : __ownsNativeInstance(false)
 {
     NativePtr = native;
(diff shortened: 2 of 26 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/mono/CppSharp?pullRequest=1615) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/mono/CppSharp?pullRequest=1615) before we can accept your contribution.<br/><br/>:x: szilvaa [sign now](https://cla.dotnetfoundation.org/mono/CppSharp?pullRequest=1615)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/mono/CppSharp?pullRequest=1615) it.</sub>

ddobrev:
@szilvaa thank you very much for your help. I'm afraid, though, that I'd need to take a closer look at such a sensitive piece of code which I cannot do right now. I do have a quick question, though: why would you call `Dispose` on an object you don't want destroyed? `Dispose` is (almost) equivalent to `delete` in `C++` and you don't call the latter on objects you actually need, do you?

szilvaa:
@ddobrev Yep. I agree this change is "sensitive" and requires close scrutiny. So I can offer 2 justifications for this change:
1. When a class implements `IDisposable` the client should be able call `Dispose()`. They shouldn't be forced somehow "know" when this is safe/desired/correct. It is the responsibility of the class to do the right thing. There are some C# language features that automatically detect `IDisposable` and call `Dispose()` so these calls can happen even in innocent looking code (e.g. Enumerator in `foreach` is automatically disposed).
2. The current behavior is rather "strange". When a non-owned wrapper is disposed the native dtor is called. This means that the dtor of the native class will be called *TWICE* because the native code (which owns the object) will also call the dtor! (In case of virtual dtor this doesn't happen because the wrapper hijacks vtable entry for the dtor but for non-virtual dtor this double call is a problem.)

Thanks for your time reviewing the PR!

ddobrev:
You got me with 2. However, I think your new parameter is useless. We already use `disposing` to check whether to call the native constructor. Please check if you can extend its usage instead of introducing one more.

szilvaa:
So I think `Dispose(bool disposing)` method comes from the [recommended implementation](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose) of the `Disposable` interface. The `disposing` argument is set to `true` when the method is called from `IDisposable.Dispose()` and it is set of `false` when called from the finalizer.
When `disposing` == `true` the `Dispose(bool disposing)` method is supposed to call `Dispose()` on any objects it has that implements `IDisposable`. 

The code generated by CppSharp does not use this recommended pattern. It is sort of confused about the meaning of `disposing`.

I'm going to try to update the code to conform to the recommended pattern. This will allow me to remove the extra argument.

szilvaa:
Ok. It is not possible to eliminate this extra argument. `disposing` cannot indicate that we come from the native dtor because it already indicates whether we come from the finalizer or Dispose(). I also can't use `__ownsNativeInstance` to decide because when `__ownsNativeInstance` we must also free the native memory. 

There are a couple of other alternatives if we really don't want this parameter:
1. Stop the detour business. It only works for virtual dtor anyway. It is an incomplete solution.
2. Restructure the dtorHook like this:
```
throw if __ownsNativeInstance == true
temp =  __Instance
Call Dispose(true) // shouldn't call dtor() because __ownsNativeInstance == false
Call temp.dtor()
```

szilvaa:
@ddobrev I had some time over the weekend to figure out why a test crashed with my changes and I fixed it.

It turns out that there are 2 scenarios when CppSharp wants direct control over native dtor call:
1. when it detours the dtor in the native vtable. It needs to call dtor even if the managed wrapper does not own the underlying native instance.
2. when it passes objects by value. It needs to avoid calling the dtor even when the wrapper owns the underlying native instance.

So Dispose() needs 2 bool parameters:
The first `disposing` allows us to tell IDispose.Dipose() and Object.Finalize() apart.
The second `callNativeDtor` allows us to directly control the native dtor call.

Prior to this PR the code base used `disposing` for both needs. This was incorrect.

tritao:
Looks good, thanks for the fixes.

PR #1650, opened 2022-02-16
Fix NotImplementedException for signed char

AhmedZero:
Fix signed char Exception


tritao:
Thanks for the contribution.

PR #1670, opened 2022-07-26
RenamePass: add option to include/exclude namespaces from renaming

fabioanderegg:
I am really not sure why this wasn't an option before, if it's simply and oversight or if this change might have more implications.

A first CI run will probably give some answers.


src/Generator/Passes/RenamePass.cs
```diff
@@ -422,7 +425,7 @@ public static string ConvertCaseString(Declaration decl, RenameCasePattern patte
                     break;
             }
 
-            return sb.ToString();
+            return CSharpSources.SafeIdentifier(sb.ToString());
```
tritao, line 428: Unfortunately this code is C# specific, while this pass should be language-agnostic.

What ideally would happen is that we would have a safe identifier generic function to call here, but it seems like right now we only have some code to deal with this in `CheckKeywordNamesPass`: https://github.com/mono/CppSharp/blob/main/src/Generator/Passes/CheckKeywordNamesPass.cs#L28

Anyway, that pass runs unconditionally after user passes, as you can see here: https://github.com/mono/CppSharp/blob/main/src/Generator/Driver.cs#L298

So either we should rework the code and call a language-specific safe identifier method here or just remove this code and let the pass deal with it.

tritao:
Also the Mac CI action is failing due to Actions deprecating macOS 10.5 which we were using.

I've since updated main branch to update it to macOS 11 but the PR won't get the fix until you rebase.

Not really necessary tho, as the other OSes should provide plenty testing anyway.

fabioanderegg:
@tritao I realized what I did wrong. I removed the existing RenamePass, and then just appended my own RenamePass at the end of the Passes list. Which then obviously runs after the CheckKeywordNamesPass.

Is there a "good" way to replace the existing RenamePass with a custom one? This is our current approach, which requires this pull request:
```c#
            int index = driver.Context.TranslationUnitPasses.Passes.FindIndex(pass => pass.GetType() == typeof(CaseRenamePass));
            int numRemoved = driver.Context.TranslationUnitPasses.Passes.RemoveAll(pass => pass.GetType() == typeof(CaseRenamePass));
            if (numRemoved != 1)
            {
                throw new Exception("Something went wrong with removing the existing rename pass. " +
                    $"Excepted to remove one CaseRenamePass, actual number of removed passes: {numRemoved}");
            }
            driver.Context.TranslationUnitPasses.Passes.InsertRange(index,
                new TranslationUnitPass[]{
                    new CaseRenamePass(RenameTargets.Any & ~RenameTargets.Parameter & ~RenameTargets.Property, RenameCasePattern.UpperCamelCase),
                    new CaseRenamePass(RenameTargets.Property, RenameCasePattern.LowerCamelCase)
                }
            );
```

PR #1678, opened 2022-09-05
Do not rename property names in GetterSetterToPropertyPass

fabioanderegg:
Currently, the GetterSetterToPropertyPass will lower-case a property name unconditionally, e.g. `int GetABCTest();` gets converted to `aBCTest`. The default RenamePass will uppercase the first letter again, and the property is correctly named `ABCTest`. 

We have added a custom RenamePass to our project that expects to work on the original name of properties, e.g. `ABCTest` from the example above, so we can differentiate properties that start with a lower or uppercase letter.

This PR _should_ not change any behavior for people using the default RenamePass setup. We tested the change on our codebase with the custom RenamePass disabled and this change does not seem to make a difference.

Maybe there are other ways to test if this does not break anything apart from the tests?


fabioanderegg:
This breaks a ton of stuff, so I found a workaround on our side and we no longer need this.

PR #1695, opened 2022-11-18
Fixed memory copy error of virtual table under multiple inheritance

hxbb00:


josetr:
Can you add a test please?

hxbb00:
I have passed the test in the project of my company. Sorry, the project environment of my company is too big, so it is not suitable to provide the test environment. My company used this in the project, I use it in practice, so it shouldn't be a problem!

hxbb00:
image
this is mycase,here the __Internal:
image

hxbb00:
image
The difference with this modification is that there is no problem when the offset is 0, but when the offset is not 0, the address is calculated incorrectly

PR #1701, opened 2022-11-26
Incorrect build instructions on Windows

aybe:
The build instructions are just plain wrong for Windows.

Here's something that works on Windows 10 + Visual Studio 2022.


aybe:
You welcome.

PR #1741, opened 2023-07-09
CSharpExpressionPrinter: Wrap expression in parenthesis

trungnt2910:
Wrap default parameter expressions in parentheses to ensure whole expressions are casted instead of only the first operand.

This fixes the case when the C++ source contains `enum` flags `or`ed with each other as default parameters.

For example, [this code](https://github.com/haiku/haiku/blob/9aa43979f8a972fb20ac16e88d230f4ccd3401ae/headers/os/app/Roster.h#L93C1-L96C36
)
```cpp
			status_t			StartWatching(BMessenger target,
									uint32 eventMask
										= B_REQUEST_LAUNCHED
											| B_REQUEST_QUIT) const;
```

gets incorrectly translated to:
```cs
        public int StartWatching(global::Haiku.App.BMessenger target, uint eventMask = (uint) global::Haiku.App.WatchingRequestFlags.B_REQUEST_LAUNCHED | global::Haiku.App.WatchingRequestFlags.B_REQUEST_QUIT)
```

which causes an error:
```
/home/trung/cppsharp/generated/App.cs(9986,88): error CS0019: Operator '|' cannot be applied to operands of type 'uint' and 'WatchingRequestFlags' [/home/trung/cppsharp/HaikuApiGenerator.csproj]
```

Wrapping with a pair of parentheses should fix this issue:
```cs
        public int StartWatching(global::Haiku.App.BMessenger target, uint eventMask = (uint) (global::Haiku.App.WatchingRequestFlags.B_REQUEST_LAUNCHED | global::Haiku.App.WatchingRequestFlags.B_REQUEST_QUIT))
```


trungnt2910:
> Can you add a test to https://github.com/mono/CppSharp/blob/main/tests/dotnet/Common/Common.h?

Done. Also added a corresponding function in `Common.cpp`.

PR #1745, opened 2023-07-12
CSharpExpressionPrinter: Recurse into operands

trungnt2910:
Recursively call `VisitExpression` on the LHS and RHS of a binary operator expression. This fixes the generation for complex default parameters involving things other than two enumeration members.


trungnt2910:
Seems like C# does not like `static` properties as default parameters, so I converted them to `const` instead in the second commit.

PR #1747, opened 2023-07-13
CSharp: More default parameter fixes

trungnt2910:
- Expression generation for `ConstructorReference` now also recursively calls `VisitExpression` for the argument if only one argument is detected. This allows correct overload generation for functions taking a variable as the default parameter value.
- Default parameters of pointer-to-enumeration types are now correctly generated similar to primitive types.


src/Generator/Generators/CSharp/CSharpExpressionPrinter.cs
```diff
@@ -68,8 +68,8 @@ public string VisitExpression(ExpressionObsolete expr)
                 case StatementClass.ConstructorReference:
                     var constructorExpr = (CXXConstructExprObsolete)expr;
                     if (constructorExpr.Arguments.Count == 1 &&
-                        constructorExpr.Arguments[0].Declaration is Enumeration.Item)
-                        return constructorExpr.Arguments[0].Declaration.Visit(typePrinter).Type;
+                        constructorExpr.Arguments[0].Class != StatementClass.Any)
```
trungnt2910, line 71: I am blocking `StatementClass.Any` here since such statements will be generated without any kind of modification and may contain code that is invalid in C#.

Otherwise, `defaultMappedToEnumAssignedWithCtor` fails because `nullptr` is not valid C#.

It would be better if CppSharp is aware that `Arguments[0]` has been defaulted away and instead report 0 arguments to avoid this mess altogether.

trungnt2910:
```
D:\a\CppSharp\CppSharp\tests\dotnet\Common\Common.Tests.cs(1153,16): error CS7036: There is no argument given that corresponds to the required formal parameter 'defaultParam' of 'Common.FunctionWithFlagsAsDefaultParameter(int)' [D:\a\CppSharp\CppSharp\tests\dotnet\Common\Common.Tests.CLI.csproj]
D:\a\CppSharp\CppSharp\tests\dotnet\Common\Common.Tests.cs(1154,16): error CS7036: There is no argument given that corresponds to the required formal parameter 'defaultParam' of 'Common.FunctionWithConstFlagsAsDefaultParameter(int)' [D:\a\CppSharp\CppSharp\tests\dotnet\Common\Common.Tests.CLI.csproj]
D:\a\CppSharp\CppSharp\tests\dotnet\Common\Common.Tests.cs(1155,16): error CS7036: There is no argument given that corresponds to the required formal parameter 'defaultParam1' of 'Common.FunctionWithPointerToEnumAsDefaultParameter(ref Enum, ref int)' [D:\a\CppSharp\CppSharp\tests\dotnet\Common\Common.Tests.CLI.csproj]
D:\a\CppSharp\CppSharp\tests\dotnet\Common\Common.Tests.cs(1157,28): error CS7036: There is no argument given that corresponds to the required formal parameter 'defaultParam' of 'Common.FunctionWithNonPrimitiveAsDefaultParameter(Bar)' [D:\a\CppSharp\CppSharp\tests\dotnet\Common\Common.Tests.CLI.csproj]
D:\a\CppSharp\CppSharp\tests\dotnet\Common\Common.Tests.cs(1158,81): error CS0117: 'Common' does not contain a definition for 'DefaultBar' [D:\a\CppSharp\CppSharp\tests\dotnet\Common\Common.Tests.CLI.csproj]
```

Somehow CI on Windows fails when I call these functions because CppSharp does not generate the default parameters on Windows?

Somehow I cannot build CppSharp on my machine; any way to retrieve the generated C# code?

trungnt2910:
Seems like C++/CLI does not generate any default parameters. I think I should move all of these tests to `CSharp` tests instead?

PR #1753, opened 2023-07-20
CSharpSources: Dereference pointer variables

trungnt2910:
Dereference pointers when generating getters for pointer variables. Otherwise, the managed instance would point to the pointer itself rather than the native instance's address.


trungnt2910:
Done. I also made sure that the added test fails without the fix.

PR #1764, opened 2023-09-24
Fix crash on Linux by `DllImport`ing correct version of libdl

Pixelstormer:
CppSharp currently `DllImport`s the `dlopen` and `dlsym` functions from the library `"dl"`. This is incorrect however, as described in https://github.com/dotnet/runtime/issues/53291#issuecomment-853037480. Due to the way that `DllImport` searches for libraries, described [here](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/native-library-loading), specifying the library name as just `"dl"` causes it to find `libdl.so`, which comes from the `libc6-dev` package, but not the correct version of the library, `libdl.so.2`. The reason this crashes is because `libc6-dev` is not guranteed to be installed everywhere, so on Linux installations where it isn't, such as mine, the `DllImport` fails and throws an exception.

This PR fixes the crash by naming the correct version of the library in the `DllImport`, which should always exist.

### DIFF
--- a/src/Runtime/SymbolResolver.cs
+++ b/src/Runtime/SymbolResolver.cs
@@ -118,10 +118,10 @@ namespace CppSharp
             return dlopen(path, RTLD_LAZY);
         }
 
-        [DllImport("dl", CharSet = CharSet.Ansi)]
+        [DllImport("dl.so.2", CharSet = CharSet.Ansi)]
         static extern IntPtr dlopen(string path, int flags);
 
-        [DllImport("dl", CharSet = CharSet.Ansi)]
+        [DllImport("dl.so.2", CharSet = CharSet.Ansi)]
         static extern IntPtr dlsym(IntPtr handle, string symbol);
 
         #endregion


Pixelstormer:
@dotnet-policy-service agree

Pixelstormer:
So, the problem is that on macOS, libraries use the `.dylib` or `.bundle` extensions, and not the `.so` extension like on Linux. Unfortunately, `DllImport` has no way of specifying multiple possible library names, and the recommended workaround of `NativeLibrary.SetDllImportResolver` isn't available on `netstandard`. I've pushed a fix that just chooses between Linux and macOS versions of the functions, similarly to how it was already choosing between Unix and Windows versions. This isn't the nicest fix, given that the functions are completely identical besides the name of the library they're imported from, but it does work.

On an unrelated note, I noticed that the `PlatformID` enum that was being used to distinguish between platforms apparently [shouldn't be used any more](https://github.com/dotnet/runtime/issues/18635), as amongst other things, it can't actually distinguish between macOS and Linux, instead grouping them both together as Unix. To solve this, I replaced it with the recommended alternative of `OSPlatform`. There's several other places in the codebase that also use `PlatformID`, but changing those are out of the scope of this PR.

PR #1772, opened 2023-10-18
Fix Value Type Out Parameters

Saalvage:
Currently, using value types as out parameters generates the following incorrect code:

Given
```c++
union MyCoolUnion
{
    int a;
    float b;
};

int DLL_API DoCoolThing(CS_OUT MyCoolUnion* myCoolUnion);
```
the following code is generated:
```cs
public static void DoCoolThing(out global::Test.MyCoolUnion myCoolUnion)
{
    myCoolUnion = new global::Test.MyCoolUnion();
    var ____arg0 = new global::Test.MyCoolUnion.__Internal();
    var __arg0 = new __IntPtr(&____arg0);
    __Internal.DoCoolThing(__arg0);
}
```

This does not use the out parameter's internal value, but a newly created one, leaving the out parameter unchanged.

Following this PR the following code is generated:
```cs
public static void DoCoolThing(out global::Test.MyCoolUnion myCoolUnion)
{
    myCoolUnion = new global::Test.MyCoolUnion();
    fixed (global::Test.MyCoolUnion.__Internal* ____arg0 = &myCoolUnion.__Instance)
    {
    var __arg0 = new __IntPtr(____arg0);
    testerB = new global::Test.MyCoolUnion();
    __Internal.DoCoolThing(__arg0);
    }
}
```

Considerations:
- [ ] The indentation is broken, the insides of the `fixed` statements are not being indented. I hope this is either 1. negligible or 2. easily fixable with greater understanding of the library.
- [ ] Instead of using `fixed` a local could be used and the result later copied into the out parameter. I assume this would be more expensive, as it'd require an extra copy of a potentially very large structure. (If this is a big concern, it should probably be benchmarked)
- [ ] The line `Context.Return.Write($"new {typePrinter.IntPtrType}(&{arg})");` is duplicated once, with only the `&` being omitted. I'm currently not seeing a clean solution, but there might be one.

### DIFF
--- a/src/Generator/Generators/CSharp/CSharpMarshal.cs
+++ b/src/Generator/Generators/CSharp/CSharpMarshal.cs
@@ -634,17 +634,20 @@ namespace CppSharp.Generators.CSharp
             {
                 if (Context.Parameter.Usage == ParameterUsage.Out)
                 {
-                    var qualifiedIdentifier = (@class.OriginalClass ?? @class).Visit(typePrinter);
-                    Context.Before.WriteLine("var {0} = new {1}.{2}();",
-                        arg, qualifiedIdentifier, Helpers.InternalStruct);
+                    Context.Before.WriteLine("fixed ({0}.{1}* {2} = &{3}.{4})",
+                        Context.Parameter.QualifiedType, Helpers.InternalStruct,
+                        arg, Context.Parameter.Name, Helpers.InstanceIdentifier);
+                    Context.Before.WriteOpenBraceAndIndent();
+                    Context.Return.Write($"new {typePrinter.IntPtrType}({arg})");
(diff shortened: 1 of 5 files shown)


Saalvage:
Thank you for the quick merge! Are there any estimates on when there'll be another nuget release? I would like to use this fix in a project of mine but would want to avoid having to deal with git submodule or nuget shenanigans.

Saalvage:
Thank you very much!!

Saalvage:
Hmmm, it doesn't seem to be working for me. The `CppSharp` namespace appears to just be missing with 1.0.76.8341, changing it to 1.0.54.626 makes it work as normal again.
Might be nuget taking a while to correctly serve the package??

PR #1776, opened 2023-10-18
Fix #1251 three parameter equality operator

### LINKED ISSUE #1251: CppSharp is generating a three parameter equality operator
##### Brief Description
Just started working with cppsharp so I apologize if i missed some setup. I am trying to wrap a smaller c++ codebase and on my first tries cppsharp is generating an equality operator that takes three parameters (which fails compilation), the Internal call that gets executed also has three parameters (incorrectly), the class wrapped is an a simple implementation of `std::optional`. This happens for currently to template types, std::string and uint, and also for the inqueality operator !=. 
Thanks in advance for any help on this.

OS: Windows

##### Used headers
exception, type_traits (directly from the included file)

##### Used settings
```
options.GenerateClassTemplates = true;
options.GenerateDebugOutput = true;
```

Target: MSVC

Other settings 
##### C++ partial source
```
template <class T>
class Optional
{
public:
        ...
	inline bool operator==(const Optional<T>& rhs) const
	{
		return (m_hasValue == rhs.m_hasValue && (!m_hasValue || m_value == rhs.m_value));
	}
	inline bool operator==(const T& rhs) const
	{
		return (m_hasValue && m_value == rhs);
	}
       ...
}
```

```
// DEBUG: inline bool operator!=(const T& rhs) const
public static bool operator !=(this global::Msta.Msta.Runtime.Optional<string> @this, global::Msta.Msta.Runtime.Optional<string> __op, string rhs)
{
    bool @thisNull = ReferenceEquals(@this, null);
    bool __opNull = ReferenceEquals(__op, null);
    if (@thisNull || __opNull)
        return !(@thisNull && __opNull);
    var __arg0 = ReferenceEquals(@this, null) ? global::System.IntPtr.Zero : @this.__Instance;
    var __arg1 = __op.__Instance;
    var __basicString2 = new global::Std.BasicString<sbyte, global::Std.CharTraits<sbyte>, global::Std.Allocator<sbyte>>();
    global::Std.BasicStringExtensions.Assign(__basicString2, rhs);
    var __arg2 = __basicString2.__Instance;
    var __ret = __Internal.OperatorExclaimEqual(__arg0, __arg0, __arg2);
    __basicString2.Dispose();
    return __ret;
}
```

Saalvage:
- Operators in generic classes do not attempt to generate as extension methods anymore
- Empty `...Extensions` classes are no longer generated
- `string` as a template argument is correctly cast
- `MarshalCharAsManagedChar` option also generates correct casts
- Suppress warning regarding returning struct field by ref
- Eliminate some tabs that snuck into the test C++ header

Considerations:
- Is there a better way to force template instantation than using them in a method? I tried explicit template instantiation, but it didn't seem to have an effect (new feature/bug?)
- I'm not entirely sure if the `IsTemplate` extension method is entirely correct or if the same functionality might exist someplace else.

### DIFF
--- a/src/AST/TypeExtensions.cs
+++ b/src/AST/TypeExtensions.cs
@@ -439,6 +439,22 @@
             return false;
         }
 
+        public static bool IsTemplate(this Type type)
+        {
+            if (type is TemplateParameterType or TemplateParameterSubstitutionType)
+                return true;
+
+            var ptr = type;
+            while (ptr is PointerType pType)
+            {
+                ptr = pType.Pointee;
+                if (ptr is TemplateParameterType or TemplateParameterSubstitutionType)
+                    return true;
+            }
+
+            return false;
+        }
+
         public static Module GetModule(this Type type)
         {
             Declaration declaration;
--- a/src/Generator/Generators/CSharp/CSharpMarshal.cs
+++ b/src/Generator/Generators/CSharp/CSharpMarshal.cs
@@ -601,7 +601,7 @@ namespace CppSharp.Generators.CSharp
                 if (Context.Context.Options.MarshalCharAsManagedChar &&
                     primitive == PrimitiveType.Char)
                 {
-                    Context.Return.Write($"({typePrinter.PrintNative(pointer)})");
+                    Context.Return.StringBuilder.Insert(0, $"({typePrinter.PrintNative(pointer)}) ");
                     if (isConst)
                         Context.Return.Write("&");
                     Context.Return.Write(param.Name);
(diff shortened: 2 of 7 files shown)


Saalvage:
> Do you mean force the symbols to get exported in the shared library? If so, you can see how we do it in https://github.com/mono/CppSharp/blob/main/src/CppParser/Bindings/CSharp/i686-apple-darwin12.4.0/Std-symbols.cpp. There are some other techniques we use for forcing constructors to get exported for instance.

I don't think that is what I mean. If I try to do explicit template instantiation (`template class Optional<unsigned int>;`) CppSharp does not generate any C# code for the type. However, explicit template instantiations marked with `DLL_API` are included in the shared library from the C++ side. So it seems to be a case of CppSharp not accurately reflecting the shared libraries contents from what I can tell. If my understanding is correct in order to fix this it just needs to handle explicit template instantiations in the same way parameters are handled to generate the corresponding template types in the C# code.

Either way, if this turns out to be a bug/missing feature I think it's preferable to just open an issue and deal with it later, the method should be fine for now.


> Just wondering about the name, should it be IsTemplateParameterType instead?

Yep! I wasn't happy with the name either, this fits much better.


> You may be able to simplify it further with GetFinalPointee.

Very helpful! The code is much cleaner now!

PR #1783, opened 2023-10-20
Fix parameterless constructors not being generated for structs

### LINKED ISSUE #1730: std::string inside of value types does not generate correct bindings for C#
Hello, I have a C++ class that I want to expose to C# as a value type using `CS_VALUE_TYPE` macro. 

However, I have noticed that if my C++ class contains `std::string` or `const char*` member, the generated code is not correct and cannot build. For example, consider the following C++ code:

```C++
CS_VALUE_TYPE class EXPORT_MACRO value_type {
public:
  std::string string_member;
  const char *char_ptr_member;
};
```

It generates `public unsafe partial struct ValueType` C# structure with the following code for member variables:

```C#
public string StringMember
{
    get
    {
        var __basicStringRet0 = global::Std.BasicString<sbyte, global::Std.CharTraits<sbyte>, global::Std.Allocator<sbyte>>.__CreateInstance(__instance.string_member);
        return global::Std.BasicStringExtensions.Data(__basicStringRet0);
    }

    set
    {
        global::Std.BasicStringExtensions.__Internal.Assign(new __IntPtr(&__instance.string_member), value);
    }
}

public string CharPtrMember
{
    get
    {
        return CppSharp.Runtime.MarshalUtil.GetString(global::System.Text.Encoding.UTF8, __instance.char_ptr_member);
    }

    set
    {
        if (___instance.char_ptr_member_OwnsNativeMemory)
            Marshal.FreeHGlobal(__instance.char_ptr_member);
        ___instance.char_ptr_member_OwnsNativeMemory = true;
        if (value == null)
        {
            __instance.char_ptr_member = global::System.IntPtr.Zero;
            return;
        }
        var __bytes0 = global::System.Text.Encoding.UTF8.GetBytes(value);
        var __bytePtr0 = Marshal.AllocHGlobal(__bytes0.Length + 1);
        Marshal.Copy(__bytes0, 0, __bytePtr0, __bytes0.Length);
        Marshal.WriteByte(__bytePtr0 + __bytes0.Length, 0);
        __instance.char_ptr_member = (__IntPtr) __bytePtr0;
    }
}
```

`StringMember` setter fails with "[CS0212](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0212?f1url=%3FappId%3Droslyn%26k%3Dk(CS0212)) You can only take the address of an unfixed expression inside of a fixed statement initializer" error message, while `CharPtrMember` setter fails with "[CS0212](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0103?f1url=%3FappId%3Droslyn%26k%3Dk(CS0103)) The name '___instance' does not exist in the current context".

Do you know what could be the issue and how to fix it?

P.S.
The generated code works correctly if `CS_VALUE_TYPE` macro is omitted, but unfortunately, I need to use C# structs. 
Here is an example that works:

```C#
public string StringMember
{
    get
    {
        var __basicStringRet0 = global::Std.BasicString<sbyte, global::Std.CharTraits<sbyte>, global::Std.Allocator<sbyte>>.__CreateInstance(new __IntPtr(&((__Internal*)__Instance)->string_member));
        return global::Std.BasicStringExtensions.Data(__basicStringRet0);
    }

    set
    {
        global::Std.BasicStringExtensions.__Internal.Assign(new __IntPtr(&((__Internal*)__Instance)->string_member), value);
    }
}

public string CharPtrMember
{
    get
    {
        return CppSharp.Runtime.MarshalUtil.GetString(global::System.Text.Encoding.UTF8, ((__Internal*)__Instance)->char_ptr_member);
    }

    set
    {
        if (__char_ptr_member_OwnsNativeMemory)
            Marshal.FreeHGlobal(((__Internal*)__Instance)->char_ptr_member);
        __char_ptr_member_OwnsNativeMemory = true;
        if (value == null)
        {
            ((__Internal*)__Instance)->char_ptr_member = global::System.IntPtr.Zero;
            return;
        }
        var __bytes0 = global::System.Text.Encoding.UTF8.GetBytes(value);
        var __bytePtr0 = Marshal.AllocHGlobal(__bytes0.Length + 1);
        Marshal.Copy(__bytes0, 0, __bytePtr0, __bytes0.Length);
        Marshal.WriteByte(__bytePtr0 + __bytes0.Length, 0);
        ((__Internal*)__Instance)->char_ptr_member = (__IntPtr) __bytePtr0;
    }
}
```

### LINKED ISSUE #1777: Value types do not generate a default constructor
The following code:
```cpp
CS_VALUE_TYPE class DLL_API value_type {
public:
    std::string string_member;
};
```
does not generate a default constructor on the C# side of things, meaning it is impossible to initialize the object correctly and causing uninitialized memory reads after a write.

See #1730 for usecase.

Saalvage:
Fixes #1777, fixes #1730 (the latter for good this time).

The only remaining consideration is that this only considers the "happy" case, you can still get value objects in an invalid state via various mechanisms such as `default`. I don't think this is particularly high priority and a fix would likely be rather tedious, which is why I think I'd just leave that as a new open issue for now and link it in the corresponding `Ignored` test.


src/Generator/Generators/CLI/CLIHeaders.cs
```diff
@@ -798,7 +798,8 @@ public void GenerateFunction(Function function)
         public static bool FunctionIgnored(Function function)
         {
             return TypeIgnored(function.ReturnType.Type) ||
-                function.Parameters.Any(param => TypeIgnored(param.Type));
+                function.Parameters.Any(param => TypeIgnored(param.Type)) ||
+                function is Method { IsConstructor: true, Parameters: { Count: 0 }, Namespace: Class { IsValueType: true } };
```
Saalvage, line 802: This is required, because CLI does not support parameterless ctors on value types, so it might remain unsupported over in CLI-land?

Saalvage:
Actually.. there seems to be a memory leak when using `char*` in value types!
```cs
while (true)
{
    var test = new CSharp.ValueTypeNoCtor();
    test.CharPtrMember = "Loooooooooooooooooong string";
}
```
This leaks memory (even if the string member is removed, it's not at fault here).

Saalvage:
Hmm, this seems like a tough one.

I see three options in presumed increasing level of complexity:
1. Disabling `char*` to `string` marshalling for value type members
2. Using `IDisposable`
3. Wrapping the unmanaged memory in a managed class that cleans up the memory when it gets finalized and having that class be an additional member of the value type.

Saalvage:
From my end it's ready to go now, I just wanted to make sure to link the correct issue in the tests.

PR #1800, opened 2023-11-18
Build `Debug`/`Release` for `Windows`.

deadlocklogic:


deadlocklogic:
Personally I never encounter 32 bit architecture in my daily usage, so I don't mind using 64 bit exclusively.
Now in my experience I notice that the `Release` builds have better performance than `RelWithDebInfo`, but not that much thought.
Does it bother to build these artifacts anyway, as it is just a flag passed to the build system.

deadlocklogic:
Looks good, I will try native debugging on windows and see what goes on.

deadlocklogic:
@tritao I tried building in Debug mode, now tests fail while building with: `Assertion failed!: Could not get base CXX record from type`.
I wonder why it so difficult to use Debug on Windows, basically how are you managing to develop and debug the project?
There must be something messed-up going on:
```C++
static clang::CXXRecordDecl* GetCXXRecordDeclFromBaseType(const clang::QualType& Ty)
{
    using namespace clang;

    if (auto RT = Ty->getAs<clang::RecordType>())
        return dyn_cast<clang::CXXRecordDecl>(RT->getDecl());
    else if (auto TST = Ty->getAs<clang::TemplateSpecializationType>())
        return GetCXXRecordDeclFromTemplateName(TST->getTemplateName());
    else if (auto Injected = Ty->getAs<clang::InjectedClassNameType>())
        return Injected->getDecl();

    assert(0 && "Could not get base CXX record from type");
    return nullptr;
}
```
Asserting and then returning `nullptr`???
Then:
```C++
auto CXXBase = GetCXXRecordDeclFromBaseType(Base.getType());
if (!CXXBase || !HasLayout(CXXBase))
    return false;
```
We shouldn't even test for `nullptr` as the assertion guarantees that the result is non null. So right here the behavior in release mode while looking correct, is just a side effect!

PR #1803, opened 2023-12-07
Fixed a bug causing the header file to be included twice

zylalx1:
bug:
in C and CLI generator
when customizing the generated file name using the Options.GenerateName
the header file will be included twice in the generated .cpp file

reproduce:
add this line to the Setup(Driver driver) function in the CLI.Gen.cs
```
driver.Options.GenerateName = file => file.FileNameWithoutExtension + "_wrapper";
```
the output file CLI_wrapper.cpp will look like this
```
...
#include "CLI_wrapper.h"
#include "CLI_wrapper.h"
...
```

### DIFF
--- a/src/Generator/Generators/C/CppHeaders.cs
+++ b/src/Generator/Generators/C/CppHeaders.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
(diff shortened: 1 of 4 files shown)


zylalx1:
@dotnet-policy-service agree

PR #1808, opened 2023-12-14
Registrable generators: add support for lua, python

deadlocklogic:
This is PR will add support for registrable generators: lua, python etc..., basically any language which has a registration API.


deadlocklogic:
I have to refactor the last part which interferes with this PR: the passes.
Right now some passes have undesired effects, like templates, bitfields are automatically excluded from generation, which shouldn't be generalized.

deadlocklogic:
I am not finished yet though, this is just is bare minimal implementation, still many concepts missing: templates, protected base classes, typemaps to name a few.
Basically from my experience, there are 3 major techniques to bind with C++:
#### Structural:
Just like the C# generator, it is C++-like code structure for binding.
**Pseudocode:**
```C++
class Type {
    Type();
    void method();
}
```
#### Registrable:
Like sol, pybind, luabind, and any C++ helper library built on top of a C style API
**Pseudocode:**
```C++
Scope {
type<Type>
.constructor<Type>()
.method("method", &method);
}
```
#### Functional:
Here bindings will heavily rely on creating a separate function for each binding entity (just like with pure C API):
**Pseudocode:**
```C++
void wrapper_type_constructor(state s) {
// ...
}
void wrapper_type_method(state s) {
// ...
}
// And then we register all the functions to the state.
```

So I thought creating a base generator with these 3 concepts will ease the extension for newer one.

deadlocklogic:
@tritao Honestly I did quite a lot of work on Lua and other languages.
While I really liked this project in the beginning, I discovered that it has the right idea but not the optimal setup.
I believe the project should've been written in C++ based on the clang API directly.
While being written in C# with an independent AST is theoretically great, but practically slow and memory hungry for big libraries.
I have been working on an alternative solution in the past months, which is more similar to SWIG, and which at some point will be open source.
I can submit my work, which is basic and covers much of the use cases (because the other generators of this project aren't fully implemented too).
But for further development I am following my approach, which is incredibly fast compared to the current approach.

PR #1809, opened 2023-12-16
Implement ClassTemplatePartialSpecialization::Parameters

### LINKED ISSUE #1801: `ClassTemplateSpecialization` missing `Parameters` property
Consider:
```C++
template<typename T1, typename T2>
struct Test {
};
template<typename T1>
struct Test<T1, int> {
};
```
Now How to enumerate the template `Parameters` of the `ClassTemplateSpecialization` `Test<T1, int>`? (Should be `T1`).
This is particularly needed for class template partial specialization as function templates can't be partially specialized.

deadlocklogic:
Closes #1801

1. I only added `Parameters` field.
2. Extracted the `headers.zip` to the build directory
3. Rebuild the whole solution

Still no new bindings are being generated.

### DIFF
--- a/src/CppParser/AST.cpp
+++ b/src/CppParser/AST.cpp
@@ -793,6 +793,8 @@ ClassTemplatePartialSpecialization::ClassTemplatePartialSpecialization()
 
 ClassTemplatePartialSpecialization::~ClassTemplatePartialSpecialization() {}
 
+DEF_VECTOR(ClassTemplatePartialSpecialization, Declaration*, Parameters)
+
 FunctionTemplate::FunctionTemplate() : Template(DeclarationKind::FunctionTemplate) {}
 
 FunctionTemplate::~FunctionTemplate() {}
--- a/src/CppParser/Decl.h
+++ b/src/CppParser/Decl.h
@@ -715,6 +715,7 @@ namespace CppSharp
             public:
                 ClassTemplatePartialSpecialization();
                 ~ClassTemplatePartialSpecialization();
+                VECTOR(Declaration *, Parameters)
             };
 
             class CS_API FunctionTemplate : public Template


deadlocklogic:
@tritao Why `CppSharp.Parser.Gen` generated bindings are in `src\CppParser\Bindings\CSharp.Generators.GeneratorKind` not in `src\CppParser\Bindings\CSharp`?
Maybe this is a bug introduced after refactoring `GeneratorKind` from `enum` to a `class` right? in #1794

https://github.com/mono/CppSharp/blob/519e97f2da35bca21dc1de8cdd4598711da406c1/src/CppParser/ParserGen/ParserGen.cs#L77-L78
I will fix it now.😔

deadlocklogic:
Indeed it was a bug, I fix it with this commit, just implemented explicitly `GeneratorKind.ToString()`.

deadlocklogic:
Ok, now I can continue this PR, but unfortunately we have to revert #1802 so I can re-test it carefully. Because I was the whole time testing against same non regenerated bindings and now tests are failing. I believe the fix should be simple but lets safely revert and rework the PR.
This is so embarrassing, such a simple bug had devastating effects.

deadlocklogic:
Indeed this step is really worth keeping an eye on whenever modifying the `CLI/C#` backend. Maybe we could add a notice on new issues and add a [new PR template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) which could make new contributors aware of this.

deadlocklogic:
@tritao This PR is ready once you decide on: #1810
### Notice:
I had to push the newly generated bindings which introduced _parameterless_ constructors, because these changes weren't introduced by me and couldn't decide on them.
As you can see, tests fail because the C# 10.0 requirement.

deadlocklogic:
Ok will do now.

deadlocklogic:
@tritao Added a test anyway. Objectively speaking the tests need refactoring.
Asserting `Declaration` count is something, but assertions based on the line number of the parsed AST entity is something else.😂
We should move such line number based tests to a separate file.

PR #1810, opened 2023-12-16
Revert "Typemap cleanup"

deadlocklogic:
Reverts mono/CppSharp#1802

### DIFF
--- a/src/Generator/AST/Utils.cs
+++ b/src/Generator/AST/Utils.cs
@@ -207,7 +207,7 @@ namespace CppSharp.AST
                 return false;
 
             var typePrinterContext = new TypePrinterContext { Type = type };
-            var mappedTo = typeMap.SignatureType(typePrinterContext);
+            var mappedTo = typeMap.CSharpSignatureType(typePrinterContext);
             mappedTo = mappedTo.Desugar();
             mappedTo = (mappedTo.GetFinalPointee() ?? mappedTo).Desugar();
             return (mappedTo.IsPrimitiveType() ||
--- a/src/Generator/GeneratorKind.cs
+++ b/src/Generator/GeneratorKind.cs
@@ -38,11 +38,6 @@ namespace CppSharp.Generators
             Registered.Add(this);
         }
 
-        public static GeneratorKind FindGeneratorKindByID(string id)
-        {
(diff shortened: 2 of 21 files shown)


deadlocklogic:
@tritao Please read carefully.
While indeed the introduced bug caused unnoticed problems with #1802 (and which I know about now and will resolve separately). I wasn't really aware that we should run `CppSharp.Parser.Gen` in order to generate bindings for the project itself.
So after reverting this PR, I regenerated the project and run the tests, which failed to compile.
I then **reverted** all the PR created by myself, and again regenerated and tried to compile, with errors too.
Notice, my other PRs (other than this one trying to revert), had no changes on the diff.
So apparently, some older PR submitted by someone didn't know about running `CppSharp.Parser.Gen` to synchronize the project's bindings.
Now the good part is, most of the diffs in the generated bindings are just:
1. removed extra spacing for better syntax
2. generation of: `#pragma warning disable CS9084 // Struct member returns 'this' or other instance members by reference`
3. generation of a routine with `callNativeDtor`
4. some `__instance` are now marked as `ref`
5. `SourceLocation` has now a default constructor

Basically 5 is generating compilation errors:
`CS8773: Feature 'parameterless struct constructors' is not available in C# 9.0. Please use language version 10.0 or greater`

### Summary:
1. accept this PR which will revert #1802 which I will rework out, now I aware of project's binding regeneration.
2. check changes which were introduces by older PR which hasn't generated the project's bindings.

Sorry for the inconvenience. I made a mistake, but looks like others have fallen for the very same one.

deadlocklogic:
> Alternatively, we would need to require C# 10.0 going forward.

Indeed this is what I am thinking about. Practically speaking C# 10.0 is really new.
Lets tag @Saalvage and we see if he has an opinion here.

By the way I located the part which needs fixing in this PR, it is solely related to `Std.Map`, should be a very simple fix.

deadlocklogic:
As you like, I can close this PR then and submit a patch which fixes it, basically this typemap should be removed as it was in the first place in the CLI backend:
https://github.com/mono/CppSharp/blob/1327971e25ea298951f64864ad3bcdecf745ca0f/src/Generator/Types/Std/Stdlib.CSharp.cs#L415-L429
Commenting/removing this will fix the issue.

PR #1813, opened 2023-12-17
regenerated bindings

deadlocklogic:


deadlocklogic:
@tritao Tests will fail because of C# version 10.0 requirement.
I will modify build props and see what goes out.

deadlocklogic:
@tritao Tests will pass now.🎉
Still I want to create tests for `std::map` just to make sure everything is working properly.

Issue #1791, opened 2023-11-10
Error parsing [CXXDeductionGuide](https://en.cppreference.com/w/cpp/language/class_template_argument_deduction)


deadlocklogic:
### Consider:
```C++
using CppSharp;
using CppSharp.AST;
using CppSharp.Parser;

class Program
{
    class TestLibrary : ILibrary
    {
        public void SetupPasses(Driver driver)
        {
        }

        public void Preprocess(Driver driver, ASTContext ctx)
        {
        }

        public void Postprocess(Driver driver, ASTContext ctx)
        {
        }

        public void Setup(Driver driver)
        {
            DriverOptions driverOptions = driver.Options;
            ParserOptions parserOptions = driver.ParserOptions;

            parserOptions.LanguageVersion = LanguageVersion.CPP17;
            parserOptions.AddCompilationOptions("-frtti");

            driverOptions.Quiet = false;
            driverOptions.DryRun = true;
            driverOptions.OutputDir = Directory.GetCurrentDirectory();

            var module = driverOptions.AddModule("test");

            string temp = Path.GetTempFileName();
            File.WriteAllText(temp, @"
template<class T>
struct container
{
    container(T t) {}
 
    template<class Iter>
    container(Iter beg, Iter end);
};
 
// additional deduction guide
template<class Iter>
container(Iter b, Iter e) -> container<typename std::iterator_traits<Iter>::value_type>;
");

            module.Headers.Add(temp);

        }
    }

    static void Main(string[] args)
    {
        ConsoleDriver.Run(new TestLibrary());
    }
}
```
Output:
```
Parsing libraries...
Parsing code...
Unhandled declaration kind: CXXDeductionGuide
  C:\Users\user\AppData\Local\Temp\tmp6CC2.tmp (line 5)
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
Repeat 2 times:
--------------------------------
   at CppSharp.Parser.ClangParser+__Internal.ParseHeader(IntPtr)
--------------------------------
   at CppSharp.Parser.ClangParser.ParseHeader(CppSharp.Parser.CppParserOptions)
   at CppSharp.ClangParser.ParseSourceFiles(System.Collections.Generic.IEnumerable`1<System.String>, CppSharp.Parser.ParserOptions)
   at CppSharp.Driver.ParseCode()
   at CppSharp.ConsoleDriver.Run(CppSharp.ILibrary)
[...]

PR #1819, opened 2023-12-30
Parser.cpp: fix WalkType for clang::Type::TypeClass == clang::Type::DependentTemplateSpecialization

deadlocklogic:
Consider this snippet from the standard string library on msvc, without skipping private declarations:
```C#
class Test : ILibrary
{
    public void SetupPasses(Driver driver)
    {
    }

    public void Preprocess(Driver driver, ASTContext ctx)
    {
    }

    public void Postprocess(Driver driver, ASTContext ctx)
    {
    }

    {
        driver.ParserOptions.UnityBuild = false;
        driver.ParserOptions.SkipPrivateDeclarations = false;

        var options = driver.Options;
        options.Quiet = false;
        options.DryRun = true;
        options.Verbose = true;
        options.GenerationOutputMode = GenerationOutputMode.FilePerModule;
        options.GeneratorKind = GeneratorKind.CPlusPlus;
        options.UseHeaderDirectories = true;

        var module = driver.Options.AddModule("test");

        string temp = Path.GetTempFileName();
        File.WriteAllText(temp, @"
template <class _Ty>
class allocator {};
template <class _Alloc>
struct allocator_traits {};
template <class _Alloc, class _Value_type>
using _Rebind_alloc_t = typename allocator_traits<_Alloc>::template rebind_alloc<_Value_type>;
template <class _Elem, class _Alloc = allocator<_Elem>>
class basic_string { // null-terminated transparent array of elements
private:
using _Alty        = _Rebind_alloc_t<_Alloc, _Elem>;
};
");

        module.Headers.Add(temp);
    }
}
```
In Debug mode, `Parser::WalkType` assert fails: `assert(TL->getTypeLocClass() == TypeLoc::DependentTemplateSpecialization);`
The actual `TypeLoc` is: `TL->getTypeLocClass() == TypeLoc::TemplateSpecialization`.

There are also other problems when `TypeLocClass == TypeLoc::Qualified` and the inner `TypeLoc == TypeLoc::Elaborated`.

By the way, why the need of `TypeLoc` in the API at all? Haven't seen its usage.

### DIFF
--- a/src/CppParser/Parser.cpp
+++ b/src/CppParser/Parser.cpp
@@ -2681,10 +2681,11 @@ Type* Parser::WalkType(clang::QualType QualType, const clang::TypeLoc* TL,
         if (TS->isSugared())
             TST->desugared = GetQualifiedType(TS->getCanonicalTypeInternal(), TL);
 
-        TypeLoc UTL, ETL, ITL;
+        TemplateArgumentList TArgs(TemplateArgumentList::OnStack, TS->template_arguments());
 
         if (LocValid)
         {
+            TypeLoc UTL, ETL, ITL;
             auto TypeLocClass = TL->getTypeLocClass();
             if (TypeLocClass == TypeLoc::Qualified)
             {
@@ -2698,19 +2699,25 @@ Type* Parser::WalkType(clang::QualType QualType, const clang::TypeLoc* TL,
                 TL = &ITL;
             }
 
-            assert(TL->getTypeLocClass() == TypeLoc::DependentTemplateSpecialization);
+            if (TL->getTypeLocClass() == TypeLoc::DependentTemplateSpecialization)
+            {
+                DependentTemplateSpecializationTypeLoc TSpecTL = TL->getAs<DependentTemplateSpecializationTypeLoc>();
+                TST->Arguments = WalkTemplateArgumentList(&TArgs, &TSpecTL);
+            }
+            else if (TL->getTypeLocClass() == TypeLoc::TemplateSpecialization)
+            {
+                TemplateSpecializationTypeLoc TSpecTL = TL->getAs<TemplateSpecializationTypeLoc>();
+                TST->Arguments = WalkTemplateArgumentList(&TArgs, &TSpecTL);
+            }
+            else
+            {
+                llvm_unreachable("Unexpected semantics");
+            }
         }
-
-        DependentTemplateSpecializationTypeLoc TSpecTL;
-        DependentTemplateSpecializationTypeLoc *TSTL = 0;
-        if (LocValid)
+        else
         {
-            TSpecTL = TL->getAs<DependentTemplateSpecializationTypeLoc>();
-            TSTL = &TSpecTL;
+            TST->Arguments = WalkTemplateArgumentList(&TArgs, (DependentTemplateSpecializationTypeLoc*)nullptr);
         }
-
-        TemplateArgumentList TArgs(TemplateArgumentList::OnStack, TS->template_arguments());
-        TST->Arguments = WalkTemplateArgumentList(&TArgs, TSTL);
 
         Ty = TST;
         break;


deadlocklogic:
This PR partially fixes #1791, I am still in the process of debugging and fixing the general use case.

deadlocklogic:
Consider parsing:
```C++
template <class...>
struct conjunction {};

template <class B>
struct conjunction<B> : B {};
```
The assertion fails when calling `GetCXXRecordDeclFromBaseType`, we shouldn't assert this as `nullptr` is an expected value which we test against in the 2 referring calls.

deadlocklogic:
Tests in `Debug` mode before this PR had 4 crashing 1 failing, now down to 1 crashing 1 failing.

deadlocklogic:
> IIRC unless we used the  TypeLoc versions we got incomplete type information in some cases. It's worth a try removing it and seeing if all tests still pass, but even then, most of the edge cases only happened with complex bindings like QtSharp.

Interesting because most of `TypeLoc` and its derivates relay their properties from the type pointer.
I will try cleaning them up in a separate branch and see what goes on.

BTW I am having trouble parsing a relatively big project (without any processing, just parsing and AST generation), with around 650 header files, the process is crashing out of memory (consuming up to 15GB).
I wonder how compiling QT, which might be a larger library, was setup up.

deadlocklogic:
I am not using `UnityBuild` so far, and was really wondering about its advantages but still haven't tried it to come up with a conclusion.

deadlocklogic:
> Since it parses all the headers as a single translation unit, it can be significantly faster than parsing header by header, at the cost of a bit more memory usage.

I can't afford any more memory. I tried `-fsyntax-only` without any effect. Is there any techniques to avoid function body parsing, or at least to free their memory footprint from the AST?

This commit closes #1791, now will keep this PR active, to add tests and fix the one remaining bug in tests which causes crashing in Debug mode.

deadlocklogic:
Well I am trying to parse this game engine [rbfx](https://github.com/rbfx/rbfx), which gets parsed with clang json dump within few minutes, and relatively slow with `libclang`, but so far crashing because of out of memory with clang frontend (CppSharp).
I am an not even doing anything rather than just getting the sources to parse before any processing, but so far with no luck.

deadlocklogic:
I will try for sure.
While urho3d has a small community, it is friendly and willing to share experience with others.
I usually chat with the guys on discord, and was interested if I could integrate CppSharp with their build system to generate binding helpers for SWIG, because in its current form, this project still falls behind in terms of customization/advanced typemaps in comparison to the former.
Even thought, I have a kind of big idea in mind, if coupled with this project, could make something way better and accurate than SWIG. Still I need a little bit more of experimenting, getting the grounds ready for such a step.

deadlocklogic:
@tritao I tried adding a test, but because the C++17 requirement for deduction guides, a new bug is being introduced when adding C++17 compiler flag. The following tests are not ignoring the default argument in the C# generator:
```C++
void defaultNonEmptyCtor(QGenericArgument arg = QGenericArgument(0));
void defaultNonEmptyCtorWithNullPtr(QGenericArgument arg = QGenericArgument(nullptr));
```
```C#
// DEBUG: void defaultNonEmptyCtor(QGenericArgument arg = QGenericArgument(0))
public void DefaultNonEmptyCtor(global::CSharp.QGenericArgument arg = 0)
{
    var ____arg0 = arg.__Instance;
    var __arg0 = new __IntPtr(&____arg0);
    __Internal.DefaultNonEmptyCtor(__Instance, __arg0);
}

// DEBUG: void defaultNonEmptyCtorWithNullPtr(QGenericArgument arg = QGenericArgument(nullptr))
public void DefaultNonEmptyCtorWithNullPtr(global::CSharp.QGenericArgument arg = nullptr)
{
    var ____arg0 = arg.__Instance;
    var __arg0 = new __IntPtr(&____arg0);
    __Internal.DefaultNonEmptyCtorWithNullPtr(__Instance, __arg0);
}
```
Original valid results were:
```C#
// DEBUG: void defaultNonEmptyCtor(QGenericArgument arg = QGenericArgument(0))
public void DefaultNonEmptyCtor(global::CSharp.QGenericArgument arg)
{
    var ____arg0 = arg.__Instance;
    var __arg0 = new __IntPtr(&____arg0);
    __Internal.DefaultNonEmptyCtor(__Instance, __arg0);
}

// DEBUG: void defaultNonEmptyCtorWithNullPtr(QGenericArgument arg = QGenericArgument(nullptr))
public void DefaultNonEmptyCtorWithNullPtr(global::CSharp.QGenericArgument arg)
{
    var ____arg0 = arg.__Instance;
    var __arg0 = new __IntPtr(&____arg0);
    __Internal.DefaultNonEmptyCtorWithNullPtr(__Instance, __arg0);
}
```
I am not sure which pass is responsible for such modifications, so I can debug further.

deadlocklogic:
Is `HandleDefaultParamValuesPass` broken then?
Because as far as I know, the pass role is to find functions with default values then create overloads and delete the default value (in my own library I had something similar).
So, I would expected something like that generated for `MethodsWithDefaultValues::defaultNonEmptyCtor()`:
```C#
public void DefaultNonEmptyCtor(global::CSharp.QGenericArgument arg)
{
    var ____arg0 = arg.__Instance;
    var __arg0 = new __IntPtr(&____arg0);
    __Internal.DefaultNonEmptyCtor(__Instance, __arg0);
}
public void DefaultNonEmptyCtor()
{
    var ____arg0 = arg.__Instance;
    __Internal.DefaultNonEmptyCtor(__Instance);
}
```

deadlocklogic:
After further debugging, I found out that `C++17` might parse `CXXFunctionalCastExpr` as `CXXFunctionalCastExpr`, so even further modifications are needed.
So best option right now, is to:
1. improve the `Expr` support
2. consider preview SWIG options discussed before
3. merge this PR but keep the issue referencing it

deadlocklogic:
The solution that will solve the default arguments forever is to: 
1. modify `HandleDefaultParamValuesPass` so it generates all possible overloads for a function with default arguments, no need to check for default arguments expressions.
2. make HandleDefaultParamValuesPass a global pass and move it to the `Driver`
3. when the `Expr` is stable, we can introduce an option which keeps trivial default values and avoid extra overload generation. But overall we should always prefer correctness over advanced features.

Then we can close this PR and the related open issue, and later improve the `Expr` API.
If you want, I can work on these fixes so we can close this PR and add tests too.

deadlocklogic:
Again after a lot of digging, and been learning many new conventions along the way, fixing the whole problem with overload generation alone won't be that trivial as it will introduce new constraints which are hard to mitigate considering the current implementation.
Lets dissect the main obstacles and conclude what would be the best solution **for the time being**:
As I discussed before my obvious solution for default function arguments would be following [what SWIG does](https://www.swig.org/Doc4.2/SWIGPlus.html#SWIGPlus_default_args): generating an overloaded function for every combination.
The only serious drawback is that I don't know how to generate the mangled name for these synthetized functions with different possible ABI. Another minor issue is that we will lose the default argument hint when using intellisense, but nothing too serious.
Because SWIG uses global C-style functions for all the CLI, it doesn't suffer from this issue.
So to keep it simple **for the time being**, we have 3 solutions each one needs some concessions: (! = strength, ? = weakness)
1. migrate to a full `Expr` API and remove the obsolete part, and do the needed refactoring
   * best for the current implementation!
   * more corner cases to consider?
   * when typemaps API is more mature, marshalling problems might occur?
2. use global c-style functions **just for the overloaded methods**
   * relatively easy!
   * needs adopting some conventions like naming etc...?
3. use global c-style functions for all the CLI backend
   * following SWIG!
   * typemaps will be implemented way easier!
   * we can provide a separate CLI backend as option!
   * codebase far from being ready?

Unfortunately, our best bet will be on 1 currently.
Actually, we can create a milestone, for many things I've discussed previously and even in other tickets, just to keep them in mind before we lose note.

deadlocklogic:
@tritao, basically this PR is valid, but I am having hard time deciding how to fix `C++17` requirements. Going forward the obsolete `Expr` API would need refactoring into the newer API.

PR #1824, opened 2024-02-02
Enable ARM64 support on LLVM builds

konistehrad:
As suggested by  in https://github.com/mono/CppSharp/issues/1708?notification_referrer_id=NT_kwDOAB9097I1MzExMjcxNDYyOjIwNjE1NTk#issuecomment-1924243972 this PR adds support for building LLVM for ARM64 hosts, and adds ARM64 decompilation libs to the linkage.

Issue #1708, opened 2023-01-12
Building for mac m1(arm)


hellozyemlya:
Hello. I am trying to build CppSharp for mac m1, I use following commands
```
 1026  ./build.sh build_llvm --platform osx-arm64
 1027  ./build.sh package_llvm --platform osx-arm64
 1028  ./build.sh generate -configuration Release -platform osx-arm64
```
the last command failed with:
```
eugenechekanskiy@MacBook-Pro-Eugene build % ./build.sh generate -configuration Release -platform osx-arm64
Archive llvm-791523-macosx-osx-arm64-Release.tar.xz already exists.
Using debug LLVM build: /Users/eugenechekanskiy/Develop/CppSharp/build/llvm/llvm-791523-macosx-osx-arm64-Debug
Using release LLVM build: /Users/eugenechekanskiy/Develop/CppSharp/build/llvm/llvm-791523-macosx-osx-arm64-Release
Error: invalid value 'osx-arm64' for option 'arch'
```
What is the proper platform keys for each of building steps?

hxbb00:
Have you implemented the arm architecture support？i need support arm64, but i do not know how to generate bindings(CppSharp/src/CppParser/Bindings/CSharp/aarch64-linux-gnu-cxx11abi)，Can you give me some help

konistehrad:
Bumping this. It'd be a big help for Unity development, of all things, which often depends on bindings to native code.

konistehrad:
Bingo, thanks for that tip on the headers that's what I needed. (I was in the middle of exploring ParserGen as you were typing.) I can generate the bindings for you for M1 if it'll help.

konistehrad:
(Also for future travelers: the `osx-arm64` above should be replaced with `arm64`)

XzuluX:
Just tried to build CppSharp on Ubuntu aarch64

Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.3 LTS
Release:	22.04
Codename:	jammy

```
./build.sh build_llvm --platform arm64
./build.sh package_llvm --platform arm64
./build.sh generate -configuration Release -platform arm64
```
got following error, any idea ?

```==== Building Encodings.Native (release_arm64) ====
Creating ../../obj/Encodings.Native/Release
Encodings.cpp
Creating ../../gen/Encodings
Linking Encodings.Native
make: Leaving directory '/home/zulu/repos/CppSharp/build/gmake'
[...]

PR #1825, opened 2024-02-03
Arm64 bindings

konistehrad:
Depends on https://github.com/mono/CppSharp/pull/1824. Expands bindings to include ARM64 for Linux and Mac, and updates Mac headers to 12.3 to allow for ARM64 usage.

Uses the following updated header package:  [headers.zip](https://github.com/mono/CppSharp/files/14151916/headers.zip)


konistehrad:
Sorry, meant to open as draft. I made a pretty silly mistake in the MacOS headers that I'll have to correct tomorrow. Would love some eyes on the linux side in the meantime tho.

konistehrad:
Took a crack at resolving the errors this morning to no avail. Updated the headers package with a clean version. Might help to have someone more familiar with the project take a swing at resolving the segfaults on ARM64 Mac.

konistehrad:
So the state of play:

-I don't think an ARM64 host is building LLVM. I'm still having to build it locally to work.
-My Debian VM fails to complete the LLVM build, so I'm having a hard time testing ARM64 on Linux.
-Mac gets segmentation faults during the final build step. Looks like one of the bindings might be incorrect somehow?

konistehrad:
That's interesting, because attempting to use the generated `llvm-6eb36a-linux-arm64-gcc-11-Release` yields a bunch of these when attempting to build on Linux ARM64:

`/usr/bin/ld: skipping incompatible ../../llvm/llvm-6eb36a-linux-arm64-gcc-11-Release/build/lib/libLLVMDebugInfoPDB.a when searching for -lLLVMDebugInfoPDB
/usr/bin/ld: cannot find -lLLVMDebugInfoPDB: No such file or directory`

Archive files are a little harder to inspect than shared libs, but I'm used to seeing this error when the generated archive contains the wrong architecture.

konistehrad:
Gotcha. Let me cook up some cmake toolchain files to do the LLVM cross compile anyway. That’d save me a ton of time. I’ll make that in a different branch to expedite review and we can kick off a new LLVM job.

konistehrad:
Actually, while I have you: what's the correct way to determine host architecture in premake? I can't seem to find a way to test to see if we're executing on an x64 host for an arm64 target.

konistehrad:
Oh, yes, I did find `os.host()`, unfortunately it doesn't seem to return architecture info. For example, from an M1 Mac in interactive mode:

```
> print(os.host())
macosx
```

Not quite enough to do the dirty work I'm afraid. I guess I can use `uname` on this path since our cross compile paths are limited to Unix-like operating systems.

konistehrad:
OK, state of play:

- Rosetta context wasn't a good enough test bed for ARM64 cross compiling. I'll need help from someone with access to a real Intel Mac to smooth that out it looks like.
- I updated the Calling Convention from Cdecl to Winapi in `std.cs` for ARM Mac and it appears to have cleaned some stuff up. It's failing to find some libraries now, but it's progress.
- It looks like we need to add Arm64 to `src/CppParser/AST.h`, but I don't know yet where else this needs to end up. I'll be working on that as able.

konistehrad:
Sorry, I'm not able to continue working on this due to some major life stuff. I'll close it for now. If anyone else wants to pick up the effort just ping me and I'll try to help as able.

perlun:
For reference, trying to get this moving now here: https://github.com/mono/CppSharp/issues/1708#issuecomment-3866466584

PR #1826, opened 2024-02-06
LLVM crosscompile x64 for ARM64

konistehrad:
Linux x64 now properly builds ARM64 versions of LLVM. Unfortunately, I can't seem to power on my last Intel Mac, so until I find a system to test on I can't verify it. If the CI could run it in the meantime I'd much appreciate it.


konistehrad:
OK good news looks like the Mac cross compile path works as well. I've updated with the changes requested. The flag you mentioned didn't seem to be needed, but if it's there it's also not hurting anything so I'm inclined to leave it in case it's needed for a particular host configuration that I haven't tested.

konistehrad:
Just did. Yep, looks like you'll need to `apt install g++-aarch64-linux-gnu gcc-aarch64-linux-gnu` before the build process. Unfortunately I'm not familiar with GitHub actions, but I'm assuming there's an easy way to do environment configuration. Mac looks like it's cooking along nicely tho. I'll try the generated artifacts shortly.

PR #1827, opened 2024-02-06
Ensure cross compiler is in the environment before building

konistehrad:
Install Aarch64 GCC before running cross compile steps. Uses naive `apt-get` method as opposed to larger caching solution like [cache-apt-pkgs-action](https://github.com/awalsh128/cache-apt-pkgs-action), but let's ensure this works before making things more complicated.


konistehrad:
One step ahead of me. Please disregard. ;)

PR #1836, opened 2024-03-13
Don't add values to enums that can't be parsed

Saalvage:
Currently if one of the values of an enum can't be parsed it's still added to the enum on the C# side of things with a value of 0. This PR changes it so that values that can't be parsed are not added to the enum on the C# side.

Although this is theoretically a breaking change I cannot imagine circumstances in which the current behavior is preferable and only a slight change to how the invalid values are handled should be required for code that manually deals with the unparseable values.


Saalvage:
Stupid me!

But as it turns out this seems to have been intended behavior at some point in time, however I really struggle to see any justification for it.

Issue #1283, opened 2020-01-12
Overriding of pure virtual method is incorrect in C# wrapper


ArtsevDmitry:
I'm playing around OpenCascade simple wrapper. I use one header "Geom2d_Line.hxx", class Geom2d_Line inherited from Geom2d_Curve which contain next method:

```
  Standard_EXPORT virtual Standard_Boolean IsPeriodic() const = 0;
```
in Geom2d_Line class this method overrides:

``` //! Returns False
  Standard_EXPORT Standard_Boolean IsPeriodic() const Standard_OVERRIDE;
```
OS: Windows

when it converted to C#, I have property in Geom2dCurve
 ```
public abstract bool IsPeriodic
{
      get;
}

```
but in Geom2dLine I have these two members:

```
        /// <summary>Returns False</summary>
        public virtual bool IsPeriodic()
        {
            var __slot = *(void**) ((IntPtr) __OriginalVTables[0] + 13 * 8);
            var ___IsPeriodicDelegate = (global::UniCad.Delegates.Func_bool_IntPtr) Marshal.GetDelegateForFunctionPointer(new IntPtr(__slot), typeof(global::UniCad.Delegates.Func_bool_IntPtr));
            var __ret = ___IsPeriodicDelegate((__Instance + __PointerAdjustment));
            return __ret;
        }
```
and this in same CS class:

```
public override bool IsPeriodic { get; }
```

There are two issues:

1. We have property and method with same names
2. Property not overrided correctly (as a minimum should return IsPeriodic() method, but issue 1 in this case)

What should I do in this case, can I solve it with custom pass or something?
Thanks.

##### Used headers
Geom2d_Line.hxx

##### Used settings

Target: MSVC

tritao:
Thanks for reporting the bug.

Maybe we should not generate properties for methods, when they are virtual.

As a workaround, you can disable the pass that generates properties out of getters (`GetterSetterToPropertyPass`).

ddobrev:
@ArtsevDmitry the correct behaviour should be to only get an overridden property with the body of the generated method which in turn should be gone. Unfortunately, I couldn't reproduce it here. The macros in your test case might be somehow involved. I would advise you to try generation with gradually removed macros to see if you can get to an even simpler test case.

angryzor:
I ran into this bug as well and made a minimal reproduction:

```c#
#include <cstdint>

namespace foo {
class Dorld {
	virtual uint64_t UnkFunc1() = 0;
};
}

[...]

PR #1855, opened 2024-07-26
Fixes issue #1283

angryzor:
See #1283 .

`GetterSetterToPropertyPass` currently fails on the following code:

```c#
namespace foo {
class Dorld {
    virtual int UnkFunc1() = 0;
};
}

class Bar : public foo::Dorld {
public:
    virtual int UnkFunc1() override;
};
```

[Resulting output](https://gist.github.com/angryzor/5f00133280bce5e0cd523eb0d2562428)

The namespace is important. If `Bar` is declared inside the same namespace the bug is not triggered. However, my original code where I encountered this bug did have both classes in the same namespace, though they were in different header files.

The bug occurs when it tries to determine whether the property overrides a base property [here](https://github.com/mono/CppSharp/blob/3f923b1c640307fc19f0c83902c81bde1dc50b3a/src/Generator/Passes/GetterSetterToPropertyPass.cs#L215).

When `Bar` is outside the namespace, `Bar` gets processed before `Dorld`, so `Dorld` doesn't have the property created yet. The check fails and the property is not created on `Bar`. The property is then later created on `Dorld` when the pass finally visits `Dorld`.

When instead `Bar` is inside the namespace, `Dorld` is visited first and the code continues as expected.

Fixed by setting the `VisitFlags.ClassBases` flag on the pass.

### DIFF
--- a/src/Generator/Passes/GetterSetterToPropertyPass.cs
+++ b/src/Generator/Passes/GetterSetterToPropertyPass.cs
@@ -40,7 +40,7 @@ namespace CppSharp.Passes
         }
 
         public GetterSetterToPropertyPass()
-            => VisitOptions.ResetFlags(VisitFlags.ClassTemplateSpecializations);
+            => VisitOptions.ResetFlags(VisitFlags.ClassBases | VisitFlags.ClassTemplateSpecializations);
 
         public override bool VisitClassDecl(Class @class)
         {


angryzor:
@dotnet-policy-service agree

angryzor:
Agh, I had it there first, then saw there was a suite for the passes. Alright, I'll move it over. The strange thing is though, this bug prevents compilation on my machine, so I expected `TestUncompilableCode` to fail when I added the fixture, but that didn't happen.

angryzor:
> The mac CI is still not running, but since its already passing on other platforms, I will just merge it. Again, thanks for the fix.

No problem!

PR #1861, opened 2024-08-20
Couldn't remove modules without translation units

HenrikAkseliValve:
I do not know I'm first to encounter this, however, loop that removes modules without translation units, could not have worked.

In C# Enumerator prohibits editing the collection so you cannot remove modules from `options.Module` while iterating `options.Module`.
CppSharp_Enumeration_bug00
CppSharp_Enumeration_bug01

Chose `RemoveAll` for its simplicity. Not necessary optimal solution although it might be.


HenrikAkseliValve:
@dotnet-policy-service agree

HenrikAkseliValve:
Thank you. Fastest merge request I have ever requested. 👏

Issue #1687, opened 2022-10-06
Use LibraryImportAttribute instead of DllImportAttribute


jasoncouture:
##### Brief Description
When code generation is targeting .NET 7, `LibraryImport` should be used in place of `DllImport` as this will pre-generate marshalling operations normally performed at runtime.

When framework target is lower than .NET 7, `DllImport` should continue to be used.

(Non os specific, code generation)

OS: N/A

caihongxu:
+1. Having the option to generate LibraryImport would be really useful!

PR #1862, opened 2024-08-22
add: options for DllImport and LibraryImport

kumja1:
Resolves issue #1687


kumja1:
Note: their are still more changes to be made

kumja1:
@dotnet-policy-service agree

kumja1:
note: there are still more things to be changed

kumja1:
@tritao How would i regenerate the bindings for the cpp parser

kumja1:
@tritao primitive strings are typically utf16 right? or are they different depending on the platform(like wchar_t is utf32 on linux)

PR #1886, opened 2025-01-01
Parsing errors even with release.

HenrikAkseliValve:
Adds two macros to the CppParser/Parser.cpp for custom assert. They print out a message regardless is release or debug enabled. Abort is only called with debug.

Macro `assertm` just prints out the message if the condition is false. Macro `assertml` also prints out the given source location to indicate where parsing fails.

Macro `assertml` is not used often but might be good to be used more.

Also, I removed an assert that seems pointless as if check before it handles it already.


HenrikAkseliValve:
Dammit Ubuntu CI ***** up...

HenrikAkseliValve:
There were spelling mistakes on the release config.

HenrikAkseliValve:
Ubuntu had problems with `auto`.

PR #1887, opened 2025-01-02
Test CustomAllocator debugging configuration fix.

HenrikAkseliValve:
On Windows 2022 "vector" module deconstructor of the `vector` is the following:
```C++
    _CONSTEXPR20 ~vector() noexcept {
        _Tidy();
#if _ITERATOR_DEBUG_LEVEL != 0
        auto&& _Alproxy = _GET_PROXY_ALLOCATOR(_Alty, _Getal());
        _Delete_plain_internal(_Alproxy, _STD exchange(_Mypair._Myval2._Myproxy, nullptr));
#endif // _ITERATOR_DEBUG_LEVEL != 0
    }
```

When debugging configuration is enabled, internals here report an error for NamespacesDerived.Native.

A fix is to add Constructor and Copy Constructor to the NamespacesDerived.h `CustomAllocator`. Windows manual seems to want even more to be defined, but I leave it here. https://learn.microsoft.com/en-us/cpp/standard-library/allocators?view=msvc-170

### DIFF
--- a/tests/dotnet/NamespacesDerived/NamespacesDerived.h
+++ b/tests/dotnet/NamespacesDerived/NamespacesDerived.h
@@ -103,6 +103,10 @@ class CustomAllocator
 public:
     typedef T value_type;
 
+    // Suppress some windows Debug errors with constructors.
+    CustomAllocator() noexcept {};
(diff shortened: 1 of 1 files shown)


duckdoom5:
This fix was already merged with #. We can close this now :)

PR #1898, opened 2025-02-03
Fix crash on auto return types

duckdoom5:
Not sure if this is resolved properly


duckdoom5:
Yeah, I found another instance if this same bug occurring some time ago. See this line https://github.com/mono/CppSharp/blob/main/src%2FAST%2FASTVisitor.cs#L522

PR #1900, opened 2025-02-03
Fix incorrect tag generation for enum classes

duckdoom5:
Currently invalid source is generated for scoped enums (`enum class`).

This patch will resolve this issue by adding `class` when it is a scoped enum.

I've also added an early out to skip printing the tag is the class isn't incomplete, since that shouldn't be necessary in those cases. (Though maybe we should add this as an option so you'd still be able to force print it. Eg. new method `PrintTagIfNeeded`?)


duckdoom5:
Hm, interesting.. I guess it's not enough to know if it's incomplete or not. Might look into it a bit tomorrow or just remove the early out for now

duckdoom5:
Okay, so I've been looking into this and it seems relatively easy to fix. I just need to check if there is an ambiguous name within the current namespace. So, I've added that function.

I'm just running into an issue:
image

Can you tell me why the `stdlib.h` functions are not included in the `ASTContext`?

duckdoom5:
Ah yeah, that makes total sense. 
I'll convert that into an option and default it to skip like it is now. That should resolve this issue

duckdoom5:
@tritao So ran into a couple of issues again which I resolved today. I'll make more PR's for those in a bit.

Last issue I'm dealing with now, is that if I enable parsing of the system headers, the generated `Std-Symbols.cpp` file adds the following invalid declaration:
`template __declspec(dllexport) std::char_traits<char>& std::char_traits<char>::operator=(std::char_traits<char>&&);`

>1 error generated.
   \CppSharp\build\gen\Common\Std-symbols.cpp(7,80): error: explicit instantiation refers to member function 'std::char_traits<char>::operator=' that is not an instantiation
    C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.42.34433\include\__msvc_string_view.hpp(465,8): note: explicit instantiation refers here

duckdoom5:
Hm, yeah I noticed some other things as well, but I'll leave that discussion for later. Want to get this fixed first and maybe then ask some questions about how some things are designed currently cus I'm a little confused about translations units atm.

PR #1902, opened 2025-02-03
Add marshalling for `std::optional` <=> `System::Nullable`

duckdoom5:
Should be mapped, since that's an almost one to one conversion.

Might want to create a test for this one


duckdoom5:
Ehm, well, if you update the bootstrap source, you also need to update the parser bindings. (Including the C++/CLI ones)

duckdoom5:
Sorry, in that case I'm having a hard time understanding what you mean. I think I'm missing some context.

All I can say though, is that I simply followed the implementation for `std::vector`a few lines below. (Which is also not implemented in the `Stdlib.CSharp.cs`file)

PR #1906, opened 2025-02-04
Update llvm build workflows

duckdoom5:
- Enabled llvm debug builds for all os's
- Enabled debug builds in CI
- Windows now builds Release builds with the release version of llvm
- Added the output of natvis files


duckdoom5:
Note: Workflows fail because we need to trigger the llvm build workflows, but I can't run those myself :)

duckdoom5:
@tritao This should help as well https://github.com/mono/CppSharp/pull/1906/commits/6914876697ee825f2cfb622afc03c35b1d0c3053

duckdoom5:
> > @tritao This should help as well [6914876](https://github.com/mono/CppSharp/commit/6914876697ee825f2cfb622afc03c35b1d0c3053). I also wonder if we actually need all the libs that are currently included with the llvm package?
> 
> I think so, we arrived at the current packaging code by mostly adding stuff as needed, so I would be surprised if there's that much "fat" in there, but of course it's likely we could find some stuff.
> 
> However the main issue here is really the packaging step failed at: https://github.com/mono/CppSharp/actions/runs/13142034349/job/36682629705
> 
> We can't call `ninja -t clean` because that would remove the files we need to package as well, so we'll need some manual code that clean some folders with build object files, I'll see if I can come up with something later today.

Yeah, by removing the downloaded archive I just freed up another 300~ MB

duckdoom5:
(See edit in previous comment)

Lastly I wanted to ask if we should update clang to a new revision?

duckdoom5:
> > > @tritao This should help as well [6914876](https://github.com/mono/CppSharp/commit/6914876697ee825f2cfb622afc03c35b1d0c3053). I also wonder if we actually need all the libs that are currently included with the llvm package?
> > 
> > 
> > I think so, we arrived at the current packaging code by mostly adding stuff as needed, so I would be surprised if there's that much "fat" in there, but of course it's likely we could find some stuff.
> > However the main issue here is really the packaging step failed at: https://github.com/mono/CppSharp/actions/runs/13142034349/job/36682629705
> > We can't call `ninja -t clean` because that would remove the files we need to package as well, so we'll need some manual code that clean some folders with build object files, I'll see if I can come up with something later today.
> 
> Yeah, by removing the downloaded archive I just freed up another ~300MB
> 
> Also, I'm not sure why this happened.. But for some reason the windows llvm `RelWithDebugInfo` build was using `Release` in it's package name. I already checked the code again, but couldn't find anything wrong atm. Would you mind running that again to confirm if it's still an issue? (I might have already resolved it with another commit)

Hmm, maybe this function just never worked correctly:
```
function get_llvm_configuration_name(debug)
	if string.find(_OPTIONS["configuration"], "DebugOpt") then
		return os.istarget("windows") and "RelWithDebInfo" or "Release"
	end
	if string.find(_OPTIONS["configuration"], "Debug") then
		return "Debug"
	end
	return "Release"
end
```

Do you see anything wrong with that? I'm not that well versed in lua tbh. Can't we use a filter here as well?

duckdoom5:
> Not sure yet about that, but was looking at the changes in more detail, specifically:
> 
> ```
> build-cfg: [Debug, RelWithDebInfo, Release]
> ```
> 
> Why do we need both `Release` and `RelWithDebInfo` builds now?

We don't really need it, but I thought it might be nice to also have the debug builds available on mac/linux. But I can totally revert that if that's not needed

duckdoom5:
Oh, duh, I'm a dummy XD. Fix it now :P

duckdoom5:
> Ok, my bad, was actually looking at the wrong workflow, I thought you were adding two different release builds for CppSharp, but it's actually for LLVM.
> 
> Is there any reason to have both `Release` and `RelWithDebInfo` builds? I think previously we just used `RelWithDebInfo` right? So what are we gaining with `Release`? Is it just for the Nuget build as you said?

It's size and speed. The compiler flags used would include:
```
RelWithDebInfo: `-O2 -g -DNDEBUG`, 
Release: `-O3 -DNDEBUG`
```
To have a debuggable  build, the compiler needs to add additional information. (depending on setting that could be embedded in the exe/dll or be a pdb).

Also AFAIK the compiler also needs to insert some additional code for debugging purposes on occasion, but I'm not sure about that last bit.

duckdoom5:
> schannel: next InitializeSecurityContext failed: CRYPT_E_REVOCATION_OFFLINE (0x80092013) - The revocation function was unable to check revocation because the revocation server was offline.

Hm, it's doing that again. I wonder why..

duckdoom5:
> > > Is the idea that the new DebugOpt build is equivalent to the old RelWithDebInfo?
> > 
> > 
> > yes
> 
> Not sure about calling it `DebugOpt` in that case, and in essence it's still a release build, which even with `-g` debug enabled, it's not really that usable to debug, especially LLVM with all inlined call stacks going on. `RelWithDebInfo` is a bit more informative IMHO to what to expect.

I do agree, but what happened here is that the `DebugOpt` configuration which is used for CppSharp's build config, is automatically mapped to LLVM's `RelWithDebInfo` build.

The reason I did that is because when debugging CppSharp, we generally won't care about the library code and want that to run as fast as possible but still get some info if it happens to crash inside of the library (hence `RelWithDebInfo`).


That said, if you think the llvm build scripts should use the `RelWithDebInfo` mapping, I can update the scripts to map `RelWithDebInfo` if ran from llvm workflows and continue mapping `DebugOpt` to `RelWithDebInfo` when running CppSharp workflows

duckdoom5:
Awesome It was built successfully. Now we only need to rebuild llvm windows `DebugOpt` and we should be all greens on this PR as well

duckdoom5:
> or automatically publish them to releases.
I'm pretty sure that's possible. I'll have a look

duckdoom5:
I found this one. Seems to be simple to setup https://github.com/softprops/action-gh-release

duckdoom5:
Hmm.. It seems the working directory value is taken as a raw string.

Let's try without it

duckdoom5:
@tritao Did you try with workspace dir override removed?

duckdoom5:
Awesome they succeeded! So now we can do the final checks here

duckdoom5:
Yeah, here's the issue. I'll get that fixed now :)

> Using cached LLVM 'Debug' build: D:/a/CppSharp/CppSharp/build/llvm/llvm-6eb36a-windows-vs2022-x64-Debug
Using cached LLVM 'RelWithDebInfo' build: D:/a/CppSharp/CppSharp/build/llvm/llvm-6eb36a-windows-vs2022-x64-DebugOpt
Using cached LLVM 'Release' build: D:/a/CppSharp/CppSharp/build/llvm/llvm-6eb36a-windows-vs2022-x64-RelWithDebInfo

PR #1918, opened 2025-03-01
Draft bootstrap decl

duckdoom5:
Okay so I managed to generate a `Declaration.h` file from clang source, but I'm a bit stuck.

- Some properties from clang are missed due to parsing/converter issues
- Some fields should be removed since they can be mapped to simple functions
- The original decl types add a bunch of properties that no longer exist

Since they don't really match it's not easy to map the new data to the old one. I'm not sure how to go from here.. I kinda need this new data to properly support everything and implement the recursive AST visitor we talked about before.

I feel like the only way to get this done is to slowly convert each of the original decls to the new version, but I also don't want to waste time. Any suggestions/ideas to reduce the amount of time required for this task?


duckdoom5:
It's 3 things really:
- Add properties we are missing currently
    - With those added, we'd be able to use clang's implementation of the AST dumper, which means we'll have a complete AST
- Future proofing. C++23 is around the corner. If we have this generator automated, it should be a click of a button and maybe some fixes to add support for new languages
- Avoid implementation bugs.
    - Make sure the properties are actually set.
    While working on this, I found that a bunch of properties that are added on the c++ side are never assigned to (eg. `Class.IsStatic` is never assigned to, but it's used to check behavior a lot).

duckdoom5:
@tritao Was there a specific reason not to include OMP and OBJC expr/stmt nodes?

duckdoom5:
Cool, was just wondering, since it would be quite easy to just support them if they are auto generated anyway

duckdoom5:
Agreed

duckdoom5:
Yeah, I see your points.
I was a bit in the same boat, which is why I wanted your feedback.

> I'm just a bit worried from a future maintenance perspective

I also agree with this. I hope if I get more familiar with the remaining parts of the codebase I can help simplify this process. 
(Eg. I'm looking for ways to not have to manually specify what variables we want to keep from clang.)
There is some logic to it, like removing those properties that are simple wrappers around data.
Like here, where these really don't need to be data:
image
image
If we can read part of the expression and see that it's a simple local data wrapping function, it can just stay a function.
But there's more examples.

I also think we should go through the codebase and deduplicate some code. Make use of existing utility functions instead. For example the codebase makes use of these lines of code all over the place:
```
TypeMap typeMap = null;
if (TypeMapDatabase.FindTypeMap(tag, out typeMap))
{
    var typePrinterContext = new TypePrinterContext { Type = tag };
    return typeMap.SignatureType(typePrinterContext).ToString();
}
```
But there's also the helper function `GetMappedType` which does exactly the same thing.
Reducing the mental load for these functions should make it more maintainable.

Right now I'm taking a step back from this, and am focusing on the features that I need to generate source for my codebase. In the meantime, I'll try to improve/fix some of the CppSharp code.

> I think overall the main issues with the current approach is the complex crashes due to bugs when doing the manual AST walking, which I hope we could make more robust by using Clang's visitor stuff.

I think so too, but it's a bit difficult to map the current decl types to Clang's visitor since it uses a lot more types.
I have frameworked a visitor implementation for `Expr`. I'll push it as a draft so you can see what I have prepared. I think that should give you an idea of what I'm dealing with

PR #1920, opened 2025-03-04
Update DevManual.md

duckdoom5:
Some info about the codebase for future contributors


duckdoom5:
Yes, I think that would be possible in the near future.

Right now I have some concerns about the way certain properties are ignored and added without proper validation that would make me say 'a developer would need to look at it anyway'

PR #1923, opened 2025-03-06
Allow static class conversion for classes with implicit/deleted constructors

duckdoom5:
Title.

Also added some tests


duckdoom5:
Some error in the CSharp bindings again. I'll check it out

duckdoom5:
Hmm, still some edge case. I'll take a look later

duckdoom5:
@tritao This is also ready for review now :)

PR #1945, opened 2025-09-22
Fix ParseHeader() can only report the last file's error in non unityBuild mode

zhaojunmeng:
```
        if (i < Headers.size() - 1)
            // Before modified, the ParseResult is just deleted without checking whether it has any error.
            delete parser.Parse({ Headers[i] });
        else
            res = parser.Parse({ Headers[i] });
```
Change to: if any file has non success ParseResult, just stop parsing and return the result.

### DIFF
--- a/src/CppParser/Parser.cpp
+++ b/src/CppParser/Parser.cpp
@@ -5017,9 +5017,16 @@ ParserResult* ClangParser::ParseHeader(CppParserOptions* Opts)
         Parser parser(Opts);
 
         if (i < Headers.size() - 1)
-            delete parser.Parse({ Headers[i] });
-        else
+        {
             res = parser.Parse({ Headers[i] });
+            if (res && res->kind != ParserResultKind::Success)
+                return res;
+            delete res;
+        }
+        else
+        {
+            res = parser.Parse({ Headers[i] });
+        }
     }
 
     return res;


tritao:
@zhaojunmeng Is this fix still necessary?

zhaojunmeng:
> @zhaojunmeng Is this fix still necessary?

I believe this is necessary. When I first used CppSharp, I couldn’t generate the wrapper files for my project because some of the header files I parsed had compilation errors—but the ParseHeader() function only reported the parse result of the last header file (which was successful). This took me one day to identify the actual issue.

But I’m having trouble passing the CI action on the Linux platform.

```
Run tests/emscripten/test.sh --dotnet-config $BUILD_CONFIGURATION
Generating bindings with .NET configuration Debug
Generating Emscripten bindings for Emscripten WASM32...
Parsing libraries...
Parsing code...
Error parsing 'Builtins.h, Classes.h, Classes2.h, Delegates.h, Enums.h, Overloads.h'
/home/runner/work/CppSharp/CppSharp/tests/Builtins.h(3,10): fatal: 'cstddef' file not found
CppSharp has encountered an error while parsing code.
```

I thought the include path isn’t set up correctly on Linux when running Emscripten’s tests for tests/Builtins.h to find cstddef.h and cstdint.h.

tritao:
Ok, I can try to help and see what's going on. Can you please re-open or send another PR with this fix?

zhaojunmeng:
> Ok, I can try to help and see what's going on. Can you please re-open or send another PR with this fix?

Thanks, that would be nice! I'll send a new PR for this.

PR #1946, opened 2025-10-01
Restore successful github action build and nuget package generation

mitchcapper:
If you folks don't want to do the nuget.org upload it might be a good idea to do releases that include the nupkg to make life easier for people:)

PR #1948, opened 2025-10-25
fix github action package creation

mitchcapper:
Fixes the build_package GH action per https://github.com/mono/CppSharp/pull/1946#issuecomment-3447724958


tritao:
Re-running them seems to have done the trick, at least for now.

Thanks for the help here.

PR #1956, opened 2025-12-02
Fix ParseHeader() only reporting last file's error in non-unityBuildMode

zhaojunmeng:
Modify ClangParser::ParseHeader() to stop parsing and return immediately if any file has a non-success ParseResult.

### DIFF
--- a/src/CppParser/Parser.cpp
+++ b/src/CppParser/Parser.cpp
@@ -5059,9 +5059,16 @@ ParserResult* ClangParser::ParseHeader(CppParserOptions* Opts)
         Parser parser(Opts);
 
         if (i < Headers.size() - 1)
-            delete parser.Parse({ Headers[i] });
-        else
+        {
             res = parser.Parse({ Headers[i] });
+            if (res->kind != ParserResultKind::Success)
+                return res;
+            delete res;
+        }
+        else
+        {
+            res = parser.Parse({ Headers[i] });
+        }
     }
 
     return res;


tritao:
Maybe the failure is due to the fact that this PR returns early in case of failure, and doesn't end up parsing all header files.

We should probably still keep attempting to parse all files, but we need to keep track in case there is a failure in any of the parsed files, and not return early.

zhaojunmeng:
The CI pipeline failed while executing the _tests/emscripten/test.sh_ script. After debugging this issue, I found the path containing the _cstddef_ header file has not been added to the include paths.

For other test projects (e.g., _CSharp.Gen_), the include path for _cstddef_ (such as _C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\include_ on Windows) is added within the _Parser::Setup()_ function:

```
void Parser::Setup(bool Compile)
{
    ...
    if (Target.getOS() == llvm::Triple::Linux)
        TC = new driver::toolchains::Linux(D, Target, Args);
    else if (Target.getEnvironment() == llvm::Triple::EnvironmentType::MSVC)
        TC = new driver::toolchains::MSVCToolChain(D, Target, Args);

    if (TC && !opts->noStandardIncludes)
    {
        llvm::opt::ArgStringList Includes;
        // calling AddClangSystemIncludeArgs() will add C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\include into Includes variable.
        TC->AddClangSystemIncludeArgs(Args, Includes);
        TC->AddClangCXXStdlibIncludeArgs(Args, Includes);
        for (auto& Arg : Includes)
        {
            if (strlen(Arg) > 0 && Arg[0] != '-')
                HSOpts.AddPath(Arg, frontend::System, /*IsFramework=*/false,
                               /*IgnoreSysRoot=*/false);
        }
    }
    ...
}
```

However, when running _tests/emscripten/test.sh_, _Target.getOS()_ returns _Emscripten_, resulting in a null _TC_ variable, and the include path for _cstddef_ is not added.

PR #1958, opened 2026-02-08
Add premake support for ARM-based macOS

perlun:
This upgrades to premake 5.0.0-beta8, which is the first version that provides ARM64-based binaries for macOS.


perlun:
With these changes in place, I can generate the makefiles without any obvious errors, but I do get some warnings:

### On an Apple Silicon (arm-based) Mac

```
❯ ./build.sh -configuration Release
Directory llvm-cd7080-macosx-arm64-Release already exists.
** Warning: gmake2 has been renamed to gmake. Use gmake to generate makefiles instead.
** Warning: the field flags has been deprecated and will be removed.
   Use dedicated APIs instead.
   @/Users/per/git/CppSharp/build/Helpers.lua(101)

** Warning: the flags value MultiProcessorCompile has been deprecated and will be removed.
   Use `multiprocessorcompile` instead.
   @/Users/per/git/CppSharp/build/Helpers.lua(101)

Using cached LLVM 'Debug' build: /Users/per/git/CppSharp/build/llvm/llvm-cd7080-macosx-arm64-Debug
Using cached LLVM 'RelWithDebInfo' build: /Users/per/git/CppSharp/build/llvm/llvm-cd7080-macosx-arm64-RelWithDebInfo
Using cached LLVM 'Release' build: /Users/per/git/CppSharp/build/llvm/llvm-cd7080-macosx-arm64-Release
Building configurations...
Running action 'gmake'...
Done (41ms).
Generating CppSharp build configuration file 'BuildConfig.cs'
Building configurations...
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CppSharp'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CppSharp.AST'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CppSharp.Parser.CSharp'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CppSharp.Parser.Bootstrap'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CppSharp.Parser.Gen'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CppSharp.Parser'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CppSharp.CLI'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CppSharp.Generator'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CppSharp.Generator.Tests'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CppSharp.Runtime'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'Encodings.Gen'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'Encodings.CSharp'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'Encodings.Tests.CSharp'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'NamespacesDerived.Gen'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'NamespacesDerived.Tests.CSharp'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'VTables.Gen'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'VTables.CSharp'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'VTables.Tests.CSharp'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CLI.Gen'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'Common.Gen'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'Common.CSharp'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'Common.Tests.CSharp'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'StandardLib.Gen'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CSharp.Gen'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CSharp.CSharp'
** Warning: Unsupported toolset 'clang' used for language 'C#' for project 'CSharp.Tests.CSharp'
Running action 'vs2022'...
Done (24ms).
```

I tried running the same on my local `linux-x64`-based machine, but it doesn't seem to detect my clang installation and uses gcc instead, leading to an "llvm is unavailable" error, so hard to compare if these warnings are expected or not:

```
❯ ./build.sh -configuration Release
Downloading: https://github.com/mono/CppSharp/releases/download/CppSharp/llvm-cd7080-linux-x64-gcc-14-Release.tar.xz
Error: llvm-cd7080-linux-x64-gcc-14-Release.tar.xz is unavailable.
Please create your own LLVM package by executing the following commands:
./build.sh clone_llvm
./build.sh build_llvm
./build.sh package_llvm
```

Hopefully, running this in CI will reveal if the premake version upgrade causes any problems for other (existing) platforms.

build/build.sh
```diff
@@ -18,8 +17,17 @@ oshost=""
...
+
+if [[ $_machine == "arm64" ]]; then
+  platform="arm64"
+elif [[ $_machine == "x86_64" ]]; then
+  platform="x64"
+elif [[ $_machine == "i686" ]]; then
   platform=x86
+else
+  echo "Unsupported machine type: ${_machine}"
+  exit 1
```
perlun, line 30: Changed the logic here a bit to be more explicit. The `x86` branch might need to support some more types than `i686` but can be added if/when needed.

build/build.sh
```diff
@@ -125,7 +133,7 @@ download_premake()
 
   if ! [ -f "$premake_path" ]; then
     echo "Downloading and unpacking Premake..."
-    premake_version=5.0.0-beta2
+    premake_version=5.0.0-beta8
     premake_archive=premake-$premake_version-$oshost.$premake_archive_ext
```
perlun, line 145: An important detail: premake 5.0.0-beta8 switches to using arm64 as the default: https://github.com/premake/premake-core/releases/tag/v5.0.0-beta8. This means that this will now "work", but will break things for Intel-based Macs. If we want to support them, the generated file name should be `premake-5.0.0-beta8-macosx-x64.tar.gz` for such platforms.

As far as I'm concerned, we could ignore this and focus on supporting ARM-based macOS for now, but if this is considered an issue, we should resolve this before merge.

tritao:
Thanks for trying to move this along, looks like the Linux build fails with: `tar: ./premake5: Not found in archive`

perlun:
Thanks for approving the workflow @tritao. :+1: Interestingly enough, the error seems to be caused by a difference in the file names inside the archive in the old/new versions:

### Files in old package (`premake-5.0.0-beta2-linux.tar.gz`)

```
❯ tar tvzf premake-5.0.0-beta2-linux.tar.gz 
drwx------ cloudconvert/cloudconvert 0 2022-09-04 05:28 ./
-rwxrwxr-x cloudconvert/cloudconvert 2105168 2022-09-04 04:28 ./premake5
-rw-r--r-- cloudconvert/cloudconvert  113256 2022-09-03 23:28 ./libluasocket.so
-rw-r--r-- cloudconvert/cloudconvert   30944 2022-09-03 23:28 ./example.so
```

### Files in new package (`premake-5.0.0-beta8-linux.tar.gz`)

```
❯ tar tvzf premake-5.0.0-beta8-linux.tar.gz 
-rw-rw-rw- 0/0         2725168 2026-01-24 20:15 premake5
```

Note the difference between `./premake5` and `premake` (without the `./` path prefix).[^1]

I've pushed an edit to the the `build.sh` script to take this into account. Let's see if the CI job works better now.

[^1]: The `tar` binary on macOS (from bsdtar) seemed to work despite this mismatch, because the script worked on macOS despite the mismatch. GNU tar (used on Linux) seems to be more picky in this regard. I've tested the branch with my latest patch on macOS and verified that it still seems to be able to unpack the file from the `.tar.gz` archive.

tritao:
```
/home/runner/work/CppSharp/CppSharp/build/premake/premake5: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by /home/runner/work/CppSharp/CppSharp/build/premake/premake5)
```

I think this is an issue on the Premake side, the builds are not fully portable.

perlun:
> `[...] /lib/x86_64-linux-gnu/libc.so.6: version 'GLIBC_2.38' not found (required by /home/runner/work/CppSharp/CppSharp/build/premake/premake5)`
>
> I think this is an issue on the Premake side, the builds are not fully portable.

Got it. I think this is because the Premake binaries were built on a "too new" system (likely Ubuntu 24.04), meaning it depends on a too-new glibc to work on the Ubuntu 22.04 we seem to be using in our CI builds. I guess we don't want to drop Ubuntu 22.04 compatibility at this point.

I added some conditional logic in the script to use different premake versions depending on system. It's a bit more complex but it seems to run on my local machine and my Mac Mini this way, and CI seems to be on its way to succeed this time.

PR #1959, opened 2026-02-20
ParserGen: Support generating bindings on Linux/macOS

perlun:
The previous approach (using backslashes for path separators) only works on Windows.


perlun:
Hmm, the CI run failed. :thinking: Amended the commit to make it try again, to see if it fails consistently.

perlun:
@tritao, I think I may have run into some intermittent CI failure, but CI seems to be green now so this should be ready for merge.

